PR #3, opened 2016-01-05
Create and delete entity

shinyichen:


shinyichen:
Add/remove rows.

I kept table.deleteRow just in case it's useful.

PR #13, opened 2016-03-02
Added new unit tests and Travis CI auto build

mashuaihh:
### Added new unit test files
Put all the files under /test, tests written using Karma  + Jasmine.

### Added Travis CI build
Added .travis.yml so that unit tests will be run every time Travis CI builds.

### Changed Makefile
Make phase will do both 'npm install' and 'bower install' before testing.


js/ngermrest.js
```diff
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
-angular.module('ERMrest', [])
+ermRestApp = angular.module('ERMrest', [])
```
robes: we don't need this `ermRestApp` variable. you can revert this please.

mashuaihh: Yes, I will revert that next time in office. Thanks for reviewing.

PR #33, opened 2016-05-16
Keys

robes:
Summary:
- Adds `memberOf[Foreign]Keys` arrays to `Column` class to make it easier for a client to see what keys or foriegnkey references a column participates in.
- Adds a boolean flag `simple` to `Key` and `ForeignKeyRef` classes so that clients can immediately see that the key contains only a single column. It is actually implemented as a getter that computes the property on demand.
- And then some minor changes to make one introspection function a little more efficient and another to clean up some lint warnings.

I'm going to assign to @karlcz for review from the perspective of whether these seem like good API changes. I would also suggest that @jrchudy or @howdyjessie weigh in on whether it helps the client processing steps they have to do in the data entry app. Finally, a mention for @shinyichen to be aware of the changes.

Additional changes to the API that might help could be:
- add a `referencedBy` array of `ForeignKeyRefs` to the `Table` class as a sort of reverse lookup of any FKs that point to the table.
- add a boolean flag called `association` to the `Table` class to indicate that if it is an association table.
- add an `associations` array of (not sure what?) either `Table`s or `ForeignKeyRef`s to the `Table` class so that clients can easily tell what other tables it is associated with via a many-to-many relationship.

These last few ideas I didn't implement because I'm not as certain about them.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -106,6 +106,7 @@ to use for ERMrest JavaScript agents.
         * [new Key(table, jsonKey)](#new_ERMrest.Key_new)
         * [.colset](#ERMrest.Key+colset) : <code>[ColSet](#ERMrest.ColSet)</code>
         * [.annotations](#ERMrest.Key+annotations) : <code>[Annotations](#ERMrest.Annotations)</code>
+        * [.simple](#ERMrest.Key+simple) : <code>Boolean</code>
     * [.ColSet](#ERMrest.ColSet)
         * [new ColSet(columns)](#new_ERMrest.ColSet_new)
         * [.columns](#ERMrest.ColSet+columns) : <code>Array</code>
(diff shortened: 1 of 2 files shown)


karlcz:
I think the API makes sense abstractly.  Does it cause any issues to create cyclic data structures for the typical browser garbage collectors?

Another key addon might be for unique keys to have a nonnull  field which is only true if *all* of its constituent columns have nullok false? This property of unique and notnull is basically what "primary key" means in SQL, and might have useful meaning to some model-driven UI steps?

robes:
It would have to be a fairly poor GC implementation not to detect that it's a cyclical subgraph. We could add some delete logic if it's an issue with the typical browser. My thoughts...

Rob Schuler

On May 16, 2016, at 2:47 PM, Karl Czajkowski <notifications@github.com<mailto:notifications@github.com>> wrote:


I think the API makes sense abstractly. Does it cause any issues to create cyclic data structures for the typical browser garbage collectors?

Another key addon might be for unique keys to have a nonnull field which is only true if all of its constituent columns have nullok false? This property of unique and notnull is basically what "primary key" means in SQL, and might have useful meaning to some model-driven UI steps?

-

karlcz:
Not sure who is going to review the code and merge it, but it's not me...

robes:
I'll reassign to Jennifer assuming she's back from jury duty now. And you say the API looks fine, so its just up to Jennifer to be okay with the code and then merge it.

PR #38, opened 2016-05-27
Error handling

shinyichen:


js/ermrest.js
```diff
@@ -128,19 +135,38 @@ var ERMrest = (function (module) {
...
-        logout: function () {
+        /**
+         *
+         * @param {String} referrer referred URL
+         * @desc login with the url to redriect back to after logged in
+         *
+         */
+        login: function (referrer) {
+            var url = this._server.uri + '/authn/preauth?referrer=' + module._fixedEncodeURIComponent(referrer);
+            ERMREST.GET(url, 'application/x-www-form-urlencoded; charset=UTF-8', successLogin, errorLogin, null);
```
robes: We should not use the legacy ERMREST module in ermrestjs.

Issue #61, opened 2016-06-23
Updates to `RowSet` API


robes:
Change `RowSet` to `Rows` and add a `Row` class, of which instances of a row are returned by the Rows instance.

```
Rows /* RENAMED from RowSet */
  .data  /* RAW array of json objects return from ermrest */
  .length()
  .get( i ) => Row  /* NEW */

Row /* NEW */
  .data /* RAW json object for the ith result in the RowSet.data */
  .names()
  .get(name)  /* where 'name' is column name */
```

howdyjessie:
The PR for this issue #63 has been merged. Closing this issue now.

PR #63, opened 2016-06-23
Add Row API; rename RowSet class to Rows (#61)

howdyjessie:
This PR:
- Renames the `RowSet` class to `Rows`
- Adds a `Rows.get(index)` method, which returns the ith `Row` from `Rows.data`.
- Adds a `Row` class, which has:
 - `.data` => raw json object for the ith result in `Rows.data`
 - `.names()` => an array of column names in the row
 - `.get(name)` => the value of the column in this row

### DIFF
--- a/js/ermrest.js
+++ b/js/ermrest.js
@@ -891,15 +891,13 @@ var ERMrest = (function (module) {
     /**
      *
      * @memberof ERMrest
-     * @param {ERMrest.Table} table Required.
      * @param {Object} jsonRow Required.
      * @constructor
      */
     function Row(jsonRow) {
         /**
          * @type {Object}
-         * @desc The row returned from the ith result in the Rows.data. It is an
-         * Object that has keys and values ....
+         * @desc The row returned from the ith result in the Rows.data.
          */
         this.data = jsonRow;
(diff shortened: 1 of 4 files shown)


js/ermrest.js
```diff
@@ -881,6 +889,46 @@ var ERMrest = (function (module) {
     };
 
     /**
+     *
+     * @memberof ERMrest
+     * @param {ERMrest.Table} table Required.
```
robes: extra @param here

js/ermrest.js
```diff
@@ -881,6 +889,46 @@ var ERMrest = (function (module) {
...
+         * @returns {Array} Array of column names
+         */
+        names: function() {
+            return Object.keys(this.data);
+        },
+
+        /**
+         *
+         * @returns {String} name name of column
+         * @returns {ERMrest.Column} column
```
robes: extra @returns

robes:
Looks good. Since this PR is labeled with "[WIP]..." I'm not sure its intended to be merged. Anyway I found a couple minor jsdoc errors. If you fix those, you can merge yourself.

PR #71, opened 2016-06-29
Error Retry

shinyichen:
Retry http calls if responses are 500, 503 or 0. Use exponential backoff delay and give up after 10 retries.


js/utilities.js
```diff
@@ -96,6 +96,90 @@ var ERMrest = (function(module) {
...
+     *
+     * @param http
+     * @param q
+     * @param method
+     * @param url
+     * @param data
+     * @returns {Promise}
+     * @private
+     */
+    module._makeRequest = function(http, q, method, url, data) {
```
robes: I think the API here should be a wrapper around the `http` service and not a new api. So you would implement these methods as wrappers: `http.get | .put | .post | .delete(...)`.

Also there's no need for the caller to pass in the `http` and `q` anyway because the module already has references to them. But with my suggested change to the api (as a wrapper on http) that goes away too.

js/utilities.js
```diff
@@ -96,6 +96,90 @@ var ERMrest = (function(module) {
...
+            // successful
+            deferred.resolve(response);
+            return deferred.promise;
+        }, function(response) {
+            if ((response.status === 0 || response.status === 500 || response.status === 503) && retryCount <= 10 ) {
+                // if error is 0, 500 or 503, and retry count is less than 10, retry
+                retryCount += 1;
+                // retry after delay
+                delay = 2^retryCount * 100; // exponential backoff
+                _sleep(delay); // not using setTimeout because setTimeout is asychronous
```
robes: I am not convinced this is a good idea. Unless I'm totally mistaken here my assumption is we should never introduce synchronicity into any of our javascript code. I am inclined to believer that there must be a way we can use the `setTimeout` here.

js/ngermrest.js
```diff
@@ -3,4 +3,46 @@ angular.module('ERMrest', [])
     .factory('ermrestServerFactory', ['$http', '$q', function($http, $q) {
         ERMrest.configure($http, $q);
         return ERMrest.ermrestFactory;
-    }]);
\ No newline at end of file
+    }])
+    .factory('myHttpInterceptor', ['$q', function($q){
```
robes: This probably doesn't belong in the _ngermrest.js_ script not as a new angular service that we are providing to all our clients. Or does Chaise now depend on it for some reason?

js/utilities.js
karlcz: Right, we can only possibly experience retry conditions in our async APIs that already use promises.

What we need is a promise wrapper that tracks retry status and issues one or more one-shot promises to the HTTP layer. It registers its own handlers on those one-shot promises and then finally resolves the outer promise just once when the retry chain terminates on success, terminal error, or retry limit exceeded.

js/ngermrest.js
shinyichen: I only used it for testing, it can be removed.

hongsudt:
@shinyichen this pull request has a conflict.

Issue #60, opened 2016-06-22
Display name hueristics for table and columns


hongsudt:
Use display name hueristics annotation to determine the display name transformation e.g. "My_Table" -> "My Table" or "my_table" -> "My Table". 

We probably need to define a chaise config, a new annotation at the schema and table level on how to derive the display name instead of annotating every columns.. Assigning this to @robes and @karlcz to come up with the annotation syntax.

karlcz:
@robes I've drafted annotation support for this in 2015:display.name_style

robes:
@karlcz The only issue I can see is that the concept of a "title case" or at least the implementation of a title case may be language dependent. That is, is this an English only feature? Would we know how to implement this for any other language? We have a couple options, either we say this is a feature that works for model element names that are use the Latin-1 charset. Other option, we deprecate this feature and have all our databases updated to use a column name exactly as they want it to be displayed. Personally, (although I think I created this monster) I am not fond of this behavior in our apps... :(

I would actually prefer to deprecate this feature. (I think)

Since we've now added so much to this annotation why don't be just update it from `tag:misd.isi.edu,2015:display` to `tag:isrd.isi.edu,2016:display` too?

hongsudt:
@robes Let's assume that this feature will only work for Latin-1 charset (which is pretty much all our deployments).

karlcz:
I think the spec should be open-ended.  We should **not** have to talk about something like Latin1 here. The concept of word-splitting and casing are language specific but well understood.  Our implementation may be best effort.  It is OK to ignore an annotation if an implementation doesn't know how to handle it...
- I expect we'll split words on whitespace, hyphen, and underscore.  A better implementation might apply language-specific heuristics.
- I expect we'll call `toUpperCase()` or `toLocaleUpperCase()` on the first codepoint in each word after splitting.  I think these methods are supposed to be quite general and should do no harm, right?

RFSH:
Does the `underline_space` mean just simply convert each underline to a space? 
Or it means split string on whitespace, hyphen, and underscore? So `a__-_b` would become `a b`.

karlcz:
I think it means the Javascript behavior `"string".replace(/_/g, " ")`

RFSH:
So shouldn't it be something like `str.replace(/[_\-]/g, ' ').replace(/\s\s+/g,' ')` to include multiple spaces and hyphens too?

karlcz:
[...]

PR #73, opened 2016-06-30
Add Displayname Heuristics

RFSH:
This PR adds displayname heuristics support for schemas, tables, and columns (issue: #60).

##### Implementation Details

- To ease support of hierarchical model, I added `_nameStyle` variable to `Schema`, `Table`, and `Column` classes. This variable stores the name style (heuristics), so that each level can easily figure out what heuristics the parent element<sup>1</sup> has.
- `_determineDisplayName()` function, gets the parent element (`parentElement`) and use it when `name_style` is not defined on the current element.
- Since I am using `_nameStyle` of parent element, I had to relocate the declaration of `.tables` in `Schema` to the end of its constructor function and after `displayname` declaration (It's the same case for `.columns`, `.keys`, and `.foreignKeys` in `Table` class). This doesn't change anything other than the way attributes are sorted in `api.md`.

- `_toTitleCase()` has been changed to support non-english characters too. Also hyphen and underscore are added as separators.
- `_underlineToSpace()` is added to convert underline to space.

--
[1]`Schema` is the parent element of `Table` and `Table` is parent element of `Column`.

### DIFF
--- a/js/utilities.js
+++ b/js/utilities.js
@@ -34,7 +34,7 @@ var ERMrest = (function(module) {
     * @param {String} str string to be manipulated.
     * @private
     * @desc
-    * Replaces all underlines with a single space.
+    * Replaces underline with space.
     */
     module._underlineToSpace = function (str) {
       return str.replace(/_/g, ' ');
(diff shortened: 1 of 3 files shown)


js/utilities.js
```diff
@@ -48,21 +59,53 @@ var ERMrest = (function(module) {
...
+                    element._nameStyle = display_annotation.content.name_style;
+                }
             }
         } catch (exception) {
             // no display annotation, don't do anything
         }
+
+        // if name styles are undefined, get them from the parent element
+        // Note: underline_space and title_case might be null.
+        if(parentElement){
```
robes: does this logic work?

I would think it needs to be a recursive call to make sure that the parent has inherited inherited parent's style (and on)

RFSH: It works because when we get to the child, we have its parent's style (During creating the Schema table, we first get its name styles and then we create the table objects and so on).

I could use recursive calls, but this implementation was easier.

robes: right, it doesn't have to literally be recursive. so yes as long as you are hitting up the whole chain thats fine.

robes:
i'm going to merge this. code looks fine to me. i can't really validate until the day we have automated tests... i can only perform a human lint.

PR #74, opened 2016-06-30
Made Ermrestjs Node.js compatible

chiragsanghvi:
@robes 

- Added separate [node.js](https://github.com/informatics-isi-edu/ermrestjs/blob/Test_Revamp/js/node.js) file for Ermrestjs to make it functional in Node.js.  This will become part of the ermrestjs once we build the code using `make build` command.

- Added a small boilerplate for the new Unit tests to be run. Will be working on it in more detail.

For now you can execute that test case by just running 
```
node spec/jasmine-runner.js
```

You might want to delete the node_modules and execute `make build` to install all dependencies again.


spec/sample.spec.js
```diff
@@ -0,0 +1,73 @@
+
+var url = "https://dev.isrd.isi.edu/ermrest", cookie = "ermrest=C6KFIQn2JS37CGovofWnjKfu;";
```
robes: can you remove these from the code? these should come from the user environment.

spec/utils/ermrest-init.js
```diff
@@ -0,0 +1,19 @@
+
+exports.init = function (options) {
+	options = options || {};
+
+	var url = options.url || "https://dev.isrd.isi.edu/ermrest",
```
robes: same with this URL. should come from user environment please.

PR #80, opened 2016-07-12
WIP: Add client id (._cid) to Server instances

howdyjessie:
This adds a `params` argument to `ermrestServerFactory.getServer` and new Servers.

### DIFF
--- a/js/ermrest.js
+++ b/js/ermrest.js
@@ -62,14 +62,18 @@ var ERMrest = (function (module) {
      * URI should be to the ERMrest _service_. For example,
      * `https://www.example.org/ermrest`.
      */
-    function getServer(uri) {
+    function getServer(uri, params) {
 
         if (uri === undefined || uri === null)
             throw new module.InvalidInputError("URI undefined or null");
 
+        if (typeof params === 'undefined' || params === null) {
+            params = {'cid': null};
+        }
+
         var server = _servers[uri];
         if (!server) {
-            server = new Server(uri);
+            server = new Server(uri, params);
 
             server.catalogs = new Catalogs(server);
             _servers[uri] = server;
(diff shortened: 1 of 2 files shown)


howdyjessie:
@robes @karlcz How do I ensure each request has the `cid` in it? Aside from going to each method in ermrestjs that makes a request and appending `&cid=this.cid` to the url, is there a more efficient way?

hongsudt:
I think that's what you are supposed to be doing---making changes to each
method that makes ermrest request.

hongsudt:
It seems like you already have an answer in ermrestjs#69. Never mind.

howdyjessie:
@hongsudt I didn't realize I posted my update on the wrong thread. I'll move it to this thread:

> @robes reminded me we can wrap the $http object, which reminded me we can write an interceptor for this, so I'm working on that.

js/ermrest.js
```diff
@@ -62,14 +62,18 @@ var ERMrest = (function (module) {
...
      * `https://www.example.org/ermrest`.
      */
-    function getServer(uri) {
+    function getServer(uri, params) {
 
         if (uri === undefined || uri === null)
             throw new module.InvalidInputError("URI undefined or null");
 
+        if (typeof params === 'undefined' || params === null) {
+            params = {'cid': null};
```
robes: it's okay if `params` is `null`. we do not need to initialize it to `{'cid': null}`. just leave it null and downsteam you just won't add the params to the url.

hongsudt: @robes: I asked Jessie to explicitly set it to "null", so we can
distinguish from the log between the version of chaise that forgets to set
cid vs the older version of chaise.

robes: Okay. that’s a good idea.

PR #81, opened 2016-07-12
Reference

robes:
This is the initial sketch of the revised `Reference` based API. It should not break any existing functionality.


robes:
I would like to sanity check this against a working instance of chaise+ermrest to make sure that the reference branch does not break existing functionality. Then it can be merged to master.

jrchudy:
I pulled the reference branch into my dev environment on dev.isrd and was still able to use viewer, recordedit, record, record-two, recordset, and search without issue from ermrestJS. I am going to merge this PR into master.

PR #82, opened 2016-07-13
Displayname Unit Tests

RFSH:
This PR,
- Removes old test files.
- Updates the test files to include test cases for displayname.
- Updates the "How to test the package" section in readme.
- Updates the .travis.yml file.

@robes @chiragsanghvi  Can you please review this?


test/configuration/displayname.spec.conf/displayname_schema.json
```diff
@@ -0,0 +1,780 @@
...
+        },
+        {
+          "comment": null,
+          "name": "rating",
+          "default": 3,
+          "nullok": false,
+          "type": {
+            "typename": "float4"
+          },
+          "annotations": {
```
robes: these are deprecated annotations that we should not be using or testing. unless i'm reading something wrong.

test/configuration/displayname.spec.conf/displayname_schema.json
```diff
@@ -0,0 +1,780 @@
+{
```
robes: general comment here. this seems like a huge schema just to test display name. i think these should be much smaller scoped and focused so that it just covers the schema needed for the test case at hand. i would expect a couple tables at most so that you can test the display name annotation in the primary entity and in nested entities.

test/specs/01.displayname.spec.js
```diff
@@ -0,0 +1,93 @@
...
+        }, function(err) {
+            console.log("Unable to import data");
+            console.dir(err);
+            done.fail();
+        });
+    });
+
+    // Test Cases:
+
+    it('schema should use its own heuristics.', function(done) {
```
robes: this is too generic 'schema should use its own heuristics'. if i'm looking in a test and see it failed i would not understand what failed from that message.

test/configuration/displayname.spec.conf/displayname_schema.json
RFSH: I copy/pasted the schema that has been created before and just added the `tag:misd.isi.edu,2015:display` annotation for some specific tables. I'll get rid of this unused annotations.

robes:
overall, it looks good. but the schema needs to be focused to just this test case. it looks like there are elements and annotations in the schema that are extraneous to this test case.

test/configuration/displayname.spec.conf/displayname_schema.json
robes: As a general practice, what I think we should all be doing is just creating disctint schemas specific to one test suite. Its okay if it covers a number of test cases but they should be a cohesive set of tests. And this is what I would want the tutorial to point out to other developers. Then for example, if I am adding a feature to displayname handling and need to extend the test case. I can “easily” look at the schema and understand how to make use of it for my test, and how to extend it if it didn’t have what I needed for my test.

test/specs/01.displayname.spec.js
RFSH: @robes I know that the descriptions that I used are too generic. But I don't want them to be very specific to the data as well. 
I changed them and the success output looks like this for now:
```
  For determining display name, 
    ✓ schema should use name styles (`underline_space`, `title_case`) that are defined in its display annotations.
    ✓ table should use its own name that is defined in display annotation.
    ✓ table should use its own name styles (`underline_space`, `title_case`) that are defined in its display annotation.
    ✓ table should use schema name styles when `name_style` and `name` are not defined in its display annotation.
    ✓ column should use its own name that is defined in display annotation.
    ✓ column should use its own name styles (`underline_space`, `title_case`) that are defined in its display annotation.
    ✓ column should use table name styles when `name_style` and `name` are not defined in its display annotation.
    ✓ column should use schema name styles when `name_style` and `name` are not defined in its display annotation nor table's display annotation.
```
Are these good enough?

robes: Yes, these are the right level of specificity, as far as I’m concerned. thanks

RFSH:
@robes I updated the schema and used names that describe the situation better. For instance
- `table_with_name` for a table that has `name` display annotation.
- `table_without_display_annotation`for a table without any display annotation.
- ...

Since the authentication has been changed, the test cases failed in travis CI.

RFSH:
@robes I updated the travis secure key and now tests are passing. Can you please review and merge it?

.travis.yml
```diff
@@ -13,3 +13,4 @@ env:
   global:
```
robes: why is this in github?

there should be private travis ci settings that we can define in our travis account and then set these. this look like (please correct me if I'm wrong) that we are sharing both our hostname and our secure token with the world?

.travis.yml
```diff
@@ -13,3 +13,4 @@ env:
...
+  - secure: TBTq6fxGLwsimS+KClPFjFltRfNiNcj01ztbPgtFARlNTPWkoq43av3rCuBZLAeudiFqFxqArVNiQ+hRnL8PDqQNldAdZEkB9jqcGQKHh+S9u3sU9mGLFvXkbIwLo3NWCSYs1S48SBMtr7LZkCU/0jxgNWk/Y4tPjs7lBxcq9aLIUJPcygv6UXrVWwlRcY3J0/2ttflaim59XIXUy2Oxmn7ulXPthIpjwbNoQE0jRkoZgwEWKnLLmUT0fdb2nR75jnqg9NklcThoYWpJeyM0f/uSMQJsvltK9nPQrMAybiNrwIWB8ijYJ8FWU91ebzek8JZhsXEQIaDiO3xTPPL/9OVUZGUiWunjfRIA+3WDTekgJyd52l+3VaIUMnz/+mDiKDqPh6ena9dpaRNcdGT72CPaewmcibGB+1iDJXjm4ph5ySNncI0zhDRMBD49hhrW9C+x2Y05qb13wGAdtoLIA9R1m/CZcXuC0LrUMK2KaWEhjFJKJ51mXi0BXYG4scVePmYp2Mrs6sjdhRnceIu7qxTfdvLq8g6WDhQrpV0LsKGHR3J40ThNqA2aLzOfuCwCBGeJ0qVknkAd4H7xRT9a3OuQUUOdzPZ5ltBxQLGxk10uO8DivaOcpgkt13zf3eVZeMGz9pTyGJk9v4vGyeq8wKb2x+ZW/Bz/xVrDcci9BfM=
```
robes: i gather this is encrypted. i wonder if we can or should do the sam e with the ERMREST_URL?

karlcz: Instead, let's please **stop** writing travis tests that talk to our servers. The travis recipe should include the ermrest install steps to deploy a transient ermrest service on the temporary travis build VM, then run tests against that instead.

.travis.yml
robes: okay so the `secure` token is obviously secure. and its not idea but okay to have our hostname in there.

robes:
@RFSH this is great! just what I wanted to see in the schema and table names and tests. Excellent work. Can I merge it?

RFSH:
@robes Thanks. Let me just add `ERMREST_URL` to the secure key.

RFSH:
@robes I think we can merge it now.

PR #88, opened 2016-07-18
[Revised apis] - Parse and Resolve

jrchudy:
Creating pull request so it's easier to see file changes.


jrchudy:
@howdyjessie, @robes: I read through some more information about promises and figured out what was wrong with my code before. I cleaned it up and think that it is all set to be merged into master.

This contains the functionality of `parse` and `resolve` as well as creating the basic structure of the `Reference` object. You can test it [here](https://dev.isrd.isi.edu/~jchudy/chaise/record-two/#1/legacy:dataset/id=269) with my corresponding changes to record-two.

js/parser.js
```diff
@@ -1,32 +1,244 @@
...
-    module._parse = function(uri) {
-        // TODO parse a uri into some internal structure
+    * This is an internal function that parses a URI and constructs an
+    * internal representation of the URI.
+    * @memberof ERMrest
+    * @function _parse
+    * @param {String} uri - uri to be parsed
+    * @private
+    */
+    module._parse = function(uri, location) {
```
robes: `_parse` should not take the `location` argument. only the `uri`. this needs to be a fully qualified url to the ermrest resource. i.e., `https://example.org/ermrest/catalog/2/entity/s:t/k=123`. That is the only acceptable input to the `_parse` function. Therefore we do not need the `location` parameter. You can head with that in the chaise parse function _before_ you pass the ermrest url to the `resolve` function.

robes: and that will have an effect on the parsing logic below which was only parsing a fragment. so it will be a little different but i don't think it will be hard.

1. chaise parser needs to take the fragment and the location and basically glue them into an ermrest url, then
2. chaise will pass that url to the ermrestjs `resolve` function
3. then `resolve` passes it to `_parse`, and
4. `_parse` then does the real parsing

sorry for the confusion, this was maybe underspecified, but not a huge impact.

robes: otherwise the parser.js looks like just what i'd expect at this stage

js/reference.js
```diff
@@ -52,9 +55,11 @@ var ERMrest = (function(module) {
      * {@link ERMrest.Unauthorized},
      * {@link ERMrest.NotFoundError},
      */
-    module.resolve = function(uri) {
+    function resolve(location) {
```
robes: so, yes, back to my comment in parser. obviously you'll just want to fix this with the revised outline i specified above. so the correct parameter is `uri`

js/reference.js
```diff
@@ -63,12 +68,27 @@ var ERMrest = (function(module) {
...
             // this method needs to internally construct a reference object that
             // represents the `uri` parameter
-            notimplemented();
+
+            // build reference
+            var reference = new Reference(location.href, location);
+            var server = this.getServer(reference._serviceUrl);
+            server.catalogs.get(reference._catalogId).then(function success(catalog) {
+                // Should I make a setter here?
+                reference.catalog = catalog;
```
robes: no, i don't think we need a setter. but i would make this declared as a private property `reference._catalog` instead

robes: same for `reference._schema`, `..._table`, and so on.

general clarification... my skeleton api should not be changed... if you see something missing that I maybe got it wrong but we should discuss it.

js/reference.js
```diff
@@ -109,16 +129,24 @@ var ERMrest = (function(module) {
...
+     *  Clients _do not_ directly access this constructor.
      *  See {@link ERMrest.resolve}.
      * @memberof ERMrest
      * @class
      * @param {!string} uri The `URI` for this reference.
      */
-    function Reference(uri) {
+    function Reference(uri, location) {
         this._uri = uri;
+        var context = module._parse(uri, location);
```
robes: the parser should be called directly in the `resolve` method above. this should not be pushed down into the `Reference` constructor

js/reference.js
```diff
@@ -321,7 +359,37 @@ var ERMrest = (function(module) {
...
-                notimplemented();
+
+                // TODO: get the server
+                var uri = server.uri + "/catalog/" + this._catalogId + "/schema/" +
+                    module._fixedEncodeURIComponent(this._schemaName) + ":" +
+                    module._fixedEncodeURIComponent(this._table.name);
+                    // TODO: filters
+
+
+                return module._http.get(uri).then(function success(response) {
```
robes: okay but this is the part that you were going to fix to only get back the `limit` tuples, right?

jrchudy:
@robes, @howdyjessie: I made the changes that @robes suggested in his comments. The Reference constructor no longer relies on the parse function and parse uses the ermrest resource URI instead of chaise URI. 

The above URL for testing the app is still valid.

js/reference.js
```diff
@@ -16,10 +16,13 @@
 
 var ERMrest = (function(module) {
 
+    // attach this to the angular factory object
+    module.ermrestFactory.resolve = resolve;
```
robes: again, this is an api change that wasn't specified...

js/reference.js
```diff
@@ -52,23 +55,36 @@ var ERMrest = (function(module) {
...
+            var context = module._parse(uri);
+            context.baseUri = ermrestUri.baseUri;
+            var reference = new Reference(context);
+
+            var server = this.getServer(reference._serviceUrl);
+            server.catalogs.get(reference._catalogId).then(function success(catalog) {
+                // Should I make a setter here?
+                reference._catalog = catalog;
+                var schema = reference._schema = catalog.schemas.get(reference._schemaName);
+                reference.table = schema.tables.get(reference._tableName);
```
robes: i don't think we specified that `reference` would have a public `.table` property did we?

js/reference.js
```diff
@@ -152,16 +170,26 @@ var ERMrest = (function(module) {
          * ```
          * @type {ERMrest.Column[]}
          */
-        columns: null,
+         columns: function() {
```
robes: another api change here. columns is a getter not a function.

jrchudy:
Changes from @robes's last comments. Made table private, resolve is attached to the module ERMrest instead of the factory, columns returned to null (will update in next set of changes).

jrchudy:
Added cid parameter based on @howdyjessie's last set of changes.

robes:
@jrchudy I just made changes to return the api to the original sketch that `resolve` should take a `uri`. I also updated the implementation of `parse` so that it can extract the base uri and path. Do you have anyway of validating that it works?

Issue #51, opened 2016-06-21
Update Columns to return visible ordered column list


hongsudt:
The function should integrate the column order and ignore annotations
* If a 2016:visible-columns annotation exists, show the columns in the annotation's specified order.
  * If the column name in the list is unknown (e.g. mis-spell), ignore that column. 
* Default: Show columns in the order defined in the database

API update.

```
Columns
 ...
 .byContext(contextstring) => Columns
```

The `Columns` returned `byContext()` should in the appropriate order per heuristics or annotations.
Use the same guideline as #68 for declaring contextstring.

howdyjessie:
To expand on determining the visibility of a column: The absence of the `2016:ignore` annotation on a column doesn't necessarily mean the column should be visible. See #59 for more details on the hierarchical scoping of `2016:ignore` annotations.

hongsudt:
if the table or schema is invisible, we should throw an error message. Maybe, this check is can be part of the ACL.

robes:
I'm guessing this was reassigned to me and karl because we had a discussion in the afternoon that may change this column order annotation significantly. I think we agreed in principle to the change but @hongsudt might be the one (?) to define her proposed annotation and then share that.

karlcz:
@robes @hongsudt  I went ahead and drafted 2016:visible-columns annotation in the ermrest annotations doc.

RFSH:
@robes 

Since the visible columns annotation is defined on table, isn't it better if we add the `byContext(context)` function to the `Table` class and rename it to something like `getColumns(context)`?

From what I understood, `Columns` may include columns from different tables and order doesn't make sense in this case (since it's defined for columns of each table individually). Unless `Columns` object is used just for columns of a single table.

RFSH:
@robes I think the main purpose of `Columns` class was to store columns of a single table. Therefore we can add a private attribute (`_table`) to this class. This way, I can get the order from the `Columns` class and the `byContext(context)` function can be defined in this class.

PR #90, opened 2016-07-18
Update Columns to return visible ordered column list

RFSH:
Related Issue: #51 

This PR,
- Changes `._columns` in `Columns` class from Object to Array.
- Adds a `.table` to `Columns` class since it is only used to store columns of a single table.
- Adds `.contextualize(context)` to `Columns` class, which contextualizes the object and returns a new `Columns` object based on [its annotation](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2016-visible-columns).
- Adds the following unit tests:

   For determining order of visible columns, 
      - should use the defined order in annotations based on its context.
      - should use `entry` context when context is `edit` or `create` and they are not present in annotation.
      - should use default (`*`) context for any context not matched by a more specific context name.
      - should use the default order when context not matched by a more specific context name and default (`*`) context is not present.
      - should ignore the column names that do not correspond to any column in the table.
      - should ignore repetitive column names and just consider the first one.

### DIFF
--- a/js/ermrest.js
+++ b/js/ermrest.js
@@ -273,7 +273,7 @@ var ERMrest = (function (module) {
     Schemas.prototype = {
         constructor: Schemas,
 
-        _push: function(schema) {
+        _push: function (schema) {
             this._schemas[schema.name] = schema;
         },
 
@@ -293,7 +293,7 @@ var ERMrest = (function (module) {
          *
          * @returns {Array} Array of all schemas in the catalog
          */
-        all: function() {
+        all: function () {
             var array = [];
             for (var key in this._schemas) {
                 array.push(this._schemas[key]);
@@ -414,7 +414,7 @@ var ERMrest = (function (module) {
     Tables.prototype = {
         constructor: Tables,
 
-        _push: function(table) {
+        _push: function (table) {
             this._tables[table.name] = table;
         },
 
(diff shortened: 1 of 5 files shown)


js/ermrest.js
```diff
@@ -1045,15 +1085,64 @@ var ERMrest = (function (module) {
...
+            return this._columns[pos];
+        },
 
+        /***
+         * Get columns based on the context.
+         *
+         * @param {string} context
+         * @returns {ERMrest.Columns} columns with predefined order based on the context.
+         */
+        contextualize: function (context) {
```
robes: i believe, this should be a private not public api change.

robes:
Overall, it looks good. But two unexpected changes were made to the _public_ api. Those `.table` and `.contextualize()` ought to be _private_ (i.e., `._...`). After you fix those issues, and resovle the conflicts (due to other's commits) and this is ready to be merged.

RFSH:
@robes But if we make the `.contextualize()` function private, then nobody can call it outside Columns class. How is it going to be called?

robes:
@RFSH private properties can be used internal to the ERMrest module itself. To clarify, if something is declared pseudo-private by convention "_abc" it means we are telling our public clients of the API, "don't use that". Its okay if _we_ use it internal to the api. It doesn't have to be internal to a class, it can be used by us across our sub-modules.

RFSH:
@robes `contextualize()` and `table` are now private. Conflicts with master are resolved too.

Issue #67, opened 2016-06-24
add .comment to all relevant model elements


hongsudt:
Add .comment to 
- [ ] table
- [ ] schema
- [ ] key
- [ ] foreign key

karlcz:
ERMrest supports comment on schema, table, column, key, and foreign key.

hongsudt:
What's the difference between column comment and key/foreign key comments?
When should key/foreign key comments be displayed to the user?

robes:
Let's expand the scope of this to all the relevant model elements that don't already have the `.comment`. Karl listed them above.

howdyjessie:
I opened PR #95 for review.

howdyjessie:
The PR has been merged but I should write some unit tests with Chirag's new unit testing setup.

howdyjessie:
I was sidetracked with work on #83, but now the tests are available for review on PR #125.

PR #95, opened 2016-07-20
Add .comment to Schema, Table, Column, Key, Foreign Key classes

howdyjessie:
This is a small PR for #67.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -1,4 +1,5 @@
 <a name="ERMrest"></a>
+
 ## ERMrest : <code>object</code>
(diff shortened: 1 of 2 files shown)


robes:
jsdoc comments should have descriptions:

```
/**
 * <descriptoin goes here>
 * @type {string}
 */
...
```

howdyjessie:
@robes I added the descriptions. Is this how you do it?

robes:
That's okay, but my example I gave you was the better way to do it. The `@desc` tag is needed when the descriptoin text _is not_ the first line(s) of text in your jsdoc comment.

PR #99, opened 2016-07-20
Fixed issue #95

chiragsanghvi:
@robes

### DIFF
--- a/js/http.js
+++ b/js/http.js
@@ -123,7 +123,8 @@ var ERMrest = (function (module) {
                         deferred.resolve(response);
                     },
                     function(response) {
-                        if (response.status in _retriable_error_codes && count < max_retries) {
+                        var status = response.status || response.statusCode;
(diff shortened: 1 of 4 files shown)


js/http.js
```diff
@@ -123,7 +123,8 @@ var ERMrest = (function (module) {
                         deferred.resolve(response);
                     },
                     function(response) {
-                        if (response.status in _retriable_error_codes && count < max_retries) {
+                        var status = response.status || response.statusCode;
```
robes: @chiragsanghvi I'm curious. I haven't seen `statusCode` as a property on the `response` from angular. Is that something that Angular used to do or still does or that other libraries use that we should handle in general?

chiragsanghvi: @robes Angular doesn't uses that but in General Node.js Http module returns *statusCode*  in response. This is the reason I added a fallback mechanism.

PR #100, opened 2016-07-20
Reference skeleton

jrchudy:
Created the reference, page, and tuple skeleton. Opening this to more easily see the set of changes.


robes:
looks good to me. only thing is it should support the `limit` parameter. its very easy to implement it i believe.

jrchudy:
Thanks for catching that @robes, the two things left are:
1. Implement `limit` parameter
2. Write the unit tests

jrchudy:
The latest commit includes the changes that @RFSH made to the test folder structure. This can be tested in my [dev environment](https://dev.isrd.isi.edu/~jchudy/chaise/record-two/#1/legacy:dataset/id=269) with the changes to record-two.

shinyichen:
I walked through the code to understand what's going on, and it all makes sense. It looks good to me so far.

PR #101, opened 2016-07-21
Restructure test folder

RFSH:
This PR separates test files into different folders, as it was discussed in the meeting.

Updated File structure:
```
test/
|-- specs
|   |-- SPEC_NAME
|   |   |-- conf
|   |   |   |-- SCHEMA_NAME
|   |   |   |   |-- data                # Data
|   |   |   |   |   `-- TABLE_NAME.json 
|   |   |   |   `-- schema.json         # Schema Definition
|   |   |-- tests                       # Test Case Descriptions
|   |   |   `-- TEST_CASES.js  
|   |   |-- SCHEMA_NAME.conf.json       # schema configuration files
|   |   `-- spec.js                     # include TEST_CASE.js files here                 
|-- support
|   `-- jasmine.json                    # change spec_files to include your spec
|-- utils
|   `-- ermrest-init.js
`-- jasmine-runner.js                   # run this file with node, to run all test specs
```

@robes @karlcz Is this structure good enough?


karlcz:
- Why would SPEC_NAME have to be repeated in directory and again in spec file name? I would assume one spec file w/ a fixed name per suite if each suite lives in its own SPEC_NAME directory.
- Why would there be multiple .schema.json files in one suite?  I would assume a single file w/ fixed name unless you've broken it down per table, in which case shouldn't they also be distributed in the same SCHEMA_NAME/TABLE_NAME hierarchy as the data files?

RFSH:
- Renamed the `SPEC_NAME.spec.js` to `spec.js`.
- About multiple .schema.js, although I renamed it as suggested, I was thinking about having multiple schemas in one test spec. Then for having multiple schemas we need multiple `spec.js` files. Should I keep it the way it is (just one `spec.js` and one `schema.js` for each SPEC_NAME folder)?

karlcz:
Yes I think simpler is better.  We can have multiple suites so there is no reason I can see to have the whole world of variation crammed into a single suite as well.   Just make it one suite == one schema == one starting dataset == one sequence of tests.

karlcz:
BTW please update the PR description with the changes to file structure so when someone merges it, the description will reflect reality.

karlcz:
We were discussing this on slack and I just wanted to summarize because I miscommunicated above. We *do* need multiple schemas (i.e. SQL namespaces) for one test suite. So what we need is a structure to easily express:
- SUITE NAME
  - set of SCHEMA NAME
    - set of TABLE NAME
      - per-table definition
      - per-table data to pre-load
  - set of test definitions to load and run

Issue #104, opened 2016-07-21
Clean up and document angular bindings


robes:
The current angular binding is messy and undocumented:
- [x] When the library is used in angular we need to clean up the global `ERMrest` library. 
- [x] For the revised API sketch we do not need the factory, we just need `ERMrest` to be the factory.
- [x] We need documentation. Ideally this can be generated from jsdocs but it may not be possible and may have to be manually written.

jrchudy:
Merged the PR #107. Closing this issue.

PR #107, opened 2016-07-22
Updates to Angular bindings

robes:
My tests of this branch indicate that it is backward compatible. I have tested against chaise recordset.

I have not been able to test the new injectors and I would like @jrchudy (if possible) to give it a try in his app. What this does is eliminate the need for the ermrestFactory (for our new apps) and cleans up the global variable ERMrest (sort of) but continues to configure the ermrestjs library.

This (if it works) resolves issue #104 .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,11 +32,11 @@ SOURCE=$(JS)/core.js \
 	   $(JS)/filters.js \
 	   $(JS)/utilities.js \
 	   $(JS)/errors.js \
-	   $(JS)/node.js \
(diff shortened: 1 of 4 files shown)


jrchudy:
I pulled this code into a separate branch and tested it with my changes to record-two. I changed the module dependency list from including `ERMrest` to `ermrestjs` and I have `ERMrest` as a dependency in the run function now, as it is available when including `ermrestjs` at the module level. This all worked out of the box and, as @robes stated, it is backwards compatible.

Merging and closing.

Issue #108, opened 2016-07-22
Unit tests for Reference Skeleton


jrchudy:
Unit tests for the reference skeleton need to be created and made sure they are passing.

jrchudy:
This issue is resolved by PR #112. This should be closed when that PR is merged.

jrchudy:
PR was merged, closing this issue.

PR #112, opened 2016-07-26
Reference unit tests

jrchudy:
Unit tests for `Reference.js` which includes tests for the `Reference`, `Page`, and `Tuple` objects. Tests also include fetching a single entity as well as multiple entities with a limit that's smaller than the range of entities.

Unit tests for `Parser.js` which includes tests for the `_parse` function and the `ParsedFilter` object. Tests also include a uri with a single filter and a uri with multiple filters.

Assigning to @chiragsanghvi, @howdyjessie, @RFSH for review. This PR resolves issue #108.


test/specs/parser/parser.conf.json
```diff
@@ -0,0 +1,5 @@
+{
+    "catalog": {},
+    "schema": {},
+    "tables": {}
+}
```
chiragsanghvi: You can delete this file, as you're not creating or depending on any existing schema.

test/specs/parser/spec.js
```diff
@@ -0,0 +1,82 @@
+var includes = require('./../../utils/ermrest-init.js').init();
```
chiragsanghvi: Hey, what you've done is correct, but instead of writing your testcases in the `spec.js` file, you need to create a `tests` folder inside your `parser` folder, which will contain the `01.parser.js` test file.

For reference, you can check the updated wiki on testing [here](https://github.com/informatics-isi-edu/ermrestjs/wiki/Unit-Testing-In-ermrestjs). It mentions the file structure.

After making changes according to the wiki, your `spec.js` file will look like this

```
require('./../../utils/starter.spec.js').runTests({
    description: 'In sample spec, ',
    testCases: [
        "/sample/tests/01.parser.js",
    ],
    schemaConfigurations: []
});
```

For your scenario, you're not depending on a schema, so you don't need to create the `conf` folder and add the `parse.conf.json` file.

test/specs/reference/spec.js
```diff
@@ -0,0 +1,194 @@
+var includes = require('./../../utils/ermrest-init.js').init();
+
+var server = includes.server;
+var ermRest = includes.ermRest;
+var ermrestUtils = includes.ermrestUtils;
+
+describe("For determining reference objects and it's child objects,", function () {
```
chiragsanghvi: As mentioned earlier for `parser` spec, you need to follow the folder structure as mentioned in the [wiki](https://github.com/informatics-isi-edu/ermrestjs/wiki/Unit-Testing-In-ermrestjs).

Apart from that, you can always access the **catalogId** using **process.env.DEFAULT_CATALOG** anywhere in your test file.

For other variables you can hardcode the values. To access the `server`, `ermRest` and other variables you can have a `beforeAll` block, which reads the values from the `options` variable passed to the **execute** function in the test file.

For more info you can refer this [link](https://github.com/informatics-isi-edu/ermrestjs/wiki/Unit-Testing-In-ermrestjs#2-test-spec)

test/specs/reference/reference.conf.json
```diff
@@ -0,0 +1,15 @@
+{
```
chiragsanghvi: This file should be in the `conf/` folder.

test/specs/reference/conf/reference/schema.json
```diff
@@ -0,0 +1,48 @@
+{
```
chiragsanghvi: This file should be in the `conf/reference_schema` folder.

test/specs/reference/conf/reference/data/reference_table.json
```diff
@@ -0,0 +1,8 @@
+[{"id":9000,"name":"Hank","value":12},
```
chiragsanghvi: This file should be in the `conf/reference_schema/data` folder.

jrchudy:
@chiragsanghvi I updated the tests and file structure based on your comments. I added another test case to both reference and parser.

RFSH:
@jrchudy Chirag updated the `.travis.yml` file. Can you please update your branch so it uses the new settings?
And when all checks pass, we can merge.

PR #115, opened 2016-07-28
Contextualize

jrchudy:
Contextualize.record function written with unit tests. _referenceCopy also written. Interesting way I found to create an Obj without using it's constructor and then copy all of the data to it.

Assigning to @chiragsanghvi, @RFSH, @howdyjessie for review.

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -139,7 +139,6 @@ var ERMrest = (function(module) {
         this._tableName  = context.tableName;
         this._filter     = context.filter;
 
-        this._parsedContext = context;
         this.contextualize._reference = this;
     }
 
@@ -263,7 +262,7 @@ var ERMrest = (function(module) {
                 newRef._columns = [];
                 for (var i = 0; i < columnOrders.length; i++) {
                     var column = columnOrders[i];
-                    if (source._columns.indexOf(column)) {
+                    if (source._columns.indexOf(column) != -1) {
                         newRef._columns.push(column);
                     }
                 }
@@ -509,7 +508,10 @@ var ERMrest = (function(module) {
      * @returns {ERMrest.Reference} The copy of the reference object.
      */
     function _referenceCopy(source) {
-        return Object.assign(Object.create(Reference.prototype), source);
+        var referenceCopy = Object.create(Reference.prototype);
+        // referenceCopy must be defined before _clone can copy values from source to referenceCopy
+        module._clone(referenceCopy, source);
+        return referenceCopy;
     }
 
 
--- a/js/reference.js
+++ b/js/reference.js
@@ -138,6 +138,8 @@ var ERMrest = (function(module) {
         this._schemaName = context.schemaName;
         this._tableName  = context.tableName;
         this._filter     = context.filter;
+
+        this.contextualize._reference = this;
     }
 
     Reference.prototype = {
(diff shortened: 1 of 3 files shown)


js/reference.js
```diff
@@ -253,8 +256,18 @@ var ERMrest = (function(module) {
...
-                // TODO: remember these are copies of this reference.
-                return undefined;
+                var source = this._reference;
+                var newRef = _referenceCopy(source);
+                var columnOrders = source._table.columns._contextualize('record').all();
+
+                newRef._columns = [];
+                for (var i = 0; i < columnOrders.length; i++) {
+                    var column = columnOrders[i];
+                    if (source._columns.indexOf(column)) {
```
chiragsanghvi: This will not consider the column which is at index **0** in the `source._columns` array. Instead this statement should check for non **-1** index instead of positive index.

```javascript
if (source._columns.indexOf(column) != -1) {
```

js/reference.js
```diff
@@ -496,8 +509,7 @@ var ERMrest = (function(module) {
      * @returns {ERMrest.Reference} The copy of the reference object.
      */
     function _referenceCopy(source) {
-        // TODO: make a (copy-on-write) copy of the source reference
-        return source; // TODO
+        return Object.assign(Object.create(Reference.prototype), source);
```
chiragsanghvi: @jrchudy 

This method of creating objects without invoking their constructor is really good. 

Should we make **copy** part of some utility function, instead of declaring it in this class ? So we can use it in other classes too.

For example
```javascript
module._copy = function(source, destClass) {
    var obj = {};
   
    // destClass parameter is optional, if you just want to clone an existing object
    if (destClass) obj = Object.create(destClass.prototype);
   
    return Object.assign(obj, source);
}; 
```

RFSH: Since we are supporting IE, we should include the [Polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Polyfill) for `Object.assign`.

test/specs/reference/tests/01.reference.js
```diff
@@ -56,6 +56,14 @@ exports.execute = function (options) {
                 expect(reference._columns).toBeDefined();
             });
 
+            it('contextualize.record should return a contextualized reference object.', function() {
+                var recordReference = reference.contextualize.record;
+
+                expect(recordReference).not.toBe(reference);
+                expect(recordReference.columns.length).not.toBe(reference.columns.length);
+                expect(recordReference.columns.length).toBe(2);
```
RFSH: The test case should also check for the actual columns:
```javascript
```

RFSH: The test case should also check for the actual columns:
```javascript
var columns = Array.prototype.map.call(recordReference.columns, function(column){
  return column.name;
});
expect(columns).toEqual(["name", "value"])
```
Also we should check that other properties of the reference are not changed.

jrchudy:
@chiragsanghvi and @RFSH, I updated the PR based on your comments that were above. I removed the use of `Object.assign()` in favor of using our `_clone` utility function. I updated the test case as well.

howdyjessie:
I've got nothing to add here; it looks good to me. :)

PR #118, opened 2016-07-28
Context enum

jrchudy:
Created an enum for the different types of contexts that is immutable.

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -16,7 +16,7 @@
 
 var ERMrest = (function(module) {
 
-    module._contexts = {
+    module._contexts = Object.freeze({
         COMPACT: 'compact',
         CREATE: 'create',
         DETAILED: 'detailed',
@@ -25,7 +25,7 @@ var ERMrest = (function(module) {
         FILTER: 'filter',
         RECORD: 'record',
         DEFAULT: '*'
-    }
+    });
 
     /**
      * This function resolves a URI reference to a {@link ERMrest.Reference}
--- a/js/reference.js
+++ b/js/reference.js
@@ -16,17 +16,6 @@
 
 var ERMrest = (function(module) {
 
-    module._contexts = Object.freeze({
-        COMPACT: 'compact',
-        CREATE: 'create',
-        DETAILED: 'detailed',
-        EDIT: 'edit',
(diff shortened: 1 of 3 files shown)


js/reference.js
```diff
@@ -16,6 +16,17 @@
 
 var ERMrest = (function(module) {
 
+    module._contexts = Object.freeze({
```
howdyjessie: @RFSH I wonder if we should we apply Object.freeze to module._annotations as well? Annotations shouldn't be mutated either right?

RFSH:
@jrchudy shouldn't this be in `utilities.js` file?

js/reference.js
RFSH: @howdyjessie You're right. `module._annotations` should be immutable as well.

jrchudy:
@hongsudt and I discussed that. Right now it appears that it will only be used by `Reference` objects and tests. Unless you can think of another reason why it should be in `utilities.js`, I'm open to moving it.

RFSH:
@jrchudy In [`core.js` inside `_getColumnOrders()`](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/core.js#L1041,L1048) I actually used some of these contexts.

js/reference.js
chiragsanghvi: I am still unsure whether we should use `Object.freeze` due to its browser compatibility issue. Can't we just use a simple javascript getter for this ? 

[object.defineproperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty)  will work in most browsers apart from IE8.

chiragsanghvi: In addition we should create a separate `context.js`, which would contain all these declarations, instead of searching them in different files.

jrchudy:
@chirag:
`Object.freeze` has nearly identical compatibility issues as `Object.defineProperty`, so I don't see the point in using that over freeze when it does exactly what we want instead of hacking another function to act like freeze. 

I also don't see the point in moving the enum to `context.js` when that file will only contain the enum object and nothing else. By that logic we should create and `annotations.js` file for the annotations enum that again, is purely defining an object.

PR #120, opened 2016-07-28
Added check to avoid sending empty filter in url

chiragsanghvi:
This **PR** fixes the issue of sending empty filter in the URI. It will also fix the recordset issue [#515](https://github.com/informatics-isi-edu/chaise/issues/515) in Chaise.

### DIFF
--- a/js/core.js
+++ b/js/core.js
@@ -642,7 +642,9 @@ var ERMrest = (function (module) {
             var uri = this._getBaseURI(api);
 
             if (filter !== undefined && filter !== null) {
-                uri = uri + "/" + filter.toUri();
+                var filterUri = filter.toUri();
+                // Set filter in url only if the filterURI is not empty
+                if (filterUri.trim().length) uri = uri + "/" + filterUri
             }
 
             // selected columns only
@@ -724,7 +726,9 @@ var ERMrest = (function (module) {
             var uri = this._getBaseURI("aggregate");
 
             if (filter !== undefined && filter !== null) {
-                uri = uri + "/" + filter.toUri();
+                var filterUri = filter.toUri();
+                // Set filter in url only if the filterURI is not empty
+                if (filterUri.trim().length) uri = uri + "/" + filterUri;
             }
 
             uri = uri + "/row_count:=cnt(*)";


hongsudt:
@karlcz can you please review? This should fix the recordset problem.

js/core.js
```diff
@@ -642,7 +642,9 @@ var ERMrest = (function (module) {
             var uri = this._getBaseURI(api);
 
             if (filter !== undefined && filter !== null) {
-                uri = uri + "/" + filter.toUri();
+                var filterUri = filter.toUri();
+                // Set filter in url only if the filterURI is not empty
+                if (filterUri.trim().length) uri = uri + "/" + filterUri
```
jrchudy: If you apply trim to a string and it is reduced to `""`, it will still return a length of 0. Try having it check for greater than that or `!= -1`.

js/core.js
```diff
@@ -724,7 +726,9 @@ var ERMrest = (function (module) {
             var uri = this._getBaseURI("aggregate");
 
             if (filter !== undefined && filter !== null) {
-                uri = uri + "/" + filter.toUri();
+                var filterUri = filter.toUri();
+                // Set filter in url only if the filterURI is not empty
+                if (filterUri.trim().length) uri = uri + "/" + filterUri;
```
jrchudy: Same comment as above.

jrchudy:
I installed the branch to my dev environment, and still see the issue noted in recordset. I'm not sure your changes make any difference or that they solve the root of the problem.

js/core.js
chiragsanghvi: Thats what I am trying to check, if the string length is 0, then don't send the filter.

chiragsanghvi:
@jrchudy  I just pushed a fix for the same on Chaise where I added null handling for `filter.filters`. You can check it [here](https://github.com/informatics-isi-edu/chaise/commit/a5b478ecd3c53c178bca8c3fcce44de9f655c369)

jrchudy:
When you create a pull request that relies on code in another repo, please make a note of that when you comment on it.  Please also provide an environment so a reviewer can see your changes. We shouldn't be testing them for you, just verifying what you have done.

I can't look at the changes that you made on chaise master because the app hasn't been updated to use the newest changes @robes made last week to require `ermrestjs` instead of `ERMrest`. To reiterate, please test your own code before asking for it to be reviewed.

karlcz:
I installed ermrestjs/emptyFilterfix on synapse-dev and now recordset works again for unfiltered URLs.

js/core.js
jrchudy: I did not know that if `String.length() = 0`, then `if (String.length()) {}` would return false.

chiragsanghvi:
You can test the fix [here](https://dev.isrd.isi.edu/~chirag/chaise/recordset/#1/legacy:dataset)

jrchudy:
With the changes to update all of the chaise apps I made to be consistent with @robes's changes he made last week, your changes are working as expected. Merging and closing this PR.

Issue #83, opened 2016-07-13
utility/library for formatting values of different data types


hongsudt:
These functions can take an optional format argument. 
- [x] Integer
```
function print_int(input_int, format) { 
  /* the input is an int or null */
  /* create a Number object and use its formatting method to format the output */
  /* Default: whole number with thousands separator. If null, return empty string */
}
```
- [x] Float
```
function print_float(input_float, format) {
  /* input is a float or null */
  /* create a Number object and use its formatting method to format the output */
  /* Default: thousands separator, 2 decimal number. If null, return empty string */
}
```
- [x] Date
```
function print_date(date_string, format) {
  /* input is a date string or null */
  /* create a javascript date object and use its formatting method to format the output */
}
```
- [x] Datetime/timestamp
```
function print_timestamp(ts_string, format) {
  /* input is a ts string or null */
  /* create a javascript Datetime object and use its formatting method to format the output */
}
```
- [x] Boolean
```
function print_boolean(input_int, format) {
  /* input: a boolean or null */
  /* Default: return "True", "False". If null, return empty string */
}
```
- [x] Markdown (Aref)

- [x]  Text

robes:
Is this a duplicate of #52 ?

howdyjessie:
@robes It's particularly clear here, but the difference is that this issue is basically `Column.formatvalue()` while #52 is meant to describe all the separate helper functions that would be used inside `Column.formatValue()`.

howdyjessie:
A couple links that @chiragsanghvi passed to me that are relevant to this issue that we'll probably want to look at later:

- A JS printf library: https://github.com/alexei/sprintf.js
- Something to help determine the type of a value: https://isi-isrd.slack.com/files/chiragsanghvi/F1WGXNLQN/type.js

howdyjessie:
Currently doing: Changes in Chaise to support display of Markdown.

howdyjessie:
This issue is under review at PR #121.

hongsudt:
Since #121 is merged, I am closing this issue.

PR #121, opened 2016-07-28
Pretty print values to strings for display

howdyjessie:
This PR introduces a skeleton of sorts for pretty printing values of different Column types. The originating issue is #83. The types supported in this PR are booleans, integers, floats, dates, and timestamps, text, and Markdown.

To process Markdown, I'm using something recommended by @RFSH called [markdown-it](https://github.com/markdown-it/markdown-it). This is the not the same Markdown parser as the one currently used in the old Record app, but markdown-it is more regularly updated than the parser used in the old Record app.


js/utilities.js
```diff
@@ -140,6 +140,154 @@ var ERMrest = (function(module) {
...
 
     /**
+     * @function
+     * @param {Object} value A boolean value to transform
+     * @param {Object} [options] Configuration options
+     * @return {string} A string representation of a boolean value
+     * @desc Formats a given boolean value into a string for display
+     */
+    module._printBoolean = function(value, options) {
+        options = (typeof options === 'undefined') ? {} : options;
```
howdyjessie: This `options` line will likely be refactored out to the larger function (`Column.formatvalue`) that will call these print utility functions when that larger function is implemented.

js/utilities.js
```diff
@@ -140,6 +140,154 @@ var ERMrest = (function(module) {
...
+     */
+    module._printInteger = function(value, options) {
+        options = (typeof options === 'undefined') ? {} : options;
+        if (value === null) {
+            return '';
+        }
+
+        // Remove fractional digits; implicitly transform value to number type
+        // TODO: Truncate or round?
+        value = Math.trunc(value);
```
howdyjessie: @hongsudt When removing the fractional part of an input, should `_printInteger` truncate or round to the nearest integer?

js/utilities.js
```diff
@@ -140,6 +140,154 @@ var ERMrest = (function(module) {
...
+            throw new module.InvalidInputError("Couldn't extract timestamp from input" + exception);
+        }
+
+        if (typeof value.getTime() !== 'number') {
+            // Invalid timestamp
+            throw new module.InvalidInputError("Couldn't transform input to a valid timestamp");
+        }
+
+        // ISOString format: YYYY-MM-DDTHH:mm:ss.sssZ
+        return value.toISOString();
```
howdyjessie: @hongsudt I'm defaulting to ISO format for now, which looks like ` YYYY-MM-DDTHH:mm:ss.sssZ`. Is that ok?

js/utilities.js
karlcz: I think it is customary to round a floating point value to integer if converting. You might think of it as printing a floating point with zero fractional digits: `printf("%.0f", 5.55555)` will result in `6`.

js/utilities.js
karlcz: Heh, timestamps from ERMrest will be in ISO formatted strings like that already except usually with a timezone offset like `-0700` rather than `Z`.  Most people find them difficult to read if you don't at least replace the 'T' with a space.

I think that a useful formatting option would need to provide localization choices and/or access to a structured timestamp representation so that custom formatting strings in an annotation could alter presentation of a column, whether to truncate precision, reorder fields, or change separators.

howdyjessie: Yeah that 'T' definitely doesn't help with readability. I replaced the 'T' with a space and the 'Z' with a UTC offset in hours. So if one is in PDT, 'UTC -7'.

RFSH:
@howdyjessie I think this branch is using the outdated `.travis.yml` file and for that the push check is failing. Can you please update your branch?

howdyjessie:
@RFSH Thanks for the heads up; I just merged `master` into this branch, and both checks pass now.

Makefile
```diff
@@ -27,7 +27,10 @@ BOWER=bower_components
 JS=js
 
 # Pure ERMrest API
-SOURCE=$(JS)/core.js \
+SOURCE=markdown-it/markdown-it.min.js \
```
howdyjessie: Added the markdown-it parser in `SOURCE` for now. I'm unsure about how plugins should be situated within ermrestjs' structure. (Maybe a `plugins` or `lib` or `vendor` folder at the same level as the `js` folder?) We can revisit this file structure if/when more third party tools are needed in ermrestjs.

howdyjessie:
So many many thanks to @chiragsanghvi for helping me figure out how to add a 3rd-party tool to ermrestjs and understanding the differences between working in the node and browser environments.

This is ready for review now. Could someone review my tests and the actual pretty print utilities?

The utitlies are in `utilities.js` and begin with `module._print...`.

test/specs/print_utils/tests/01.print_utils.js
```diff
@@ -0,0 +1,72 @@
...
+            expect(_printInteger(23.000000000)).toBe('23');
+            expect(_printInteger(.0000)).toBe('0');
+            expect(_printInteger(23.101)).toBe('23');
+            expect(_printInteger(.1)).toBe('0');
+        });
+
+        it('_printDate() should format dates correctly.', function () {
+            var _printDate = module._printDate;
+            expect(_printDate(null)).toBe('');
+            expect(_printDate('2012-04-25 13:00:00.00 PST', {
```
chiragsanghvi: This testcase if failing on **Travis**. You can check it [here](https://travis-ci.org/informatics-isi-edu/ermrestjs/builds/149269930#L1891).

I see that the build is **green** though one testcase has failed. I have pushed some changes to handle the test suite exit code handling in PR #126 .

howdyjessie: Okay, I fixed the failing test case. I forgot that I shouldn't assume Travis' environment is in the same timezone/locale than my node environment.

howdyjessie:
Assigned to @robes to ask [this question](https://github.com/informatics-isi-edu/ermrestjs/pull/121/files#r73213424) about a file structure for adding 3rd-party tools into ermrestjs.

PR #122, opened 2016-07-29
Enhance ForeginKey Classes

RFSH:
This PR, 

- Adds `. referredBy` to `Table` class.
- Adds `.constraint_names`, `.from_name`, and `.to_name` to `ForeignKeyRef` class.
- Changes the `.get(colset)` function in `ForeignKeys` to return an array instead of just a `ForeignKeyRef`.
- Adds `.toString()` to `Column`, `Colset`, `ForeignKey`, and `Mapping` classes to make the unit testing possible.
- Adds unit tests for foreignkeys.

**Important Notes:**
- I didn't wanted to create a separate test spec just for foreignkey. Instead I named the test spec `common` so that other test cases that are general and related to the `core.js` could be there as well.
- I wanted to test foreign keys across multiple schemas, so I created two schemas under the `common/conf`. The test cases pass and it works properly. However if you look at the [test logs](https://travis-ci.org/informatics-isi-edu/ermrestjs/builds/148452185#L1635), it shows that it couldn't delete the schemas since they are dependent to each other. 

 Should I not support foreign keys across schemas or just ignore this message? 
 It can only cause problem if other test suits try to create a schema with `common_schema_1` or `common_schema_2` name, since these schemas are already in catalog.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: node_js
 
+sudo: required
+dist: trusty
+
 node_js:
 - stable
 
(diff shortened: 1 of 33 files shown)


chiragsanghvi:
It seems good to me, @hongsudt and @jrchudy  can we merge it ?

PR #125, opened 2016-08-02
Unit tests for .comment property on models

howdyjessie:
I have some tests for issue #67. The functionality was reviewed in a previous PR already so I'm more focused on getting feedback for the tests themselves. The tests are in the `test/comment` folder.


test/specs/comment/tests/01.comment.js
```diff
@@ -0,0 +1,30 @@
...
+        var schemaName = "schema_with_comment", schema;
+
+        beforeAll(function (done) {
+            schema = options.catalog.schemas.get(schemaName);
+            done();
+        });
+
+        // Test Cases:
+        it('schema_with_comment should have a non-null .comment property.', function () {
+            expect(schema.comment).toBe("schema with a comment");
```
chiragsanghvi: You might want to add asserts for checking whether **schema** has **comment** variable set in its class, i.e. `schema.hasOwnProperty("comment")`. This was we could make sure that the property exists on particular classes.

The same applies for **table**, **foreignKey** and **key**.

chiragsanghvi: In addition should we be checking for null value for schema/table/key/foreignkey without a comment ?

howdyjessie: @chiragsanghvi Ok I added checks for (1) `<model>.hasOwnProperty('comment')`; and (2) `null` values in `.comment` when the schema doesn't have a comment.

Issue #49, opened 2016-06-21
a function to retrieve 'row name'


hongsudt:
The row-name is derived from:
* 2015:display.row_name annotation
* Otherwise, look for a column with title, name, or label (case insensitive) 

It should be implemented as a dynamic property (i.e., using the getter approach) on the `Tuple` class.

```
Tuple.prototype = {
  /* other methods here */

  get displayName() {
    if (!this._displayName) {
      ...compute the row name based on the annotations and heuristics
      this._displayName = ...computed row name
    }
    return this._displayName;
  }
}
```

Clients will use it like a normal property.
And it will only be computed at the time of use, and only once.

@robes @karlcz I am assuming that we will be using displayName for the row name.

shinyichen:
It wasn't clear to me what 'rowname' meant. How about 'value' or 'displayValue'?

hongsudt:
@shinyichen I am sorry.. This description is out of date. I will update the issue with the Reference API..

PR #127, opened 2016-08-03
Implement Tuple.displayname #49

shinyichen:
The changes correspond to issue #49 to implement Tuple.displayname method. The displayname is derived from:
* 2016:table-display.row_name annotation
* Otherwise, look for a column with title, name, or label

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -750,9 +750,9 @@ var ERMrest = (function(module) {
             if (!this._displayname) {
                 var table = this._ref._table;
                 // if table has display row_name annotation
-                if (table.annotations.contains("tag:isrd.isi.edu,2016:table-display") &&
-                    table.annotations.get("tag:isrd.isi.edu,2016:table-display").contains("row_name")) {
-                    this._displayname = table.annotations.get("tag:isrd.isi.edu,2016:table-display").get("row_name"); // pattern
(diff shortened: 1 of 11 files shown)


js/reference.js
```diff
@@ -747,21 +747,36 @@ var ERMrest = (function(module) {
...
-            //             //    -  implement pattern expansion funcion?
-            //         }
-            //     }
-            // }
-            //
-            // return this._displayname;
+            if (!this._displayname) {
+                var table = this._ref._table;
+                // if table has display row_name annotation
+                if (table.annotations.contains("tag:isrd.isi.edu,2016:table-display") &&
```
robes: @shinyichen, Aref introduced an enumeration for annotation names and we decided that we will add any in-use annoation to that list and then use the enum like `module._annotations.TABLE_DISPLAY` for example. You can see that enum here:

https://github.com/informatics-isi-edu/ermrestjs/blob/280944127798b7ddd07762838e33d45f1e58aa6c/js/utilities.js#L329

robes:
Overall this looks good to me. Just address that issue of the annotations enum and it looks ready to go.

shinyichen:
@robes Just updated the annotation enum.

robes:
@shinyichen looks like its good to go except this error in the travis ci test. can you see what's going on? it seems unrelated to your tests? @chiragsanghvi  any thoughts on this error?

PR #132, opened 2016-08-04
Errortests

chiragsanghvi:
Implemented HTTP Exception Handling test cases for Error Codes **400**, **401**, **403**, **404**, **409**, **500** and **503** in following scenarios .

- **Catalog(GET)**
- **Schema(GET) - NonHTTP** - throws NotFoundException
- **Table(GET) - NonHTTP** - throws NotFoundException
- **Entity(GET, PUT, POST, DELETE)**
- **Attributes(GET)**
- **Aggregate(GET)**

The test-cases mock HTTP requests to simulate respective error scenarios.


hongsudt:
@karlcz @robes Can you guys take a look at the test scenarios?

test/utils/jasmine-matchers.js
```diff
@@ -122,6 +123,36 @@ exports.execute = function() {
                         };
                     }
                 }
+            },
+
+            toThrow: function(util, customEqualityTesters) {
```
howdyjessie: Why did you add your own `toThrow` matcher? Does it override Jasmine's native `toThrow` matcher? What's the difference between this implementation of `toThrow` with the native `toThrowError`?

chiragsanghvi: @howdyjessie  Yes, it overrides Jasmine's `toThrow` matcher. The difference is between the matching criteria. **Jasmine** matches the whole **exception** object, whereas my method matches it based on either the **errorcode**, **message** or the **exception** object itself.

It still works the same way as it should, with some added benefits.

hongsudt:
@chiragsanghvi it reports having a conflict. Do you have to resolve the conflict first?

chiragsanghvi:
@hongsudt  There have been commits in master, so I will need to resolve them and push my changes.

robes:
In the future, it would be better to keep distinct changes in different branches/PRs. This PR says it is about 'Errortests' but includes changes to the packaging and installer, which don't seem directly relevant to the primary purpose of this change.

robes:
This looks good to me. You could merge it any time, in my opinion.

Issue #135, opened 2016-08-04
Packaging of third party dependencies


robes:
We have recently introduced a third party dependency into ermrestjs, namely the "markdown-it" library. At present we have simply concatenated it into our `ermrest.js` as a bundle effectively. While this may be sufficient for now, this is probably not the right way to do it or the advisable way to do it. For instance, what if we bundle a library but our client (like Chaise or other applications) needs that same library but a different version of it and it therefore conflicts.

We should explore advice from the javascript world and see what other suggested approaches are out there. For instance, one might instead just:
- declare the prerequisites for ermestjs in its documentation
- let users of ermrestjs (like the Chaise UI) include ermrestjs and its prerequisites independently.

If we did the above, we would need to make changes to Chaise's installation and deployment scripts.

robes:
@hongsudt I think this is pretty important (but maybe not urgent) and if he had time @chiragsanghvi could be a good person to look into this.

hongsudt:
@robes Pull request #136 is waiting for your review.

chiragsanghvi:
This was done in PR #148 and is in Master. So closing it.

PR #136, opened 2016-08-04
Package install

chiragsanghvi:
This **PR** addresses the issue #135 - Packaging of third party dependencies.

Implemented a way to install dependencies using **bower**. In the file [node.js](https://github.com/informatics-isi-edu/ermrestjs/blob/ff8386bac5002bb67d7e9affe4401f818d57a223/js/node.js) if browser is detected then all the dependencies are injected in the head of the page where `ermrest.js` is included. 

You can have a look at the **network** tab of [this](https://dev.isrd.isi.edu/~chirag/chaise/recordedit/#2985/error_schema:valid_table_name/id=4) page to determine whether the scripts `markdown-it.min.js`, `markdown-it-sup.min.js` and `markdown-it-sub.min.js` are loaded or not.

I need to come up with a way of accepting different value for the [ermrestjs path](https://github.com/informatics-isi-edu/ermrestjs/blob/ff8386bac5002bb67d7e9affe4401f818d57a223/js/node.js#L74) variable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,11 +50,12 @@ before_install:
   - sudo apt-get install -y nodejs
   - cd ermrestjs
   - sudo npm install
-  - sudo bower install
+  - bower install
 
 install:
   - sudo make build
   - sudo service apache2 restart
+  - sudo ls -lR /var/www/html/ermrestjs
 
 before_script:
   - sudo -H -u webauthn webauthn2-manage adduser test1
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,7 @@ before_install:
   - sudo apt-get install -y nodejs
   - cd ermrestjs
   - sudo npm install
+  - sudo bower install
 
 install:
   - sudo make build
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,6 @@ before_install:
 install:
   - sudo make build
   - sudo service apache2 restart
-  - sudo ls -lR 
 
 before_script:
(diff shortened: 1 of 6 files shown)


hongsudt:
@robes can you take a look at this?

hongsudt:
This looks fine to me. Unless anyone else object, you can merge it on Monday.

robes:
@hongsudt thanks for the nudge.

@chiragsanghvi great job with this!

side note... for the record, I feel slightly uneasy about dynamically pulling in scripts, but I have a feeling this is the "right" way to do it in javascript.

Go ahead and merge it anytime.

robes:
As it turns out, on further inspection, this PR breaks the makefile and it seems inconsistent with the inclusion of dependencies (bower or not bower).

I've tried to make the `make install` work without pulling in any node dependencies. We should consult with people doing devops before we change that, because pulling in the node deps pulls in a lot of stuff. We could relax that requirement if we like.

chiragsanghvi:
@robes What is the issue in the makefile. It runs fine on my machine and [TRAVIS](https://travis-ci.org/informatics-isi-edu/ermrestjs/builds/150313944) as well. 

Apart from that the `make install` command has a dependency on `$(BIN)` which is dependent on `$(MODULES)`  to build the package  which requires to install the node dependencies. If we don't do this we won't be able to package `ermrest.js`. 

If this is wrong, can you suggest me what would be the right way to avoid installing node deps on.

Issue #137, opened 2016-08-05
foreign key alias to foreign key ref look up functionality


hongsudt:
We need the ability to look up a foreign key reference based on its different aliases. The plan is to implement a `foreignKeyRefMap` object for each schema. 

@RFSH  Please note that, the schema can sometimes have an empty string. So the code should deal with this.

hongsudt:
@karlcz what cause ermrest to return an empty string schema name? E.g. how should @RFSH set up the table to test this case?

chiragsanghvi:
@hongsudt We can mock the server response for testing this case.

karlcz:
The empty string schema name is used for psuedo foreign key constraints which cannot be manipulated through the REST API alone.

Normally they would only be involved if you use REST API to define a foreign key constraint involving two SQL views or a SQL view and table. The view would have to be defined directly through the RDBMS since we provide no REST API to define views at this point.

RFSH:
@hongsudt Since we decided to add `ForeignKeyMap` to schema, I am using the schema string to find the schema and add `ForeignKeyRef` to its corresponding `ForeignKeyMap`. What should I do In case of an empty schema string? Should I just ignore it?

karlcz:
You should keep one additional special map somewhere in the model to store the constraints qualified by empty string schema name. This is an important feature we'll encounter in several of our use cases where we've customized our catalogs with SQL views that are presented to users via Chaise, and these are involved in the pseudo foreign key constraints to link them to related tables and views.

hongsudt:
Will this be simpler if we implement the map as an equivalence of 2D lookup
table e.g. foriengKeyMap[schema-name][alias-name] and attach the map at the
catalog level?

RFSH:
We decided to add the constraint lookup function to Catalog;
```javascript
Catalog.prototype = {
  byConstraintName(schemaName, constraintName){
    ...
 }
}
```

karlcz:
I think there should be a trivial way to lookup by what is in the annotations, either making this method more clever or making another wrapper like:

   function constraintByNamePair(pair){
     return constraintByName(pair[0], pair[1]);
   }

so you don't have code all over the place picking out the array elements to put them as separate parameters...

robes:
@karl are these _blank named_ schemas still represented as a Schema resource  in ERMrest?

karlcz:
No, the `""` schema name is being used to qualify psuedo constraints
[...]

Issue #124, opened 2016-08-01
setup uri and default columns of Reference.related[i] correctly


hongsudt:
* The uri should be 
`<Main reference URI>/(id1,id2)=(Person:school,Person:city)`
where (id1, id2) are the columns of the main entity, and (Person:school,Person:city) are the related entity's columns. 

* We should remove (Person:school,Person:city) from the default columns of the related entities to remove redundancy.

Issue #123, opened 2016-08-01
Implement heuristic for Referece.related[i].displayName


hongsudt:
Priorities: 
* If from_name is there, use it
* If not, use the table displayName of the related entities

@karlcz suggested data-dependent naming e.g. Person with employee = "USC".  @robes we need your opinion on how to lazily achieve this.

Issue #102, opened 2016-07-21
Reference.related


hongsudt:
This is for related entities (e.g. tables that have foreign key to this table). 
If R1 is the main reference: then R1.related should return an array of all foreign keys existed in other tables (e.g. Table B) that point to R1. 
![reference related](https://cloud.githubusercontent.com/assets/414732/17073844/75a03af8-5029-11e6-8668-b16e81d40119.jpg)

PR #138, opened 2016-08-05
Related reference

RFSH:
This PR,
- Adds `constraintByNamePair` function to `Catalog` which can be used to lookup the constraint names that are defined in `ForeignKeyRef`.
- Adds `_visibleForeignKeys(context)` to `Table` that returns the visible foreign keys that are defined in the corresponding context.
- Adds `has(schema)` function to `Schemas` to check for existence of a schema name.
- Adds the Implementation of `related` reference for inbound foreign keys.
- Updates some of the `toString()` functions that are used for unit testing.
- Adds unit test for visible foreign keys and inbound related reference.

Related Issues: #137, #124, #123, #119, #102


js/core.js
```diff
@@ -600,6 +636,39 @@ var ERMrest = (function (module) {
...
+            var orders = -1;
+            try {
+                var annot = this.annotations.get(module._annotations.VISIBLE_FOREIGN_KEYS);
+                if (context !== undefined && annot && annot.content) {
+                    orders = module._getAnnotationValueByContext(context, annot.content);
+                }
+            } catch (exception) {}
+
+            // no annoation, return all outbound and inbound fks
+            if (orders == -1) {
```
hongsudt: We should only list inbound fks, not outbound in the list, right?

RFSH: Since the annotation doesn't specify anything about inbound or outbound; I am returning all visible fks in `_visibleForeignKeys` function, regardless of their type (inbound, outbound). 

However since I only have implemented inbound related references, when I am using the result of this function, I only check for inbound fks.

js/core.js
```diff
@@ -600,6 +636,39 @@ var ERMrest = (function (module) {
                 // add to referredBy of the key table
                 foreignKeyRef.key.table.referredBy._push(foreignKeyRef);
             }
+        },
+
+        // returns visible foreignkeys.
+        _visibleForeignKeys: function (context) {
+            var orders = -1;
+            try {
+                var annot = this.annotations.get(module._annotations.VISIBLE_FOREIGN_KEYS);
```
shinyichen: You should use the `annotations.contains()` method to check whether this annotation exists, then do the `get()` call. Otherwise you will get `NotFoundException` here.

RFSH:
@shinyichen 
I changed the try-catch blocks to use `contains()` instead; as suggested.

jrchudy:
The set of changes look good to me.

shinyichen:
Ya the rest looks good to me.

RFSH:
@hongsudt Should I merge it?

Issue #129, opened 2016-08-03
Add sorting to Reference


shinyichen:
Add sorting to the Reference class:
- [ ] Parse sort from URI, also implement multiple sort columns.
- [ ] On read(): 
* If sort is available, use that. Append missing key columns.
* If sort is not available, used row_order from table-display annotation. Append missing key columns.
* Otherwise, no sorting
- [ ] When sorting is changed, a new sortReference(sort) method that makes a new Reference with the new sort.

robes:
I'm glad you went this way with it. I got thinking about it yesterday night and felt like this would be the right way too.

I would suggest a change to the interface, instead of `sortReference(sort)`, just name it `sort(ordering)` where `ordering` is an object like this `{ 'column-name': ` either `'asc'` or `'dsc'` `...}`.

Usage will be like:
```
var sortedref = ref.sort( { 'title': 'asc', 'gene': 'dsc', 'id': 'asc' } );
```

And this leads to a separate but related issue that I believe we will need to expose the `uri` from the reference in a way that the client can use it. Because I would anticipate that an application like 'recordset' will make a `sort(...)` call like above and then it will need to get the URI from the refernece in order to display in the permalink. I have thought about this but won't document here.

shinyichen:
I wrote something similar, except input param is in the same format as in the schema annotation for table-display.row_order. `{"column":colname, "descending":true}...]`

karlcz:
The sort needs to be an ordered list of per-column sort instructions
(column name and ascending or descending).

An object/mapping has no ordering.

robes:
Ugh... @karlcz thanks, I keep forgetting about js object keys aren't ordered...

So yes, we should keep with the way Jennifer has it specified.

But I still prefer the shorter function name `sort(...)`. That's a superficial detail though.

Issue #130, opened 2016-08-03
Implement Page.previous and Page.next


shinyichen:
* Update URI parsing to parse `@before` and `@after`, keep it in `Reference`.
* Update `Reference.read()` to using `@before` and `@after`
* Make sort mandatory. Keep the shortest key for Reference, and use that as sort if sort is not specified.
* Implement `previous()` and `next()`
* Remember to remove @before/@after from duplicating Reference with `related()` call.
* Remember to remove @before/@after from duplicating Reference with `sort()` call.

PR #145, opened 2016-08-08
Reference read with sort #129

shinyichen:
This pull request cover issue #129 and relevant test cases.

* Parse sort from URI, also implement multiple sort columns.
* Sort on Reference.read(): 1) Use sort if available, else used row_order from table-display, otherwise don't specify sorting.
* sort() method that makes a new Reference with a different sort.


test/specs/reference/tests/02.reference_sort.js
```diff
@@ -0,0 +1,137 @@
...
+        var multipleEntityUri = options.url + "/catalog/" + catalog_id + "/entity/" + schemaName + ":"
+            + tableName;
+
+
+        // Test Cases:
+        describe('Reference.sort() method, ', function() {
+            var reference, reference2, page;
+            var limit = 10;
+
+            it('resolve should return a Reference object that is defined.', function(done) {
```
RFSH: This test case should not be here. If you want to access the reference object, just use `beforeAll`.
```javascript
beforeAll(function(done) {
  options.ermRest.resolve(singleEnitityUri, {cid: "test"}).then(function(response) {
    reference = response.contextualize.record;
    done();
  }, function(err) {
    console.dir(err);
    done.fail();
  })
});
```

jrchudy:
The changes to ermrestJS look good to me.

js/reference.js
```diff
@@ -396,6 +446,29 @@ var ERMrest = (function(module) {
...
+        sort: function(sort) {
+
+            // make a Reference copy
+            var newReference = _referenceCopy(this);
+            // change ._sort
+            if (!sort || sort.length === 0) {
+                delete newReference._sort;
+            }
+            else {
+                verify((sort instanceof Array), "input should be an array");
```
robes: You could also add a deeper verification here.

```javascript
function isValidElement (element, index, array) {
    return (typeof element == 'object' &&
            typeof element.column == 'string' &&
            typeof element.order == 'string');
}
verify(sort.every(isValidElement), "invalid arguments in array");
```

MDN says that Array.every is supported fairly well.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

karlcz:
In order to have `@before(...)` or `@after(...)` paging modifiers on References, you're going to have to have a sort on the Reference. So, I think the behavior described in this issue is not quite right. We should set a sort if it is missing and also augment any specified sort to include a primary key (if not already there) so that any subsequent Page result from Reference.read() will be able to expose its next/previous page references...

shinyichen:
How about keeping a shortest key in Reference and use that when sort is missing? And I will add that in the Paging issue #130

karlcz:
You also need to append the shortest key to the sort if the requested
sort doesn't include at least one complete key.

shinyichen:
Yes, I already implemented that part here.

PR #147, opened 2016-08-08
Implement Column.formatvalue

howdyjessie:
This PR fleshes out the calculation of the `formatvalue` property in the Column class. In Column instance, the instance's formatvalue property is determined by checking the instance's type and then running the appropriate pretty-print function to return a value in string format.

Since `Column.formatvalue` is basically just a switch statement that calls various pretty-print functions, this PR's tests primarily test that `formatvalue` will route to the correct pretty-print function based on the column's type. Each of the pretty-print functions have tests that should more thoroughly test all the range of potential inputs; they can be found [here](https://github.com/informatics-isi-edu/ermrestjs/blob/master/test/specs/print_utils/tests/01.print_utils.js).

### DIFF
--- a/test/specs/common/tests/02.column.js
+++ b/test/specs/common/tests/02.column.js
@@ -113,6 +113,32 @@ exports.execute = function(options) {
                         expect(formattedValue).toBe(new Date(testVal).toLocaleString());
                     });
                 });
+
+                describe('should call printFloat() to format,', function() {
+                    var formattedValue = undefined;
+
+                    it('float4 columns correctly.', function() {
+                        var spy = spyOn(formatUtils, 'printFloat').and.callThrough();
+                        var col = table1_schema2.columns.getByPosition(7);
+                        formattedValue = col.formatvalue('11.1');
+                        expect(spy).toHaveBeenCalledWith('11.1', undefined);
+                        expect(formattedValue).toBe('11.10');
+                    });
+
+                    it('float8 columns correctly.', function() {
+                        var spy = spyOn(formatUtils, 'printFloat').and.callThrough();
+                        var col = table1_schema2.columns.getByPosition(8);
+                        var options = {numDecDigits: 7};
+                        formattedValue = col.formatvalue('234523523.023045230450245', options);
+                        expect(spy).toHaveBeenCalledWith('234523523.023045230450245', options);
+                        expect(formattedValue).toBe('234,523,523.0230452');
+                    });
+
+                    afterEach(function() {
+                        expect(formattedValue).toEqual(jasmine.any(String));
+                        formattedValue = undefined;
+                    });
+                });
             });
         });
     });
--- a/test/specs/common/tests/02.column.js
+++ b/test/specs/common/tests/02.column.js
@@ -24,7 +24,7 @@ exports.execute = function(options) {
         describe('a column in Table class, ', function() {
             var column;
             beforeAll(function() {
-                column = table1_schema2.columns.getByPosition(2);
+                column = table1_schema2.columns.get('table_1_text');
             });
 
             it('should be defined.', function() {
@@ -39,7 +39,7 @@ exports.execute = function(options) {
                 var formatUtils = options.includes.ermRest._formatUtils;
 
                 it('should be empty string if column value is null.', function() {
-                    var column = table1_schema2.columns.getByPosition(2);
+                    var column = table1_schema2.columns.get('table_1_text');
                     expect(column.formatvalue(null)).toEqual(jasmine.any(String));
                     expect(column.formatvalue(null)).toBe('');
                 });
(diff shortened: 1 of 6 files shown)


test/specs/common/tests/02.column.js
```diff
@@ -0,0 +1,194 @@
...
+
+                    it('float8 columns correctly.', function() {
+                        var spy = spyOn(formatUtils, 'printFloat').and.callThrough();
+                        var col = table1_schema2.columns.getByPosition(8);
+                        var options = {numDecDigits: 7};
+                        formattedValue = col.formatvalue(234523523.023045230450245, options);
+                        expect(spy).toHaveBeenCalledWith(234523523.023045230450245, options);
+                        expect(formattedValue).toBe('234,523,523.0230452');
+                    });
+
```
hongsudt: Should also cover a case where both command and decimal will show up (e.g. a float > 1000)

test/specs/common/tests/02.column.js
```diff
@@ -0,0 +1,194 @@
...
+                    it('int8 columns correctly.', function() {
+                        var spy = spyOn(formatUtils, 'printInteger').and.callThrough();
+                        var col = table1_schema2.columns.getByPosition(11);
+                        var int8 = 9007199254740991; // Max safe integer in JS
+                        var options = {numDecDigits: 7};
+                        formattedValue = col.formatvalue(int8);
+                        expect(spy).toHaveBeenCalledWith(int8, undefined);
+                        expect(formattedValue).toBe('9,007,199,254,740,991');
+                    });
+
```
hongsudt: positive and negative numbers that are less than 1000?

test/specs/common/tests/02.column.js
```diff
@@ -0,0 +1,194 @@
...
+            describe('the .formatvalue property, ', function() {
+                var formatUtils = options.includes.ermRest._formatUtils;
+
+                it('should be empty string if column value is null.', function() {
+                    var column = table1_schema2.columns.getByPosition(2);
+                    expect(column.formatvalue(null)).toEqual(jasmine.any(String));
+                    expect(column.formatvalue(null)).toBe('');
+                });
+
```
hongsudt: undefined?

test/specs/common/tests/02.column.js
```diff
@@ -0,0 +1,194 @@
...
+                    it('date columns correctly.', function() {
+                        var spy = spyOn(formatUtils, 'printDate').and.callThrough();
+                        var testVal = '2016-05-02 13:00:00.00 PST';
+                        var col = table1_schema2.columns.getByPosition(4);
+                        var options = {separator: '.', leadingZero: false};
+                        var formattedValue = col.formatvalue(testVal, options);
+                        expect(spy).toHaveBeenCalledWith(testVal, options);
+                        expect(formattedValue).toEqual(jasmine.any(String));
+                        expect(formattedValue).toBe('2016.5.2');
+                    });
```
hongsudt: So the default format is dot separator, instead of slash separator?

test/specs/common/tests/02.column.js
howdyjessie: I don't understand this question. I don't test for an `undefined` column value in these cases.

test/specs/common/tests/02.column.js
howdyjessie: The default separator is a slash (`/`), but I created an option to specify your own separator if desired.

test/specs/common/tests/02.column.js
howdyjessie: I added a couple more test cases to address this in [this commit](https://github.com/informatics-isi-edu/ermrestjs/pull/147/commits/7bf8ead12754ada56fed7c6ce8b339ef5770e300).

hongsudt:
The code looks good to me.. Unless anyone else has any objective, you should go ahead and merge.

RFSH:
@howdyjessie It's not an objection, but why didn't you use `Columns.get(name)` instead of `Columns.getByPosition(position)`? 
It makes the test cases easier to read, since you can figure out their type with their name. Also if someone adds some new columns, it may break the test suit.

test/specs/common/tests/02.column.js
```diff
@@ -0,0 +1,194 @@
+exports.execute = function(options) {
+    var sampleValues = {
+        date: '2016-05-02',
+        timestamptz: '2011-05-06T08:25:25-07:00'
+    };
+
+    describe('About the Column class, ', function() {
+        var schemaName1 = "common_schema_1",
+            schemaName2 = "common_schema_2";
+        var table1_schema1, // has two outbound fks to table2_schema1. with foreign-key annotation.
```
RFSH: Since you're just using one table (table1_schema2), you don't need this variable. The same goes for `schemaName1`, `table1_schema1`, and `table2_schema1.`

howdyjessie:
@RFSH Ahh you're right; I don't know why I didn't use `Columns.get` instead of `.getByPosition`. I also removed the unnecessary variables from the top of the specs.

robes:
(Very) Minor comment: I noticed that `printInteger` uses round but in your test cases I only noticed a round down case. 
https://github.com/informatics-isi-edu/ermrestjs/blob/master/test/specs/print_utils/tests/01.print_utils.js#L22

howdyjessie:
@robes Okay I just added a case to round up case (`.5` becomes `"1"`) in [this commit](https://github.com/informatics-isi-edu/ermrestjs/pull/147/commits/5fe81bd67a07513bec5c4b2c9e3a568583badfbe).

If there isn't anymore feedback soon-ish, I'll merge this PR.

jrchudy:
No objections from me. The code changes look good.

PR #148, opened 2016-08-10
Package install

robes:
i just want to create a PR here so we can discuss it.


robes:
@chiragsanghvi Let's not pull in the dependencies with Bower. Let's keep markdown-it in our repo like we are doing in Chaise. In Chaise, we have the angular and other dependencies in our vendor folder.

robes:
@chiragsanghvi Let me pull markdown-it in and fix the makefile

chiragsanghvi:
@robes I have made and pushed the changes.

robes:
I made some changes to the makefile. This can be merged any time I think.

Issue #139, opened 2016-08-05
Referece.related should handle pure binary association table


hongsudt:
* `binary` : Has 2 outbound foreign keys. 
* `pure`: The table has only one composite key constraint. This key includes all the columns from both foreign keys.
(As an exception, the table can have one extra key that is made of one serial type column.)

One way to implement this is to set a flag `._isPureBinaryAssociation` flag to indicate whether the table is a pure binary association. Then while the Reference.related is created, add the logic that will update the related entities' `display_name`, `uri`, and `columns` appropriately. See diagram below.

![img_20160805_164444](https://cloud.githubusercontent.com/assets/414732/17453248/938922be-5b2d-11e6-97e4-7abd9861e14c.jpg)

hongsudt:
For now, I would like to make this flag reflect the strict definition above. 

Note: The binary association table could have other literal columns such as rank in protocol_author scenario. However, until we know how to merge tables together, let's not include this scenario.

RFSH:
I changed the definitions accordingly:
* `binary` : Has 2 outbound foreign keys. The table has only one composite key constraint. This key includes all the columns from both foreign keys.
(As an exception, the table can have one extra key that is made of one serial type column.)
* `pure`: All its columns are part of some key constraints.

RFSH:
#167 changed the definitions based on the previous comment and resolved this issue.

PR #150, opened 2016-08-11
Pure and Binary Association Related Reference

RFSH:
This PR:
- Adds `_isPureAndBinaryAssociatio()` function to `Table` class.
- Adds support for pure and binary association related reference.
- Changes `ForeignKeyRef.toString()` function to make the `_uri` change easier.
- Adds test cases.

related Issue: #139


js/core.js
```diff
@@ -666,6 +666,22 @@ var ERMrest = (function (module) {
             }
 
             return result;
+        },
+
+        // figure out if Table is pure and binary association table.
+        // binary: Has 2 outbound foreign keys.
+        // pure: There are no other columns apart from the foreign key columns and columns that are part of key and have serial4 type.
+        _isPureBinaryAssociation: function () {
```
karlcz: I don't see any logic here to verify that there is a composite key constraint combining all the columns from both foreign keys.

js/core.js
```diff
@@ -1964,21 +1980,23 @@ var ERMrest = (function (module) {
...
+        toString: function (reverse){
+            var leftString = "", rightString = "";
+            var columnsLength = this.colset.columns.length;
+            for (var i = 0; i < columnsLength; i++) {
+                var fromCol = this.colset.columns[i];
+                var toCol = this.mapping.get(fromCol);
+                var separator = (i < columnsLength -1 ?",": "");
+
+                leftString += (reverse ? fromCol.name : toCol.name) + separator;
+                rightString += (reverse ? toCol.toString() : fromCol.toString()) + separator;
```
karlcz: I assume the col.toString() prints out the whole qualified name `S:T:C`?  You actually only need to include the qualifier on the *first* column of the rightString, and could just use bare column names for any remaining columns.  This will make the URLs a bit shorter and perhaps easier to read.

RFSH: At first I was doing it as suggested. But it cannot  figure out the columns and returns error. You can check it out in [here](https://dev.gpcrconsortium.org/ermrest/catalog/1/entity/iobox_data:virusproduction/id=3/(site_prov,userid)=(iobox_data:userlist:id,site_prov))

RFSH: At first I was doing it as suggested. But it cannot  figure out the columns and returns error. You can check it out in [here](https://dev.gpcrconsortium.org/ermrest/catalog/1/entity/iobox_data:virusproduction/id=3/(site_prov,userid)=(iobox_data:userlist:id,site_prov))

karlcz: Hmm, OK, this seems like an ERMrest bug then...

shinyichen:
Hi. Why not keep a flag for isPureAndBinaryAssociation? Each time  `_isPureAndBinaryAssociation()` is called, it has to run that code.

RFSH:
@shinyichen I wasn't sure where to add the logic of declaring this flag. Should I add it to the `_introspect` function? 
I was also wasn't sure if it's worth the preprocess. Since it's going to be used just for the related reference.

karlcz:
In Python there is an idiom to define a getter method, i.e. one you access as if it were an attribute, and it runs once on first access and then saves the value to return on subsequent access.  This means it is lazy and only calculates on demand but without repeating the work on frequent access.  Can you do something similar in JS...?

RFSH:
@karlcz What we usually do, and have done in reference, is having a private variable; if that variable is defined we return that value, otherwise we create it.
```javascript
get value() {
  if (this._value === undefined) {
    this._value = "value";
  }
  return this._value;
}
```
Since this flag is private, I didn't want to introduce another private variable. Should I?

karlcz:
I don't see why you shouldn't follow the same approach... just need to make an obvious private property name to store the cached value, e.g. something like this._value_cached

RFSH:
@karlcz I changed the code to verify that there is a composite key constraint combining all the columns from both foreign keys. 

In [line 683 to 693](https://github.com/informatics-isi-edu/ermrestjs/pull/150/files#diff-d85c7ecdc144ac4aec3ee7a0b24c3f9fR683), I verify that all columns from both foreign keys are part of the same key; and they're not part of any other keys.

At first I implemented it the other way around (the code below). Going through all the keys and verify that the composite key combining all the columns from both foreign keys exists.

```javascript
var fkColLength = this.foreignKeys.colsets().reduce(function(prev, next) {
  return prev + next.columns.length;
}, 0); // number of columns from both foreignkeys.
var checkKeyColumns = function(el, index, columns) {
  return el.memberOfForeignKeys.length > 0;
};

// // there is a composite key constraint combining all the columns from both foreign keys
for (var i = 0, keyColumns; i < this.keys.length() && result; i++) {
  keyColumns = this.keys.all()[i].colset.columns;
  if (keyColumns.length === 1 && keyColumns[0].type.name == "serial4") {
    // there can be an extra key with just a column with type=serial4
    continue;
  }

  // all key columns must be part of foreignkey columns
  if (keyColumns.length != fkColLength) {
    result = false;
  } else {
    // verify that all the columns are part of a foreign key
    result = keyColumns.every(checkKeyColumns);
  }
}
```

I am not sure which one is preferable.

karlcz:
I think it's actually OK if the foreign keys in an association table overlap one another! This could happen in some domains where there is natural domain information encoded in composite keys.

You want to test that two mathematical sets are equal:

1. The set of columns from the candidate key
2. The set of referencing columns from both foreign key constraints

If I were doing this, I'd actually write tiny helper functions to do this set manipulation and make the logic really obvious. One approach would be to use hashmaps for each set.

In psuedo-code:

```
function make_set(cols):
   s = {}
   for col in cols:
     s[col] = true
   return s

function sets_equal(s1, s2):
   for e in s1:
     if e not in s2:
       return false
   for e in s2:
     if e not in s1:
       return false
   return true

function is_benign_serial(key, fcols):
   return length(key.columns) == 1 
      && key.columns[0].type is serial 
      && key.columns[0] not in fcols

function is_binary_association(table):
   if length(table.foreign_keys) != 2:
     return false

   fcols = make_set(table.foreign_keys[0].columns + table.foreign_keys[1].columns)

   temp_keys = table.keys.copy()

   for key in temp_keys:
     if is_benign_serial(key, fcols):
       temp_keys.remove(key)
       break

   if length(temp_keys) != 1:
     return false

   return sets_equal(fcols, make_set(temp_keys[0].columns))
```

RFSH:
@karlcz Changed the function based on your pseudo-code.

RFSH:
@karlcz Can you please take a look at it?

js/core.js
```diff
@@ -666,7 +666,46 @@ var ERMrest = (function (module) {
...
+                return false; // not binary
+            }
+
+            var fkColset = new ColSet(this.foreignKeys.colsets().reduce(function(res, colset){
+                return res.concat(colset.columns);
+            }, [])); // set of foreignkey columns
+
+            var tempKeys = this.keys.all().filter(function(key) {
+                var keyCols = key.colset.columns;
+                return !(keyCols.length == 1 && keyCols[0].type.name == "serial4"  && !(keyCols[0] in fkColset.columns));
```
karlcz: There are actually serial2, serial4, serial8 types so this test needs to be broadened...

js/core.js
```diff
@@ -666,7 +666,46 @@ var ERMrest = (function (module) {
...
+            }); // the key that should contain foreign key columns.
+
+            if (tempKeys.length != 1 || !fkColset._equals(tempKeys[0].colset)) {
+                return false; // not binary
+            }
+
+            // check for purity:
+            for (var i = 0, col; i < this.columns.length(); i++) {
```
karlcz: This "check for purity" seems like stray code?  The preceding tests follow the psuedo-code and have already covered the requirements to find exactly one candidate key that has a colset equal to the fkColset (after excluding trivial serial keys).  Just return true here.

RFSH: The other part of the code doesn't check for " There are no other columns apart from the foreign key columns and columns that are part of key and have serial type."
The pseudo-code just checks for binary association, while we need to check for pure too.

karlcz: The temp keys started as all keys and removed the trivial serial keys.

Then, you check that there is only one temp key left and that it is
exactly the foreign key columns (that's what the colset equality test
means).

There is nothing else that needs to be checked since this covered
everything needed for "pure" binary association.  The impure parts,
aside from trivial serial keys we removed earlier, would be left as
extra temp keys and/or a final temp key that fails the colset equality
test.

RFSH: These just check for columns that are involved in keys. But what if there is a column that is not part of any keys.
Based on what we defined, a table that has some columns that are not part of any keys, should not be considered as a pure binary association. 

If a table has a `modification_date` column that is not a foreign key column and is not part of any keys, should we still consider it as "pure binary association"?

karlcz: Ah, I understand now.

As long as the pair of foreign keys forms the key of the table, it means each row associates one foreign key combination. That's all I have meant by "pure" association. The kind of association table that is really impure is one that includes other material into that key combination. Then, more than one row could associate the same pair of foreign keys but with different modifying material, e.g. association modifiers. This means you can get duplicative cross-products of associated entities if you treat it as a simple association and ignore the modifier values.

So, we don't need to exclude other non-key material, but I guess you can if you want.  If you do, it should be through another colset operation rather than this loop. Create a colset of all columns, filter out serial columns, and check that the remaining colset also equals the fk colset.  Try to stick to one set-manipulation paradigm within one routine like this...

RFSH: Oh, ok. I am going to change our definition of "pure" in the issue then; and get rid of this loop.

RFSH:
I Changed the definition of "pure" association and added support for other serial types. 

@karlcz Should I merge it?

PR #151, opened 2016-08-11
Paging #130

shinyichen:
Implement Page.previous() and Page.next() #130 

* Update URI parsing to parse @before and @after
* Update Reference.read() to using @before and @after
* Use that as sort if sort is not specified.
* Implement previous() and next()

### DIFF
--- a/js/parser.js
+++ b/js/parser.js
@@ -61,7 +61,7 @@ var ERMrest = (function(module) {
         } else if (uri.indexOf("@after(") !== -1) {
             context.uri = uri.split("@after(")[0]; // remove modifiers from uri
         } else if (uri.indexOf("?limit=") !== -1) {
-            context.uri = uri.split("?limit=(")[0]; // remove modifiers from uri
+            context.uri = uri.split("?limit=")[0]; // remove modifiers from uri
         }
 
         var modifierPath = uri.split(context.uri)[1];
--- a/js/parser.js
+++ b/js/parser.js
@@ -95,7 +95,7 @@ var ERMrest = (function(module) {
                     var row = modifierPath.match(/@before\(([^\)]*)\)/)[1].split(",");
                     for (var i = 0; i < context.sort.length; i++) {
                         // ::null:: to null, empty string to "", otherwise decode value
-                        var value = (row[i] === "::null::" ? null : (row[i] === "" ? "" : decodeURIComponent(row[i])));
+                        var value = (row[i] === "::null::" ? null : decodeURIComponent(row[i]));
                         context.paging.row[context.sort[i].column] = value;
                     }
                 }
@@ -108,7 +108,7 @@ var ERMrest = (function(module) {
                     var row = modifierPath.match(/@after\(([^\)]*)\)/)[1].split(",");
                     for (var i = 0; i < context.sort.length; i++) {
                         // ::null:: to null, empty string to "", otherwise decode value
-                        var value = (row[i] === "::null::" ? null : (row[i] === "" ? "" : decodeURIComponent(row[i])));
+                        var value = (row[i] === "::null::" ? null : decodeURIComponent(row[i]));
                         context.paging.row[context.sort[i].column] = value;
                     }
                 }
--- a/js/parser.js
+++ b/js/parser.js
@@ -54,22 +54,58 @@ var ERMrest = (function(module) {
             baseUri: uri.slice(0,svc_idx+_service_name_len)
         };
         var path = context.path = uri.slice(svc_idx+_service_name_len); // string after service
+        if (uri.indexOf("@sort(") !== -1) {
+            context.uri = uri.split("@sort(")[0]; // remove modifiers from uri
+        } else if (uri.indexOf("@before(") !== -1) {
+            context.uri = uri.split("@before(")[0]; // remove modifiers from uri
+        } else if (uri.indexOf("@after(") !== -1) {
+            context.uri = uri.split("@after(")[0]; // remove modifiers from uri
+        } else if (uri.indexOf("?limit=") !== -1) {
+            context.uri = uri.split("?limit=(")[0]; // remove modifiers from uri
+        }
+
+        var modifierPath = uri.split(context.uri)[1];
 
         // Parse out @sort(...) parameter and assign to context
         // Expected format:
         //  ".../catalog/catalog_id/entity/[schema_name:]table_name[/{attribute::op::value}{&attribute::op::value}*][@sort(column[::desc::])]"
-        if (uri.indexOf("@sort(") !== -1) {
+        if (modifierPath) {
+            if (modifierPath.indexOf("@sort(") !== -1) {
 
-            var sorts = path.match(/@sort\((.*)\)/)[1].split(",");
-            path = path.split("@sort(")[0];  // anything before @sort(..)
-            context.uri = uri.split("@sort(")[0]; // remove @sort from uri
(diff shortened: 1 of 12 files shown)


js/parser.js
```diff
@@ -54,22 +54,58 @@ var ERMrest = (function(module) {
...
+            if (modifierPath.indexOf("@sort(") !== -1) {
+
+                var sorts = modifierPath.match(/@sort\(([^\)]*)\)/)[1].split(",");
+                path = path.split("@sort(")[0];  // anything before @sort(..)
+
+                context.sort = [];
+                for (var s = 0; s < sorts.length; s++) {
+                    var sort = sorts[s];
+                    var column = (sort.endsWith("::desc::") ?
+                        decodeURIComponent(sort.match(/(.*)::desc::/)[1]) : sort);
```
karlcz: Am I misreading this, or will it only do decodeURIComponent() with descending sorts, but leave the column name encoded for regular ascending sorts?  The name should be decoded the same way in either case...

js/parser.js
```diff
@@ -54,22 +54,58 @@ var ERMrest = (function(module) {
...
-            context.uri = uri.split("@sort(")[0]; // remove @sort from uri
+            // sort must specified to use @before and @after
+            if (modifierPath.indexOf("@before(") !== -1) {
+                if (context.sort) {
+                    context.paging = {};
+                    context.paging.before = true;
+                    context.paging.row = {};
+                    var row = modifierPath.match(/@before\(([^\)]*)\)/)[1].split(",");
+                    for (var i = 0; i < context.sort.length; i++) {
+                        context.paging.row[context.sort[i].column] = decodeURIComponent(row[i]);
```
karlcz: You also need to handle the special value `::null::` here and below for each paging modifier component, before decoding the strings.  This special value indicates an actual `null` for the page key component, while an empty component would be the empty string `""`.

js/reference.js
```diff
@@ -73,6 +73,11 @@ var ERMrest = (function(module) {
                 reference._table   = reference._schema.tables.get(reference._tableName);
                 reference._columns = reference._table.columns.all();
 
+                var keys = reference._table.keys.all().sort(function (a, b) {
+                    return a.colset.length() - b.colset.length();
```
karlcz: We might also want to prefer integer/serial keys to text keys for keys of the same length.

karlcz:
Besides my in-line comments above, I also don't see where the logic is to make sure that every sort has a key in it.  I am not sure how you want to split this requirement between ermrestjs and the chaise apps, but someone needs to append the preferred/default key to the sort component list if an application sort is desired by non-key columns.  This has to happen before the page modifiers are formed, since they need to include the column values from the additional key column(s) for the boundary row...

js/reference.js
```diff
@@ -376,59 +382,92 @@ var ERMrest = (function(module) {
...
+                // add them if they are missing
+                for (var i = 0; i < this._shortestKey.length; i++) { // all the key columns
+                    var col = this._shortestKey[i].name;
+                    // add if key col is not in the sortby list
+                    if (!sortCols.includes(col)) {
+                        this._sort.push({"column":module._fixedEncodeURIComponent(col), "descending":false}); // add key to sort
+                        uri = uri + "," + module._fixedEncodeURIComponent(col);
+                    }
+                }
+                uri = uri + ")";
```
shinyichen: This is where I insert missing key columns in sorting.

karlcz: OK, but, the default or *only* key might be composite, so you can't just check for presence of one column...

js/parser.js
shinyichen: My mistake.

js/reference.js
shinyichen: I am looping through shortestKey, which is an array of columns, and append it to the sort if a column is missing from sort. Is that what you mean?

shinyichen:
Can I merge this branch?

js/parser.js
```diff
@@ -92,7 +92,9 @@ var ERMrest = (function(module) {
                     context.paging.row = {};
                     var row = modifierPath.match(/@before\(([^\)]*)\)/)[1].split(",");
                     for (var i = 0; i < context.sort.length; i++) {
-                        context.paging.row[context.sort[i].column] = decodeURIComponent(row[i]);
+                        // ::null:: to null, empty string to "", otherwise decode value
+                        var value = (row[i] === "::null::" ? null : (row[i] === "" ? "" : decodeURIComponent(row[i])));
```
karlcz: Is this test of empty string (and similar one below) necessary?  Won't `decodeURIComponent("")` return `""` already?

karlcz:
I think this is OK to merge except for a very minor comment I made above.

js/reference.js
robes: that seems correct to me

js/reference.js
```diff
@@ -376,65 +378,101 @@ var ERMrest = (function(module) {
...
-                } else {
-                    var keys = this._table.keys.all().sort( function(a, b) {
-                        return a.colset.length() - b.colset.length();
-                    });
-                    keycols = keys[0].colset.columns;
+                // if no sorting provided, use schema defined sort if that's present
+                // If neither one present, use shortestkey
+                if (!this._sort || (this._sort.length === 0)) {
+                    if (this._table.annotations.contains(module._annotations.TABLE_DISPLAY) &&
+                        this._table.annotations.get(module._annotations.TABLE_DISPLAY).contains("row_order")) {
```
robes: @hongsudt Here's something we haven't explicitly addressed. Notice `"row_order"` which is a key _within_ an annotation, in this case the table display annotation. Do we want those keys to also be from an enumeration? Might be `module._annotations.TABLE_DISPLAY_KEY_ROW_ORDER`? I don't feel strongly about this but I don't mind if we add it to the enumerations either.

robes:
looks fine to merge

js/reference.js
karlcz: Unlike with the annotation `tag:` URIs  I don't see any value in an enumeration for a constant field-name in the JSON structure itself.  I see that as part of the object model and would have just written code like   `annot.row_order` to access it if I were writing the above...

js/parser.js
```diff
@@ -50,33 +50,75 @@ var ERMrest = (function(module) {
...
+
+            } else if (modifierPath.indexOf("@after(") !== -1) {
+                if (context.sort) {
+                    context.paging = {};
+                    context.paging.before = false;
+                    context.paging.row = {};
+                    var row = modifierPath.match(/@after\(([^\)]*)\)/)[1].split(",");
+                    for (var i = 0; i < context.sort.length; i++) {
+                        // ::null:: to null, empty string to "", otherwise decode value
+                        var value = (row[i] === "::null::" ? null : (row[i] === "" ? "" : decodeURIComponent(row[i])));
```
karlcz: I see I commented on an outdated diff, but the same line is still here... Can't `(row[i] === "" ? "" : decodeURIComponent(row[i]))` be simply `decodeURIComponent(row[i])` with exactly the same effect, both here and in the previous conditional block?

shinyichen: I think weird character came up with decode on empty string, that's why I did it this way.

karlcz: Hmm, that sounds like a bug in the decoder if it turns empty string into something else...

karlcz:
@shinyichen yes that was my minor comment.  So I think this can be merged once you fix the merge conflicts again...

shinyichen:
Fixed merge conflict @karlcz

Issue #152, opened 2016-08-12
support show_null annotation 


hongsudt:
This property is context dependent.

PR #158, opened 2016-08-15
Show null

RFSH:
This PR

- adds support for `show_nulls` display annotations.
- Updates the `_getAnnotationValueByContext` function to support multi level sub-contexts. e.g. lookup `entry/create`.  if not found, lookup `entry`.  if not found, lookup `*`. (This function assumes that application contexts are separated by "/").

Related Issue: #152

### DIFF
--- a/js/core.js
+++ b/js/core.js
@@ -1401,9 +1401,9 @@ var ERMrest = (function (module) {
                 value = "";
             } else if (typeof value !== "string") { // default
                 if (context === module._contexts.DETAILED) {
-                    value = null;
+                    value = null; // default null value for DETAILED context
                 } else {
-                    value = "";
+                    value = ""; //default null value
                 }
             }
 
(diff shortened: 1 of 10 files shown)


js/core.js
```diff
@@ -1371,6 +1372,43 @@ var ERMrest = (function (module) {
...
+        },
+
+        // find the null value for the column based on context and annotation
+        _getNullValue: function (context) {
+            if (context in this._nullValue) { // use the cached value
+                return this._nullValue[context];
+            }
+
+            var value = -1,
+                displayAnnot = module._annotations.DISPLAY,
```
howdyjessie: Why not declare `displayAnnot` and `elements` with `var`s?

RFSH: It looks cleaner to me. Avoids having repetitive and unnecessary `var`s.

howdyjessie: Gah sorry I knew I was missing something. You're right; this is fine. I just missed the comma at the end of line 1383.

jrchudy:
No specific comments from me. This PR looks ready to merge.

Issue #156, opened 2016-08-13
update templating library?


hongsudt:
[Mustache](https://github.com/janl/mustache.js/) is a lightweight templating library. 
A heavy version with more set of features is [Handlebars](http://handlebarsjs.com/)

hongsudt:
We have decided to go with mustache. so I am closing this issue.

Issue #140, opened 2016-08-06
update Tuple.value according to the markdown_pattern (e.g. url) annotation  


hongsudt:
Tuple.value should handle the url annotation defined in informatics-isi-edu/ermrest#93.  

If the annotation is defined, the value should be formatted accordingly. We also need a method that the app can check whether to treat the string as an html or not.

hongsudt:
@chiragsanghvi you mentioned a couple of template plugin. Can you list them here?

chiragsanghvi:
[Mustache](https://github.com/janl/mustache.js/) is a lightweight templating library. 
A heavy version with more set of features is [Handlebars](http://handlebarsjs.com/)

hongsudt:
@karlcz @robes can you check these templating libraries? For now, I will assume that we will continue to use the same template that we have been using.

chiragsanghvi:
@hongsudt 

Yes, just to ennumerate…the 4 cases are

1. **Normal Link** : `[Chaise](https://dev.isrd.isi.edu/chaise/search)`

2. **Normal Image**: `![Image](http://assets.barcroftmedia.com.s3-website-eu-west-1.amazonaws.com/assets/images/recent-images-11.jpg)`

3. **Image with size**: `![ImageWithSize](http://assets.barcroftmedia.com.s3-website-eu-west-1.amazonaws.com/assets/images/recent-images-11.jpg =200x150)`

4. **Download Link**: `<a href="http://https://code.jquery.com/jquery-3.1.0.js" download>Download</a>`


Copy following javascript code [here]() to check for their appearance.

```javascript
var md = require('markdown-it')({ html: true }).use(require('markdown-it-imsize'));

console.log(md.renderInline("[Chaise](https://dev.isrd.isi.edu/chaise/search)"));

console.log(md.renderInline("![Image](http://assets.barcroftmedia.com.s3-website-eu-west-1.amazonaws.com/assets/images/recent-images-11.jpg)"));
            
console.log(md.renderInline("![ImageWithSize](http://assets.barcroftmedia.com.s3-website-eu-west-1.amazonaws.com/assets/images/recent-images-11.jpg =200x150)"));   


console.log(md.renderInline("<a href='http://cdn.appacitive.com/sdk/js/appacitive-js-sdk-v1.0.10.min.js' class='btn btn-primary' download>Download</a>"))
```

chiragsanghvi:
@hongsudt @karlcz  Will this syntax work ?

[...]

PR #163, opened 2016-08-18
Display values according to  "markdown_pattern" in "tag:isrd.isi.edu,2016:column-display" annotation

chiragsanghvi:
Please refer this [documentation](https://github.com/informatics-isi-edu/ermrestjs/wiki/Using-%22table-display%22-Annotation) for detailed usage.

This **PR** addresses issues #140, #156, #164 and #165 

`tuple.values` handles the [tag:isrd.isi.edu,2016:column-display](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2016-column-display) annotation as discussed in https://github.com/informatics-isi-edu/ermrest/issues/93.

It also uses [Mustache](https://github.com/janl/mustache.js/) for templating with **encode** ability as mentioned by @hongsudt . You check the **encode** functionality [here](https://tonicdev.com/chiragsanghvi/57b4b5f94c7bbd13004b43f6).

The markdown formats are as follows
```sh
# Iframe with height and width
::: iframe [Chaise](https://dev.isrd.isi.edu/chaise/search){width=800 height=300} \n:::

# Link
[ChaiseLink](https://dev.isrd.isi.edu/chaise/search)

# Download Link
[Jquery Download](https://code.jquery.com/jquery-3.1.0.js){download .btn .btn-primary}

# Image
![Image](http://assets.barcroftmedia.com.s3-website-eu-west-1.amazonaws.com/assets/images/recent-images-11.jpg)

# Image with size
![ImageWithSize](http://assets.barcroftmedia.com.s3-website-eu-west-1.amazonaws.com/assets/images/recent-images-11.jpg){width=800 height=300}

```
You can see the output [here](https://tonicdev.com/chiragsanghvi/57b36c60150dc61300d37918/branches/master) for the same. 

**NOTE**: If the value of any of the columns which are being used in the `markdown_pattern`  are either null or empty, then the pattern will fall back on the `show_nulls` display annotation to return respective value.


js/core.js
```diff
@@ -1274,6 +1274,71 @@ var ERMrest = (function (module) {
...
+            for (var i = 0; i < columns.length; i++) {
+                var cname = columns[i].name;
+                
+                keyValues[cname] = values[i];
+
+                // Check for a match for the search string
+                var search = "{{" + cname + "}}";
+
+                // If the search column is found in the pattern
+                if (pattern && pattern.match(search)) {
```
karlcz: This feels wrong to me... I think we should make it so the template engine itself returns null or raises exception if an expansion fails.  Perhaps we can map raw null values into undefined (absent from the template environment) before doing the expansion?

I think it would be OK/preferrable to patch the template engine if necessary to do this, rather than replicating the template expansion logic outside of it like here.  You don't want to have to search for all possible template syntax that can reference/expand a null column.

js/reference.js
```diff
@@ -850,6 +850,19 @@ var ERMrest = (function(module) {
                     var col = this._ref.columns[i];
                     this._values[i] = col.formatvalue(this._data[col.name], {context:this._ref._context});
                 }
+
+                /*
+                 * use this variable to avoid using computed formatted values in other columns while templating 
+                 */
+                var formattedValues = [];
```
karlcz: The spec says to pre-format all columns before expanding any templates, and to use the pre-formatted values.  This comment seems to contradict that, but I haven't traced all the logic yet to be sure...

js/utilities.js
```diff
@@ -436,6 +437,101 @@ var ERMrest = (function(module) {
...
+
+                            // If there is a caption then add it as a "div" with "caption" class
+                            if (attrs[0].children[1].type == "text") {
+                               html = '<div class="caption">' + md.renderInline(attrs[0].children[1].content)  + "</div>" + html;
+                            }
+
+                            // Encapsulate the iframe inside a div
+                            html = '<p>' + html + "</p>";
+                        }  
+                    }
```
karlcz: This doesn't feel right.  If the conditions are false, we return an empty HTML string?  Shouldn't we drop the iframe container and render the inner markdown if we don't know what to do with it?

karlcz:
Aside from my in-line comments, I also don't understand the examples in the PR description.  It says it is demoing escaping but I don't see anything like that.

Also, I think that if a column contains something that looks like markdown, we have to decide whether it should be passed through (by a template expansion) and rendered as markdown, escaped so that only the template can apply markup, or some more fine-grained control whether by different expansion syntax or just by column type?  I am not sure we've explicitly specified the behavior for this yet.

js/reference.js
chiragsanghvi: The `pre_format` will take place in the `formatPresentation` function in [core.js](https://github.com/informatics-isi-edu/ermrestjs/pull/163/files#diff-d85c7ecdc144ac4aec3ee7a0b24c3f9fR1287).

chiragsanghvi:
@karlcz @hongsudt I have made the changes as suggested.

1. For `iframe` if the internal markdown is invalid then render it directly instead of wrapping it in iframe.
2. For columns with `markdown` type, generate markdown after formatting them .i.e. at the end.
3. Added test-cases to handle above scenarios and the special character and markdown column value in markdown_pattern template scenario.

hongsudt:
@karlcz can you please take a look at this? If it is ok, then @chiragsanghvi can merge.

test/support/single.spec.js
```diff
@@ -1,10 +1,7 @@
 require('./../utils/starter.spec.js').runTests({
     description: 'In reference,',
     testCases: [
-        "/reference/tests/01.reference.js",
```
karlcz: Is this intentional or did you disable other tests during development and forget to reenable them?

chiragsanghvi: Yes, I disabled other tests during development. This file is used when we are doing development to run a single spec file. Updating or removing testcases from this file doesn't hamper running all other test-cases as they're part of some other configuration.

https://github.com/informatics-isi-edu/ermrestjs/wiki/Unit-Testing-In-ermrestjs#test-single-spec-file

karlcz: ok then I think it can be merged assuming no conflicts pop up...

chiragsanghvi:
Made changes as suggested in issues  #164 and #165 to add gene_sequence isHTML support and isHTML getter

howdyjessie:
This looks good to me. 👍

hongsudt:
@chiragsanghvi if this pull request is merged, let's close the issue.

js/core.js
```diff
@@ -1351,6 +1351,65 @@ var ERMrest = (function (module) {
...
+            var utils = module._formatUtils, keyValues = options.keyValues, columns = options.columns;
+
+            /*
+             * TODO: Add code to handle `pre_format` in the annotation
+             */
+
+            var isMarkdownPattern = false, isMarkdownType = false;
+
+            if (this.annotations.contains(module._annotations.COLUMN_DISPLAY) && this.annotations.get(module._annotations.COLUMN_DISPLAY).contains("markdown_pattern")) {
+                isMarkdownPattern = true;
```
shinyichen: Looking at the annotation document, the payload of column display annotation should be 
`{context : { "markdown_pattern": ... }}`. The code assumes the format `{"markdown_pattern": ...}`

hongsudt: Good catch.

chiragsanghvi: I have handled it in PR #173

Issue #141, opened 2016-08-06
Add 'location' property to 'Reference'


robes:
Now that clients will be using `Reference` objects they will know less (or nothing) about the manipulation of the underlying URI, yet they will need _part of_ the URI to use in the Permalinks.

For example, the Chaise recordset app starts up, passes the ERMrest URL to `ERMrest.resolve(uri)`. At this point, the client knows what the permalink should look like (chaise resource reference + catalog and "data path" URL components).

But, let's say the user of the recordset sorts on a column or pages to the next page of results, they will now need access to a "data path" and other URL components in order to form the Permalink.

But also keep in mind that the client needs the URL parsed not just the full string. So giving clients direct access to `Reference._uri` is not enough. What we probably really need is a _location_ property with an interface (something) like this.

A property `Reference.location`:
```
Location
  .uri // getter that returns the full ermrest uri
  .catalog // getter that returns just the catalog name ?
  .path // getter that returns the rest of the path ?
```

This object should make it easy for clients like Chaise to build the Permalink with minimal effort.

shinyichen:
While working on the sorting, I needed the uri up to the filter (@sort and ?limit removed). Should we keep that in Location as well?

robes:
I think the answer is yes, but I'd have to look at an example to be sure.

robes:
Just to add some notes here after discussion.

1. we can clean up the `Reference` a bit. Seems like we don't really need: `_tableName`, `_schemaName`, or `_schema`.
2. the parser function (in js/parser.js) can return the (new) `Location` instance rather than the `context` object it currently returns
3. therefore the `Reference(...)` constructor will take the `location` instead of the `context`, and it might need other parameters (like the `table`?)
4. the `Location` can include the parsed _sort object_ needed internally and can have a getter to serialize the sort object into the `@SORT(...)` form

PR #166, opened 2016-08-23
Add 'location' property to 'Reference' #141

shinyichen:
* cleaned up the Reference fields that are not needed: _tableName, _schemaName and _schema.
* parser function returns the (new) Location instance rather than the context object
* Reference(...) constructor will take the location instead of the context
* the Location include the parsed sort object

### DIFF
--- a/js/parser.js
+++ b/js/parser.js
@@ -17,6 +17,7 @@
 var ERMrest = (function(module) {
 
     module.ParsedFilter = ParsedFilter;
+    module.Location = Location;
 
     /**
      * The ERMrest service name. Internal use only.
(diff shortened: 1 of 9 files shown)


js/parser.js
```diff
@@ -191,11 +196,181 @@ var ERMrest = (function(module) {
...
-                context.filter = new ParsedFilter(type);
-                context.filter.setFilters(filters);
+                this._firstFilter = new ParsedFilter(type);
+                this._firstFilter.setFilters(filters);
+            }
+        }
+
+    }
+
+    Location.prototype = {
```
robes: The getters in this prototype should be documented (jsdoc comments). Most likely in the following style so they appear as properties.

```javascript
/**
 * Some descrpitoin
 * @type {String}
 */
get xyz() {
...
}
```

js/parser.js
```diff
@@ -17,6 +17,7 @@
 var ERMrest = (function(module) {
 
     module.ParsedFilter = ParsedFilter;
+    module.Location = Location;
```
robes: In my opinion, you should not need to make the `Location` constructor a member of the `module`.

js/parser.js
```diff
@@ -49,105 +50,109 @@ var ERMrest = (function(module) {
             throw new module.InvalidInputError('The URI does not contain the expected service name: ' + _service_name);
         }
 
-        var context = {
-            uri: uri, // full uri without modifiers
-            baseUri: uri.slice(0,svc_idx+_service_name_len), // ermrest service url
-            path: uri.slice(svc_idx+_service_name_len) // full path without service
-        };
+        return new Location(uri);
```
robes: The `_parse` function's jsdoc `@return` should be updated.

js/parser.js
```diff
@@ -191,11 +196,181 @@ var ERMrest = (function(module) {
...
+
+        get path() {
+            return this._path;
+        },
+
+        get compactPath() {
+            return this._compactPath;
+        },
+
+        get firstSchemaName() {
```
robes: Technically, its not really "first" schema name or table name. The real semantics here are that these are what in ERMrest-speak we sometimes call the "context" of the URI. Meaning, the thing being operated on (either updated or retrieved or deleted).

A more conventional way of saying that _might_ be to say this is the "projection" of the URL. Meaning that you have some URL `entity/s:t/c=123` and in an "entity" interface you are "projecting" all of the columns of the table "`s:t`".

This isn't a _blocker_ and we need to give this more thought in order to adequately resolve the issue. For the purpose of this PR, I suggest just putting a note in the jsdoc string indicating that these two properties are subject to change soon.

js/reference.js
```diff
@@ -377,70 +372,52 @@ var ERMrest = (function(module) {
 
                 var defer = module._q.defer();
 
-                var uri = this._uri;
+                var uri = this._location.compactUri;
```
robes: We should not need `Reference.uri` anymore. I would think it should be completely replaced by `Reference.location` and if a client needs to get the uri, they do so through that interface.

js/parser.js
```diff
@@ -191,11 +196,181 @@ var ERMrest = (function(module) {
...
+                this._firstFilter = new ParsedFilter(type);
+                this._firstFilter.setFilters(filters);
+            }
+        }
+
+    }
+
+    Location.prototype = {
+
+        get uri() {
```
robes: Ideally, a client would just get the complete uri via the Location object itself. Like

```javascript
...
console.log("This reference has the following uri", reference.location);
...
```

I'm not sure how that is done, but I would expect it to work just like `window.location`. Not sure if that's a `toString()` function that gets called implicitly, but that was the idea I had in mind.

js/reference.js
robes: Nevermind, I think I'm confused here because I'm looking at the diff and cannot see enough of the surrounding code. Scratch this coment.

js/reference.js
```diff
@@ -588,19 +563,15 @@ var ERMrest = (function(module) {
...
 
-                            newRef._schema = otherFK.key.table.schema;
-                            newRef._schemaName = otherFK.key.table.schema.name;
                             newRef._table = otherFK.key.table;
-                            newRef._tableName = otherFK.key.table.name;
+                            newRef._shortestKey = newRef._table.shortestKey;
                             newRef._columns = otherFK.key.table.columns.all();
                             newRef._displayname = otherFK.to_name ? otherFK.to_name : otherFK.key.table.displayname;
-                            newRef._uri = this._uri + "/" + fkr.toString() + "/" + otherFK.toString(true);
+                            newRef._location = new module.Location(this._location.compactUri + "/" + fkr.toString() + "/" + otherFK.toString(true));
```
robes: Seems like this should be a call to _parse:

```javascript
newRef._location = module._parse(this._location.compactUri ...);
```

js/reference.js
```diff
@@ -612,7 +583,7 @@ var ERMrest = (function(module) {
                             }
 
                             newRef._displayname = fkr.from_name ? fkr.from_name : newRef._table.displayname;
-                            newRef._uri = this._uri + "/" + fkr.toString();
+                            newRef._location = new module.Location(this._location.compactUri + "/" + fkr.toString());
```
robes: Same here. Can be replaced with call to `module._parse`

js/reference.js
```diff
@@ -717,14 +688,17 @@ var ERMrest = (function(module) {
...
+                // update paging by creating a new location
+                var paging = {};
+                paging.before = true;
+                paging.row = [];
+                for (var i = 0; i < newReference._location.sortObject.length; i++) {
+                    var col = newReference._location.sortObject[i].column;
+                    paging.row.push(this._data[0][col]); // first row
+                }
+                // create new _location
+                newReference._location = new module.Location(newReference._location.compactUri + newReference._location.sort + module._getPagingModifier(paging));
```
robes: If `Location` had a `clone()` function that returned a new instance of Location based on itself, could this be done like this:

```javascript
var location = this._location.clone();
location.sortObject = ...;
location.pagingObject = ...;

newReference._location = location;
```

js/reference.js
```diff
@@ -755,14 +729,17 @@ var ERMrest = (function(module) {
...
+                // update paging by creating a new location
+                var paging = {};
+                paging.before = false;
+                paging.row = [];
+                for (var i = 0; i < newReference._location.sortObject.length; i++) {
+                    var col = newReference._location.sortObject[i].column;
+                    paging.row.push(this._data[this._data.length-1][col]); // last row
+                }
+                // create new _location
+                newReference._location = new module.Location(newReference._location.compactUri + newReference._location.sort + module._getPagingModifier(paging));
```
robes: Same comment as above about cloning.

js/reference.js
```diff
@@ -783,15 +760,41 @@ var ERMrest = (function(module) {
...
+                    this._pageRef._location.api + "/" + this._pageRef._table.schema.name + ":" + this._pageRef._table.name + "/";
+                for (var k = 0; k < this._pageRef._shortestKey.length; k++) {
+                    var col = this._pageRef._shortestKey[k].name;
+                    if (k === 0) {
+                        uri = uri + module._fixedEncodeURIComponent(col) + "=" + module._fixedEncodeURIComponent(this._data[col]);
+                    } else {
+                        uri = uri + "&" + module._fixedEncodeURIComponent(col) + "=" + module._fixedEncodeURIComponent(this._data[col]);
+                    }
+                }
+                this._ref._location = new module.Location(uri);
```
robes: Another case for `module._parse(uri)` instead of needing the Location constructor.

js/utilities.js
```diff
@@ -437,6 +437,50 @@ var ERMrest = (function(module) {
     };
 
     /**
+     * given sort object, get the string modifier
+     * @param {Object[]} sort [{"column":colname, "descending":boolean}, ...]
+     * @return {string} string modifier @sort(...)
+     * @private
+     */
+    module._getSortModifier = function(sort) {
```
robes: My suggestions should make it unnecessary for this function `_getSortModifier` and the next one `_getPagingModifier` to be properties of the `module`. They could just become private functions within the `parser.js` closure.

The only reason to have them then would be if you want to write unit tests for them (I don't personally think its necessary to have unit tests for small utility functions).

But even then, the right place for these would to put them in `parser.js` not in `utilities.js`.

js/parser.js
shinyichen: I need to call `new Location()` from reference.js, so I made that available.

robes:
Overall, this is looking almost right. I would like to see some or most of my line comments be addressed, unless they are too time consuming.

shinyichen:
@robes I made the all the changes you suggested.

Issue #161, opened 2016-08-17
Reference.related should return a sorted set of references


jrchudy:
Reference.related needs to return a set of references that is sorted based on a default heuristic.

Assigning to @hongsudt to fill in the default order that we want to use. It was discussed to have it alphabetical with some other circumstances when two related references have the same displayname.

RFSH:
When visible-foreign-keys annotation was not defined in the table, we should sort related references using the following steps:
1. Sort by display name.
2. Sort by position of key columns in the table.
3. Sort by position of foreign key columns in the related table.

RFSH:
This issue was resolved by #167 PR. This should be closed when PR is merged.

PR #167, opened 2016-08-23
Update Related Reference

RFSH:
This PR,
- Sorts related references when the ordering is not specified; based on displayname, position of key columns, and position of foreign key columns ([code](https://github.com/informatics-isi-edu/ermrestjs/blob/related-reference/js/reference.js#L617,L632)).
- Changes the definition of `Pure and Binary Association` to not include tables that have extra columns that are not part of any keys.
- changes the `_visibleForeignKey` to `_visibleInboundForeignKeys`.
- Updates the `related` function based on the previous changes.
- Updates the test cases for visible foreign key.
- Adds test cases for when the visible foreign key annotation is not present.
- Adds test cases for when there are some extra columns in the association table that are not part of any keys.

Related Issues: #161, #139 

**Question**: @robes 
I sort columns based on their position in the table definition. For that, I simply use `indexOf`([code](https://github.com/informatics-isi-edu/ermrestjs/pull/167/files#diff-d85c7ecdc144ac4aec3ee7a0b24c3f9fR1880)). However I could use `column.position`. But this property is not used anywhere and it's always undefined. Should I add the logic to change this value and then use it for sorting? or Should we remove this property?


RFSH:
@hongsudt @karlcz 
Based on what we discussed, I changed back the definition of  "pure and binary association".  If table has a column that is not part of any constraint keys, I don't consider it as pure and binary association.

robes:
@RFSH sorry missed your initial query. I don't see a problem (from taking a quick glance at it) with the indexOf use. On the other hand, this might be a good time to implment the `column.position` property if it would make this code more compact.

And the change for the "pure and binary association" seems like the right thing to do too.

RFSH:
@robes I can implement the `column.position` as part of this PR. However I don't think this variable is necessary.

The definition says "The ordinal number or position of this column relative to other columns within the same scope.".
I can think of three different places we are changing "position" of a column in its scope.

- During introspection,  in respect to the schema structure.
- During contextualization, in respect to the visible columns annotation.
- During `reference.related`, since we hide columns that are involved in the FKR.

Correct me if I'm wrong, but I think we want the `Column` object to be immutable, so we can only initialize it in the first scenario and we can't change it afterwards. Then what is the point of having this variable when it's not going to be updated?

robes:
@RFSH The Column object is immutable from the perspective of the user. Internally, we may create new instances of Column objects when applicable to different scenarios, rather than carrying along on copy from the first introspection through various contextualized references.

That said... If its not necessary here, we do not need to spend time on it.

RFSH:
@robes I just wanted to make sure `column.position` is useful in general. In this case, it doesn't help that much, so I am not going to add it.

robes:
That’s fine

hongsudt:
@RFSH should this pull request be merged now?

RFSH:
@hongsudt If nobody has an objection yes.
The only important change that this pull request is going to make, is in association definition. As long as in our test cases the association tables don't have any extra columns, it shouldn't change anything.

hongsudt:
let's go ahead and merge it.

Issue #157, opened 2016-08-13
support gene_sequence use case


hongsudt:
This will either be specified through a markdown syntax or a different annotation syntax. We need this to support gpcr and maybe facebase.

hongsudt:
@karlcz @robes I am going to suggest creating a different custom type gene_sequence and get the pretty_print to deal with the formatting.. Any objection?

karlcz:
Should this be able to be composed with markdown features?  I.e. will we ever expect to have a markdown_pattern that expands gense sequence column data into a formatted presentation?

Basically, I am pondering whether gene-sequence should be another special markdown container that we can wrap around the raw sequence string.

howdyjessie:
Do gene sequences always come in increments of 10? If not, what should I do with the remainder?

howdyjessie:
This is currently under review in PR #168.

howdyjessie:
PR #168 has been merged.

PR #168, opened 2016-08-24
Pretty print gene sequences

howdyjessie:
This PR addresses #157, "support gene_sequence use case".

By default, sequences are segmented into groups of 10 characters with a space between each group.

This PR introduces a new pretty pretty utility function (`printGeneSeq`)for columns of `gene_sequence` type, which depends on the `printMarkdown` utility function in order to return the gene sequence in a fixed-width font.


shinyichen:
Looks good to me.

jrchudy:
I don't see any issues. Go ahead and merge at your leisure, @howdyjessie.

PR #173, opened 2016-08-29
Context support in Column-display annotation 

chiragsanghvi:
This PR addresses the context issue in column-display annotation. It handles context-specific formatting now.

Once we merge this PR the record2 test-cases in Chaise will fail. I will fix them after merging this PR.


hongsudt:
It looks fine to me.

PR #174, opened 2016-08-30
Reference canUpdate and canCreate implemented

jrchudy:
This pull request adds some functionality to the reference object as well as the catalog object. These changes are meant to check if a user has the proper permissions to edit or create a record.

`Catalog.meta` was added as a way to get the associated metadata for the catalog object. This request is made as part of the `Catalog.get()` request.

`Reference.session` setter and getters added. The setter was added so that a session object was available in ermrestJS without making another request to the server because the session is grabbed in the UI initially. The getter was required because of the linter. It's a common paradigm to define getters along with setters.

`Reference.canUpdate` and `Reference.canCreate` do the permission checks. Each calls the same function and behaves identically currently, but that can change in the future.


jrchudy:
This PR needs to be closed before chaise [PR #565](https://github.com/informatics-isi-edu/chaise/pull/565) can be merged.

js/core.js
```diff
@@ -177,6 +177,8 @@ var ERMrest = (function (module) {
                 var catalog = new Catalog(self._server, id);
                 catalog._introspect().then(function () {
                     self._catalogs[id] = catalog;
```
robes: Recall that I said you have to defer this assignment until after you are sure everything has been resolved without errors. So you need to move this line down to in between line 181 & 182, just before you resolve the promise.

js/reference.js
```diff
@@ -322,6 +347,47 @@ var ERMrest = (function(module) {
...
+         * This is a private funtion that checks the user permissions for modifying the affiliated entity, record or table
+         * Sets a property on the reference object used by canCreate/canUpdate/canDelete
+         * @memberof ERMrest
+         * @private
+         */
+         _checkPermissions: function (permission) {
+            var editCatalog = false,
+                ignoreRecord = false,
+                ignoreUri = module._annotations.IGNORE,
+                metadata = this._meta[permission],
```
robes: It's confusing to call this variable `metadata`. It would be more accurate to call it `acl`, for access control list.

js/reference.js
```diff
@@ -322,6 +347,47 @@ var ERMrest = (function(module) {
...
+                users = [];
+
+            var ignoreOnTable = this._table.annotations[ignoreUri];
+            var ignoreOnSchema = this._table.schema.annotations[ignoreUri];
+
+            var ignoreTable = (ignoreOnTable !== undefined && (ignoreOnTable === null || ignoreOnTable === true || ignoreOnTable.indexOf('edit') > -1 || ignoreOnTable.indexOf('entry') > -1));
+            var ignoreSchema = (ignoreOnSchema !== undefined && (ignoreOnSchema === null || ignoreOnSchema === true || ignoreOnSchema.indexOf('edit') > -1 || ignoreOnSchema.indexOf('entry') > -1));
+            if (ignoreTable || ignoreSchema) ignoreRecord = true;
```
robes: I don't follow this part. It seems unnecessary to check whether the table or schema are ignored. If they were ignored you probably wouldn't even see them in the first place. I would say we should not mix annotation logic and acl logic together.

js/reference.js
```diff
@@ -166,6 +167,24 @@ var ERMrest = (function(module) {
         },
 
         /**
+         * The session object from the server
+         * @param {Object} session - the session object
+         */
+        set session(session) {
```
howdyjessie: It seems strange to add a useless getter (and then tell people not to use it) to avoid a lint warning. Would it be preferable to tell jshint to ignore this specific warning for this line instead? i.e. 

```
/* jshint -W078 */ //Tells jshint to not lint for warning #078 starting at this line
set session(session) {
    ...
},
/*jshint +W078 */ //Tells jshint to resume linting for warning #078 starting at this line
```

And then remove the getter.

js/reference.js
```diff
@@ -322,6 +347,47 @@ var ERMrest = (function(module) {
...
+            var ignoreOnTable = this._table.annotations[ignoreUri];
+            var ignoreOnSchema = this._table.schema.annotations[ignoreUri];
+
+            var ignoreTable = (ignoreOnTable !== undefined && (ignoreOnTable === null || ignoreOnTable === true || ignoreOnTable.indexOf('edit') > -1 || ignoreOnTable.indexOf('entry') > -1));
+            var ignoreSchema = (ignoreOnSchema !== undefined && (ignoreOnSchema === null || ignoreOnSchema === true || ignoreOnSchema.indexOf('edit') > -1 || ignoreOnSchema.indexOf('entry') > -1));
+            if (ignoreTable || ignoreSchema) ignoreRecord = true;
+
+            for (var i = 0; i < metadata.length; i++) {
+                if (metadata[i] === '*') {
+                    editCatalog = true;
```
robes: `return true;` since don't need further evaluation

test/specs/reference/conf/reference.conf.json
```diff
@@ -1,5 +1,10 @@
 {
-    "catalog": {},
+    "catalog": {
+        "acls": [
```
robes: I'm not sure if its just not visible here but just in case... You'll need tests for various scenarios, and not just the anonymous case.

js/reference.js
jrchudy: Should I be checking for these annotations still in ermrestJS but in the canUpdate/canCreate functions?

jrchudy:
I update the PR with the changes suggested in your comments @rob and @howdyjessie. However, there's a bug with the test suite that doesn't allow you to define a set of ACLs on the catalog object. It currently uses the ACLs that come from dev.isrd.

jrchudy:
This is blocked by #176.

jrchudy:
The tests that fail, because of ACLs not being properly parsed from the test setup file, are marked with `xit` so they won't run. I will fix this when that is resolved. @hongsudt asked that that issue doesn't hold up the PR.

Issue #153, opened 2016-08-12
Support `2016 Table Display: set_presentation


hongsudt:
We should figure out how to use this to support iframe using a markdown syntax specified in the annotation doc.

chiragsanghvi:
@robes @karlcz After having a discussion with @hongsudt this is the API proposition that we came up with.

The reference object will have a `display` property with a getter function as follows.
```js
 /*
  *  An object which contains row display type for this reference. 
  *  Will be populated on basis of  "table-display" annotation. 
  *
  *  The object has following properties
  *  {
  *    type: "markdown",  // Possible values are table/markdown/module(Default is "table")
  *
  *    // If type is "markdown" then you will get these properties 
  *    separator: "\n",  // Default is new line "\n"
  *    suffix: "SOME_MARKDOWN",  //Default is empty string ""
  *    prefix: "SOME_MARKDOWN",  //Default is empty string ""
  *  
  *    // If type is "module" then you will get these properties
  *    modulePath: "pathsuffix"
  *  }
  * ```
  *
  * Usage :
  * ```
  * var displayType = reference.display.type; // the displayType
  *  if ( displayType === 'table') {
  *    // go for default rendering of rows using tuple.values
  * } else if (displayType === 'markdown') {
  *    // Use the separator, suffix and prefix values while rendering tuples
  *    // Tuple will have a "tuple.content" property that will have the actual markdown value
  *    // derived from row_markdown_pattern after templating and rendering markdown
  * } else if (displayType ===  'module') {
  *   // Use modulePath to render the rows
  * }
  * ```
  * @type {Object}
  */  
  get display() {
     // ...
     // ...
  }
```


In addition the `tuple` object will have a `content` property that will contain all the markdown generated as a result of the `row_markdown_pattern` property of the annotation.


Does this works ? Or should we add a `content` property to `tuples` object that will compute the final HTML generated for each tuple based on `row_markdown` and combining the separator, prefix and postfix markdown to be displayed for a page depending on the display type ?

hongsudt:
@chiragsanghvi Let's try to have `Page.display` return the composition of all rows associated with a page.

robes:
This seems fine to me. The only real question I have is what happens when you have Reference to an ERMrest URL using the `attribute` interface rather than an `entity` interface? In that case, we should probably default to display type of `table`, for default rendering of the results, right?

karlcz:
[...]

PR #175, opened 2016-08-30
'tag:isrd.isi.edu,2016:table-display' annotation Implemenation

chiragsanghvi:
This **PR** implements the **tag:isrd.isi.edu,2016:table-display** annotation mentioned in issue https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2015-display and https://github.com/informatics-isi-edu/ermrestjs/issues/153.

In addition it adds other parameters like `row_name` and `row_order` in `reference.display` depending on current context and use them across in `reference.read` and `tuple.displayname`. 

`tuple.displayname` will now generate HTML from markdown in `row_name` if provided or will default.

The **Reference** object will have a `display` property with a getter function. The **Page** object will have a `content` property with a getter function which will compute the HTML based on the markdown parameters.

```js
/* 
 * An object which contains row display type for this reference. 
 * Will be populated on basis of  "table-display" annotation. 
 * The object has following properties
 */
{
    _row_order: [..], // Order of the rows
    _row_name: "", // some row markdown to be used for displayName of tuple
    type: "markdown",  // Possible values are table/markdown/module (Default is "table")
    // If type is "markdown" then you will get these properties 
    _mardkownPattern = "ROW_MARKDOWN",
    _separator: "\n",  // Default is new line "\n"
    _suffix: "SOME_MARKDOWN",  //Default is empty string ""
    _prefix: "SOME_MARKDOWN",  //Default is empty string ""

    // If type is "module" then you will get these properties
    modulePath: "pathsuffix"
}
```
### Usage :
```js
var displayType = reference.display.type; // the displayType
var page = {reference.page} // some page retrieved after reference.read
 if ( displayType === 'table') {
   // go for default rendering of rows using tuple.values
} else if (displayType === 'markdown') {
  /*
   * Page will have a "page.content" property that will have the actual HTML value
   * derived from row_markdown_pattern after templating and rendering markdown
   * for all the tuples  using the separator, suffix and prefix values
   */

   var content = page.content;

} else if (displayType ===  'module') {
  // Use modulePath to render the rows
}
```


chiragsanghvi:
Update `displayname` handling according to issue #179

hongsudt:
@chiragsanghvi we discussed that the pull request should be for a specific purpose. We shouldn't try to merge multiple issues into one request. If the review is too slow, push on the review.

doc/api.md
```diff
@@ -1916,6 +1919,45 @@ the referenced resource(s). In some cases, this permission cannot
...
+ 
+   // If type is "module" then you will get these properties
+   modulePath: "pathsuffix"
+ }
+```
+
+Usage :
+```
+var displayType = reference.display.type; // the displayType
+ if ( displayType === 'table') {
```
hongsudt: should we make these constants e.g. 'table' an enum object e.g. the same way we did with annotation?

doc/api.md
```diff
@@ -1916,6 +1919,45 @@ the referenced resource(s). In some cases, this permission cannot
...
+An object which contains row display type for this reference. 
+ Will be populated on basis of  "table-display" annotation. 
+
+ The object has following properties
+ {
+   
+   rowOrder: [{ column: "NAME", descending: true/false }] || undefined,
+   
+   type: "markdown",  // Possible values are table/markdown/module (Default is "table")
+
```
hongsudt: There is now a `page_size` in the annotation.. maybe @RFSH need to update this in his pull-request?

doc/api.md
chiragsanghvi: We can.

doc/api.md
chiragsanghvi: @hongsudt I kept the code in this pull request because it had changes which were needed and I couldn't proceed without them. 

In addition the 2 issues used the same annotation, and my PR aims at handling the table-display annotation. So I thought that instead of making a new branch I should push the code to the same branch.

chiragsanghvi: @hongsudt The code that @RFSH has written is redoing the stuff that I have written. So he might need to merge his branch with mine or wait until my code gets merged in master.

js/reference.js
```diff
@@ -794,6 +878,64 @@ var ERMrest = (function(module) {
...
+        get content() {
+            if (this._content !== null) {
+                // If display type is markdown which means row_markdown_pattern is set in table-display 
+                if (this._ref.display.type === 'markdown') {
+
+                    // If the number of records are zero then simply return the value
+                    if (!this._data || !this._data.length) {
+                        return "";
+                    }
+
```
hongsudt: Maybe we should return `null` in this case? if there is nothing to display, we need to communicate to the app so this entire related entity can be ignored. @jrchudy I saw (# of records) on the record2 page, so there should be a way to determine this already, correct?

js/reference.js
```diff
@@ -1018,31 +1154,65 @@ var ERMrest = (function(module) {
...
+                // no row_name annotation, use column with title, name, or id:text type
+                else if (typeof this._data.title === 'string') {
+                    col = table.columns.get("title");
+                    this._displayname = col.formatvalue(this._data.title, { context: self._pageRef.context });
+                }
+                else if (typeof this._data.name === 'string') {
+                    col = table.columns.get("name");
+                    this._displayname = col.formatvalue(this._data.name, { context: self._pageRef.context });
+                }
+                else  {
```
hongsudt: I just updated the issue to include `term` which is used frequently for vocabulary tables.

js/reference.js
```diff
@@ -1018,31 +1154,65 @@ var ERMrest = (function(module) {
...
+
+                    // get templated patten after replacing the values using Mustache
+                    var pattern = module._renderTemplate(template, keyValues);
+
+                    // Render markdown content for the pattern
+                    this._displayname = module._formatUtils.printMarkdown(pattern, { inline: true });
+                }
+                // no row_name annotation, use column with title, name, or id:text type
+                else if (typeof this._data.title === 'string') {
+                    col = table.columns.get("title");
```
hongsudt: Is this operation case insensitive? If not, can you make it case-insensitive e.g both column names `Title` and `title` should work.

chiragsanghvi: Yes, I was thinking of making it case-insensitive.

js/reference.js
```diff
@@ -1018,31 +1154,65 @@ var ERMrest = (function(module) {
...
+                    });
+
+                    // If id column exists
+                    if (idCol.length && typeof this._data.id === 'string') {
+                        this._displayname = idCol[0].formatvalue(this._data.id, { context: self._pageRef.context });
+                    } else {
+                        // Get the columns for shortestKey
+                        var keyColumns = table.shortestKey;
+                        var values = [];
+
```
hongsudt: If this consists of all columns, maybe we should return `null` for now. See my comment in the issue.

hongsudt:
@chiragsanghvi I added a few comments.. Apart from that, it looks good to me. I skipped over the testing part though.. @RFSH please review.

js/reference.js
hongsudt: I realized that and updated the issue this evening.

doc/api.md
hongsudt: can you guys communicate? I asked @RFSH to review this pull request also; that should help.

RFSH: @chiragsanghvi @hongsudt I will merge the other PR, after this PR gets merged.

chiragsanghvi:
@hongsudt I have made the changes as you recommended in the comments above.

js/reference.js
```diff
@@ -1175,42 +1175,59 @@ var ERMrest = (function(module) {
...
+                    var setDisplaynameForACol = function(name) {
+                        if (typeof self._data[name] === 'string') {
+                            col = table.columns.get(name);
+                            self._displayname = col.formatvalue(self._data[name], { context: self._pageRef.context });
+                            return true;
+                        }
+                        return false;
+                    };
+
+                    var columns = ['title', 'Title', 'name', 'Name', 'term', 'Term', 'label', 'Label'];
```
hongsudt: I am not sure that we can consider this case-insensitive e.g. if my column is `TITLE` then this won't detect it. This is especially important for the ID column since I have seen `id`, `Id`, and `ID`. 
 
Should we create a tmp_columns that consists of all column names that have been converted to all lower cases and use this set for checking/mapping in the heuristics?

js/reference.js
```diff
@@ -1175,42 +1175,59 @@ var ERMrest = (function(module) {
...
+                        if (setDisplaynameForACol(columns[i])) {
+                            return this._displayname;
+                        }
+                    }
+
 
                     // Check for id column whose type should not be integer or serial
                     var idCol = table.columns.all().filter(function (c) { 
-                        return ((c.name == "id") && (c.type.name.indexOf('serial') === -1) && (c.type.name.indexOf('int') === -1));  
+                        return ((c.name == "id" || c.name == 'Id') && (c.type.name.indexOf('serial') === -1) && (c.type.name.indexOf('int') === -1));
```
hongsudt: `ID` should be included as well.

RFSH:
The changes look fine to me. @hongsudt Should we merge it?

hongsudt:
Yes.

Issue #177, opened 2016-08-31
Support 2016 Table Display : "page_size" 


hongsudt:
I am adding "page_size" to indicate the default page size associated with the page. Please modify the read call associated with the `compact` and `compact/brief` contexts to use it.

RFSH:
Instead of changing the limit in read function implicitly, we decided to just expose the defined "page_size" in `Reference.display._defaultPageSize`.

PR #180, opened 2016-09-01
Reference.read default page size

RFSH:
This PR, adds default value for page size in read function.

If the limit is not defined, it will look for `page_size` in Table Display annotation, using the current context. Otherwise, it will use "1" as a default page size.

Related Issue: #177

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -417,9 +417,20 @@ var ERMrest = (function(module) {
          */
         read: function(limit) {
             try {
-                verify(limit, "'limit' must be specified");
-                verify(typeof(limit) == 'number', "'limit' must be a number");
-                verify(limit > 0, "'limit' must be greater than 0");
+                if (typeof(limit) != 'number' || limit <= 0) {
+                    var content = -1;
+                    // get default page size from annotation
+                    if (this._table.annotations.contains(module._annotations.TABLE_DISPLAY)) {
+                        content = module._getRecursiveAnnotationValue(this._context, this._table.annotations.get(module._annotations.TABLE_DISPLAY).content);
+                    }
+
+                    if (content != -1 && typeof(content.page_size) == 'number') {
+                        limit = content.page_size; // default page_size is defined
+                    } else {
+                        limit = 1; // fixed default limit
+                    }
+                }
+
 
                 var defer = module._q.defer();
 
--- a/test/specs/reference/conf/reference_schema/schema.json
+++ b/test/specs/reference/conf/reference_schema/schema.json
@@ -377,7 +377,10 @@
             ],
             "annotations": {
                 "tag:isrd.isi.edu,2016:table-display": {
-                    "row_order": [{"column":"name", "descending":false}]
+                    "row_order": [{"column":"name", "descending":false}],
+                    "detailed": {
+                        "page_size" : 6
+                    }
                 }
             }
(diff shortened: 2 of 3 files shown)


js/reference.js
```diff
@@ -417,9 +417,20 @@ var ERMrest = (function(module) {
...
+                    var content = -1;
+                    // get default page size from annotation
+                    if (this._table.annotations.contains(module._annotations.TABLE_DISPLAY)) {
+                        content = module._getRecursiveAnnotationValue(this._context, this._table.annotations.get(module._annotations.TABLE_DISPLAY).content);
+                    }
+
+                    if (content != -1 && typeof(content.page_size) == 'number') {
+                        limit = content.page_size; // default page_size is defined
+                    } else {
+                        limit = 1; // fixed default limit
```
hongsudt: @karlcz convinced me that the apps should know how many record it tries to read.. So if there is no limit specified, we should throw an error.

js/reference.js
```diff
@@ -417,9 +417,20 @@ var ERMrest = (function(module) {
...
-                verify(limit > 0, "'limit' must be greater than 0");
+                if (typeof(limit) != 'number' || limit <= 0) {
+                    var content = -1;
+                    // get default page size from annotation
+                    if (this._table.annotations.contains(module._annotations.TABLE_DISPLAY)) {
+                        content = module._getRecursiveAnnotationValue(this._context, this._table.annotations.get(module._annotations.TABLE_DISPLAY).content);
+                    }
+
+                    if (content != -1 && typeof(content.page_size) == 'number') {
+                        limit = content.page_size; // default page_size is defined
```
hongsudt: Should we expose this value as a Reference.display.default_page_size (e.g. make it a getter) as well? That way, this value is available for the app to use. If there is no annotation, we can leave it undefined. The app can then choose to use it or ignore it. E.g. recordset can use this if it exist, or else use its own default.

js/reference.js
RFSH: If limit is not specified, I should use the value that is defined in the annotation and if that is not defined too, then I'll throw an error. Right?

js/reference.js
RFSH: Is the `Reference.display` is part of #175? I can add `default_page_size` when that PR gets merged.

js/reference.js
hongsudt: Yes.

js/reference.js
hongsudt: @RFSH can you review #175, so the code can be merged?

js/reference.js
RFSH: @hongsudt Then what is the point of default_page_size? this whole annotation was for the case that users don't specify anything. If app should explicitly specify the limit, then `page_size` annotation is not needed.

hongsudt: It can be used to inform the app, when it is applicable. For example, the resultset app should honor this default, if exists. On the other hand, the record should always specify a limit of 1 despite this value. 

The main difference is whether we should make the page size explicit or hide some logic from the app. From the past experiences, it seems like making page_size explicit will make it less error-prone.

RFSH:
These changes are not relevant anymore.

We decided to not change the default value explicitly, and just expose the default page size as `reference.display.default_page_size `.

PR #191, opened 2016-09-13
Make defaultPageSize available.

shinyichen:
Made `defaultPageSize` available so that Chaise apps can access it.


jrchudy:
No comments from me. This is a simple change.

Issue #184, opened 2016-09-08
Reference.contextualize.compact behaves oddly


jrchudy:
When I contextualize a reference with the `detailed` context, I can verify that the references are very similar with slight variations. When I use the compact context in `record2`, all of my related references are contextualized to the same reference.

The request for reference.related does in fact return `x` number of distinct references, but once I start contextualizing them, each of my related references are the same. One oddity is that it ends up being the 10th related reference in the set of 27 that are returned, in the case that I'm testing.

RFSH:
`contextualize.detailed` and `contextualize.compact` use the same function and the only difference is the context string they use. So they should behave the same.

Are you sure for compact you're using `compact` and not `compact/brief`? The `contextualize.compact` in reference is using `compact` as the context name but the chaise issue is about `compact/brief` (which is not implemented in reference).

PR #192, opened 2016-09-13
Resolve Issue #184 (Extract Contextualize class)

RFSH:
Related Issue: #184 

**Why**
To create a new copy of `Reference` we use `_referenceCopy`. This function creates a new Reference and copies all the attributes of the current Reference object into the new one. Now if an attribute is an object, it will copy the reference of this object.

In `Reference` we have an object called `contextualize`. Using `_referenceCopy` in `Reference` just copies the reference of `contextualize` attributes. Therefore after
```javascript
var newRef = _referenceCopy(ref);
```
`newRef.contextualize` and `ref.contextualize` point to the same object and changing `newRef.contextualize._reference` will result in changing `ref.contextualize._reference`.

In `Reference.related`I used
```javascript
var newRef = _reference(this);
newRef.contextualize._reference = newRef;
```
Based on what I said, this will change`this.contextualize._reference` too. And since I am doing this for all the related references, all of them will have the same `contextualize._reference`.

**Solution**
The main issue is setting a pointer to `Reference` in `contextualize`. To avoid the problem, I extracted `contextualize` as a separate Class. Now instead of changing attributes of `contextualize`, I just create a new `Contextualize` object (`_referenceCopy` function takes care of this).

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -194,16 +194,12 @@ to use for ERMrest JavaScript agents.
         * [.setBinaryPredicate(colname, operator, value)](#ERMrest.ParsedFilter+setBinaryPredicate)
     * [.Reference](#ERMrest.Reference)
         * [new Reference(location)](#new_ERMrest.Reference_new)
+        * [.contextualize](#ERMrest.Reference+contextualize)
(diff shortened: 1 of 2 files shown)


jrchudy:
No comments from me about the changes.

Issue #190, opened 2016-09-12
Parse URL with related tables


shinyichen:
Right now ermrestjs only parses simple URL with the projection schema and table. Now it needs to handle complex URL with related tables so that when record-two links related tables to recordset, recordset can show the right table.

hongsudt:
This should have been taken care of by Reference.related. Please also take
a look at the alternative tables and app_links before you do any work on
this.

hongsudt:
The current recordset assumes that the table to be displayed is the left most S:T in the url, which is not the case when tables are joined. 

 `S:T1/k1=v1/k2=v2/.../(k1,k2,...)=(S:T2:fk1,fk2,...)` will return S:T2 as the context of the request.

If we split path on "/", then 
- items[0] is the base table 
- items[1]...items[n-2] are filters for key (composite conjunction) 
- items[n-1] is a join to related table where - LHS: columns from base table - RHS: colums from related table where the first column is in the form `S:T:C`, and 2nd column is in the form `C`. 

In this case, ermrest will return the right most table. Ermrestjs needs to set up the right table for the above url.

hongsudt:
@robes should the plan be to check for the pattern above and if so, set the correct table for the reference. We probably need to work on a parser soon, but in the meantime I want the apps to support a simple url.

karlcz:
It's worth noting that the URL structure described above is true only under limiting assumptions for how the record and recordset apps will navigate between each other:

1. The entity reference is always of the form `S:T/k=v`.
2. The related-entities reference is always of the form `S:T/k=v/(k)=(S:T2:fk)` (where the joining columns MAY be other than the entity's key constraint columns in the general case).
3. A single related-entity reference is then of the form `S:T2/k=v`.

We will not build deeply nested joins as we navigate between entity and related-entities references since we can always form a canonical reference using the primary key constraint.

robes:
Right, I think there are these simplifying assumptions based on the URLs that will be generated by the record(-two) and recortset apps.

[...]

PR #193, opened 2016-09-13
Updated URL parser to handle simple entity linking #190

shinyichen:
Changes for #190 
Updated the parser to handle URL with entity linking. This allows record-two to link related tables to recordset.


jrchudy:
I can't say for sure if this is right, but it looks good to me. The test you wrote looks like it is parsing the full path and checking that both parts are parsed as expected.

I think this is good to merge.

Issue #171, opened 2016-08-26
Use attribute group api instead of entity api when making a call to ermrest


hongsudt:
The attribute group api will allow all the foreign keys to be fetched together with the main entity. 
Currently, it is assumed that all foreign keys are not null. @karlcz will extend the API to support null foreign keys. 

The purpose of this issue is to re-factor the ermrest call without impacting the rest of the apps. So `Tuple._data` should remain the same. Let's save other outbound linked data in Tuple._linkedData. 

![img_20160826_143923-1](https://cloud.githubusercontent.com/assets/414732/18022399/61529506-6ba3-11e6-9b50-df5035133b44.jpg)

RFSH:
`Related.read()` uses the url that users specify by `ERMrest.resolve(url)`. So it can be "entity", "attribute", "attributegroup", or "aggregate" ermrest api call ([parser](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/parser.js#L82) allows all these api names). However the current "reference.js", only supports "entity" ermrest api and it doesn't understand the value that other apis return.

Based on these, it was assumed that ermrestjs is only using "entity" api; which is not correct and this decision is made by the user (e.g. chaise apps).

After discussion with @hongsudt, we decided that as a temporary solution for this issue, we change the api to "attributegroup" and ignore the api (service) that user has specified.

PR #199, opened 2016-09-15
Use attribute group api instead of entity api when making a call to ermrest 

RFSH:
Issue: #171 

This PR adds a logic to `Reference.read()` function to change the uri in order to get the foreign key data as well as current reference data. This doesn't affect the actual `Reference._location` object and just changes the uri internally for the http request.

The `Page._data` works exactly as it used to and `Page._linkedData` stores the data for foreign keys. I didn't add any new test cases for `Page._linkedData` because I wasn't sure about the data structure that I used for holding the foreign key data. However, the fact that other test cases are not failing shows that `Page._data` has been set correctly.

The data structure that I used for `Page._linkedData` is as follows:
```javascript
 // for the i-th returned data
 this._linkedData[i] = {
  'schemaName:constraintName': foreignKeyData,
  ...
}
```
```javascript
// for retrieving the foreign key data in Page using this._linkedData
var i, j, fkData, foreignKeys = this._ref._table.foreignKeys.all();
for (i = 0; i < this._data.length; i++) {
  for (j = 0; j < foreignKeys.length; j++) {
    fkData = this._linkedData[foreignKeys[j].constraint_names[0].join(":")];
  }
}
```
Is this good enough? Do I need to change it?

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -425,6 +425,29 @@ var ERMrest = (function(module) {
 
                 var uri = this._location.compactUri;
 
+                // change api to attributegroup
+                if (this._table.foreignKeys.length() > 0) {
+                    var loc = this._location, i;
+
+                    uri = [loc.service, "catalog", loc.catalog, "attributegroup", loc.compactPath].join("/");
+                    // add the M !
+                    //TODO the difficult part
+
+                    // add other alias
+                    var fkList = "";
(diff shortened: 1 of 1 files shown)


RFSH:
I will close this PR, as a new PR will replace it with the complete set of changes related to foreign keys.

PR #208, opened 2016-09-22
Script load race

chiragsanghvi:
This **PR** fixes Chaise issue https://github.com/informatics-isi-edu/chaise/issues/612.

I've added `_onLoad` function which is called by http class. It resolves promises by calling this function to make sure thirdparty scripts are loaded.


robes:
@chiragsanghvi What I can't tell from looking at the PR is whether the _scriptsLoaded flag actually indicates that the scripts have been loaded or just that our code has injected the `<script>` tags into the html. Do we _actually_ know that the scripts have been _loaded_?

Issue #170, opened 2016-08-24
Support 2016 Table Alternatives


hongsudt:
This is needed to allow a view to be displayed in recordset, but when clicking on an individual record, it should lead to a record in an actual table. 

We could extend contextualize to take this annotation into account and set the Reference.uri correctly. @robes thoughts?

robes:
Yes, but...

How does one tell the difference between a 'base entity storage table' and a table that abstracts it? I think things are getting messy with all of these annotations and we are missing some core concepts to simplify our data services.

hongsudt:
From chaise perspective, there is no difference, except through the annotation. And Views seems to be what data modelers use to work-around frontend limitation.  

If there is a way to aggregate selected related tables into the compact view of the main entity, maybe this will reduce the use of view?

karlcz:
I think this was one of the first drafts introducing the 'context'
concept and it looks a little rough because of it.

We could also turn it around and make the annotation on the base
storage table to list which abstracted view to prefer for each
context.  This would probably be easier to maintain as a DBA and less
chance for ambiguity.  We could also use the same name-pairs notation
we use for constraints now:


```
table.annotations['tag:isrd.isi.edu,2016:table-alternatives'] = {
  context1: [ "Schema1", "View1" ],
  context2: [ "Schema1", "VIew2" ],
  ...
};
```

We'd still declare the same model requirements: the alternative table
or view should have a (psuedo) primary key that is a foreign key to
the base storage table.  The base storage table is the one bearing
this annotation.

hongsudt:
What about this strategy?

* Create a Table._baseTable. If it is undefined, the baseTable is the same as the table itself.

* Modify Tuple to support
  - Add Tuple.getReference() which returns an uncontextualized reference to get to the tuple.
  - override the above reference.read() to use the original tuple._data instead of going to the server to get the same data.
  
* Modify contextualize to take into account the alternative table and base table.
  - get the baseTable
  - add a getter to retrieve alternative table (._alternativeTable) for each context using the baseTable. Only the base tables should have this annotation!
  - if alternativeTable is defined, create a contextualize reference to the alternativeTable. Otherwise, the contextualization is done at the same reference.

* To support the app_links, add Reference.appLocation object 

From recordset:
```
resolve -> Reference (R0)
  - R0.contextualize.compact -> Reference (R1)
    - R1.read -> Page (P0)
      - For each P0.tuple (T)
        - T.getReference (R2)
[...]

PR #211, opened 2016-09-26
Alternative Table & App Linking

shinyichen:
PR for issue #170: Alternative tables and app linking
Task-breakdown:

- [ ] Create a Table._baseTable during the schema introspection taking  `tag:isrd.isi.edu,2016:table-alternatives` annotation into account. If there is no annotation, the base table points to its own table. This needs to be done after all the tables have been created and all the foreign keys are built. All constraints above should be checked.

- [ ] check table alternative constraints. If fail, ignore the annotation.  
  - [ ] The base table and all participating alternative tables must have not-null unique constraints. 
  - [ ] Each alternative table must have exactly one not-null unique key which is a foreign key to the base table. This implies that an alternative table has exactly one base-table. (Alternative table can have more than one key, but only one that map to the base table. For our current use case, we are allowing exactly one foreign key).
  - [ ] All alternative tables associated with the base table have not-null unique keys that are foreign keys to the SAME primary keys of the base table. We could do this by iterate through all not-null unique keys of the base-table, and check whether all alternative tables have foreign keys pointing to it.  
  - [ ] There is no in-bound foreign keys to the alternative tables. 
  - [ ] a base table cannot be an alternative table for other base tables (i.e. flat 2-level forest) e.g. a base table cannot have ._baseTable that is not itself. 

- [ ] Modify Tuple by:
  - [ ] adding Tuple.reference() which returns an un-contextualized reference to get to the entity represented by the tuple. The reference url should be of the form `basetable/key=value`. We already have this getter, but the context is set to its page.reference.context. Need to modify this to support base table.  
  - [ ] overriding the above reference.read() to use the original tuple._data instead of going to the server to get the same data (if applicable).

- [ ] Modify contextualize to take into account the alternative table and base table.
  - [ ] add a getter to retrieve alternative table (._alternativeTable) for each context using the baseTable. Only the base tables should have this annotation!
  - [ ] If alternativeTable is defined, create a contextualize reference to the alternativeTable. Otherwise, the contextualization is done at the base table.
  - [ ] to support the `tag:isrd.isi.edu,2016:app-links` annotation, add Reference.appLocation object 
  - [ ] When contextualizing a reference, if ._table._baseTable is not ._table (this is an alternative table), use the base table to get the alternative table of that context.


js/core.js
```diff
@@ -271,11 +271,11 @@ var ERMrest = (function (module) {
                 // find alternative tables
                 // requires foreign keys built
                 for (s = 0; s < schemaNames.length; s++) {
-                    schema = self.schemas.get(schemaNames[s]);
-                    tables = schema.tables.names();
-                    for (t = 0; t < tables.length; t++) {
-                        table = schema.tables.get(tables[t]);
-                        table._findAlternatives();
+                    var schema1 = self.schemas.get(schemaNames[s]);
```
chiragsanghvi: Instead of renaming the variables, you can simply declare them before both the for loops start. For example, you could have declared `schema`, `tables` and `table` initially and then used it inside the for loop.

```js
var schema, table, tables
```

chiragsanghvi: Same goes for other cases in this PR too.

shinyichen:
Just made major update to this branch. @robes @hongsudt would you check if it looks ok? Test should cover all the possible cases now.

js/parser.js
```diff
@@ -231,7 +231,11 @@ var ERMrest = (function(module) {
             // columns in the filters are on the projection table
             // use the mapped columns to replace column name to the linked table's column names
             if (colMapping) {
-                _replaceFilterColumns(this._filter, colMapping);
+                var res = _replaceFilterColumns(this._filter, colMapping);
+                if (res === -1) {
+                    console.log("WARNING: Unable to convert filter.");
```
robes: this error message would be more informative if you would append to it the offending filter.

js/reference.js
```diff
@@ -827,8 +827,11 @@ var ERMrest = (function(module) {
             return this._related;
         },
 
-        getAppLink: function(context) {
-            return this._table.getAppLink(context);
+        getAppLink: function() {
```
robes: I would prefer this be a getter defined as:
```
get appLink() {
 ...
}
```

js/reference.js
```diff
@@ -915,85 +918,154 @@ var ERMrest = (function(module) {
...
-                newRef.setNewTable(altTable);
+                // swap to new table
+                newRef.setNewTable(newTable);
 
+                // case 2: no filter
                 if (source._location.filter === undefined) {
                     // case 1: no filter
                     newRef._location = module._parse(source._location.service + "/catalog/" + source._location.catalog + "/" +
-                            source._location.api + "/" + newRef._table.schema.name + ":" + newRef._table.name);
+                            source._location.api + "/" + newTable.schema.name + ":" + newTable.name);
```
robes: do any of these `name`s need encoding?

js/reference.js
```diff
@@ -915,85 +918,154 @@ var ERMrest = (function(module) {
...
+                            if ((source._table.isAlternativeTable() && filter.column === source._table.altForeignKey.colset.columns[0].name) ||
+                                (!source._table.isAlternativeTable() && filter.column === sharedKey.colset.columns[0].name)) {
+
+                                if (newTable.isAlternativeTable()) // to alternative table
+                                    filterString = newTable.altForeignKey.colset.columns[0].name + "=" + filter.value;
+                                else // to base table
+                                    filterString = sharedKey.colset.columns[0].name + "=" + filter.value;
                             }
+
+                            newLocationString = source._location.service + "/catalog/" + source._location.catalog + "/" +
```
robes: or any of these need encoding?

shinyichen:
I made changes according to all the comments except encoding/decoding  (is on a separate issue).

Issue #197, opened 2016-09-15
implement Reference.create


hongsudt:

robes:
Josh and I reviewed this in detail. For starters we'll have the create(...) function take an array of objects which should correspond to the columns _not_ set by defaults.

robes:
It just occurred to me... @jrchudy Maybe we should extend that create signature to take the list of defaults as an array of column definitions. That way its up to the client to determine which columns they wish to use defaults for.

jrchudy:
But aren't the defaults defined by which columns are not showing (i.e. the visible_columns annotation) and any columns that have a `type.name` property set to `serial`?

jrchudy:
Or should we allow a column to be a default and shown in the UI for input. If they don't define a value for that input, use the default instead? 

That's the only case I can think of where the user would want to alter the defaults that are being used.

karlcz:
Almost any field could support a "use default" decision by the user. Serials are just the most obvious case where they'd want to and where we might not even want to let them override the default behavior.

Whether the UX is to interpret blanks as "use default" or have something else like a checkbox, this decision is not something you can solely predict from the model content.  Therefore, the ermrestjs API needs a way to configure different possible requests.

jrchudy:
PR #212 merged. Closing this issue.

PR #212, opened 2016-09-27
Reference create

jrchudy:
Reference.create implementation is complete. I wrote a number of unit tests for checking the contextualize functions as well as a create function that makes sure the data is posted to the database properly.

This issue resolves issue #197.


js/reference.js
```diff
@@ -346,21 +346,82 @@ var ERMrest = (function(module) {
...
-                // TODO
                 //  verify: data is not null, data has non empty tuple set
+                verify(data, "'data' must be specified");
+                verify(data.length > 0, "'data' must have at least one row to create");
+
+                var defer = module._q.defer();
+
                 //  get the defaults list for the referenced relation's table
-                //  get the data
+                var defaults = getDefaults();
```
robes: I thought we decided that the client should decide what defaults to use? If you implement it like this, its not possible for a client to set the properties for those fields for which there are defaults. Is this consistent with our UX?

js/reference.js
```diff
@@ -346,21 +346,82 @@ var ERMrest = (function(module) {
...
+                });
+
+                return defer.promise;
             }
             catch (e) {
                 return module._q.reject(e);
             }
+
+            function getDefaults() {
+                var defaults = module._columnDiff(self._table.columns._columns, self.columns);
```
robes: So 'defaults' are defined by the {table columns} - {reference columns}. Is there some heuristic about why? I thought that 'defaults' are something that are in the table specification in ermrest.

test/specs/reference/tests/01.reference.js
```diff
@@ -8,20 +8,52 @@ exports.execute = function (options) {
...
 
-        var multipleEntityUri = options.url + "/catalog/" + catalog_id + "/entity/" + schemaName + ":"
-            + tableName + "/id::gt::" + lowerLimit + "&id::lt::" + upperLimit;
+        var singleEntityUri = baseUri + "/id=" + entityId;
+
+        var multipleEntityUri = baseUri + "/id::gt::" + lowerLimit + "&id::lt::" + upperLimit;
 
 
         // Test Cases:
+        describe('for a single entity,', function () {
```
robes: How about tests 'for multiple entities'?

test/support/single.spec.js
```diff
@@ -1,6 +1,6 @@
 require('./../utils/starter.spec.js').runTests({
     description: 'In print utilities, ',
```
robes: 'In print utilities'?

js/reference.js
```diff
@@ -346,21 +346,82 @@ var ERMrest = (function(module) {
...
-                //  make a page of tuples of the results (unless error)
-                //  new page will have a new reference (uri that filters on a disjunction of ids of these tuples)
-                //  resolve the promise, passing back the page
-                notimplemented();
+                module._http.post(uri, data).then(function(response) {
+                    //  new page will have a new reference (uri that filters on a disjunction of ids of these tuples)
+                    var uri = self._location.compactUri + '/',
+                        keyName;
+
+                    self._shortestKey.push(self._table.columns.get("c_int"));
```
robes: what is "c_int"?

test/specs/reference/tests/07.contextualize.js
```diff
@@ -0,0 +1,103 @@
...
+            // The new reference should be different from the original (a deep copy)
+            expect(ref).not.toBe(reference);
+
+            // Some values should be the same as the original
+            expect(ref.uri).toBe(reference.uri);
+            expect(ref._location.service).toBe(reference._location.service);
+            expect(ref._table).toBe(reference._table);
+        }
+
+        /** Test case Contexts:
```
robes: Is this still useful or was this just "note to self" that can now be removed?

js/reference.js
jrchudy: Oops, that was left over from me testing when the shortest key includes multiple columns.

test/specs/reference/tests/01.reference.js
jrchudy: This implementation is currently tested and working for single entities while I make sure the UI can handle that properly. I can wait to merge this until I clean up `recordedit` and include multiple entities. (It might already work for multiple entities)

test/support/single.spec.js
jrchudy: This is just a file used in testing so we can run 1 of the specs instead of the whole suite.

test/specs/reference/tests/07.contextualize.js
jrchudy: Ah yes, that was a note to self to remove.

js/reference.js
jrchudy: I think we need to meet and discuss this tomorrow with @hongsudt and @karlcz. Our UX doesn't have a way for the user to state what values need to be set by default and what ones they want to override. This becomes more complex when there are multiple entities being posted and you want `entity 1` to have `col A` set by default and want to define `col A` for `entity 2`.

js/reference.js
jrchudy: The defaults that are set on each column were part of the discussion @hongsudt and I had today. Like the above comment, I think we should all sync up tomorrow about this implementation.

test/support/single.spec.js
robes: So the plan here is:
- mv `single.spec.js` to `single.spec.js.sample`
- add `test/support/*.spec.js` to the `.gitignore`

jrchudy:
I addressed @robes's comments and am merging this PR.

Issue #218, opened 2016-09-28
markdown handling


hongsudt:
By default, unless the markdown syntax is guarded (e.g. the syntax contains `{{#.*}}` or `{{^.*}}`, if any of the variable in the markdown syntax is null, treat the result as null. 

The guarded syntax can be used to change the default behavior above.

PR #220, opened 2016-09-29
Change template rendering according to issue #218

chiragsanghvi:
Now the `_renderTemplate` function checks for existence of `{{#.*}}` or `{{^.*}}` strings in the template. If they're not present, then it explicitly checks for null values in the template, returning a null.

### DIFF
--- a/test/specs/print_utils/tests/01.print_utils.js
+++ b/test/specs/print_utils/tests/01.print_utils.js
@@ -136,9 +136,10 @@ exports.execute = function (options) {
         it('module._renderTemplate() should function correctly', function() {
             expect(module._renderTemplate("My name is {{name}}", {name: 'John'})).toBe("My name is John");
             expect(module._renderTemplate("My name is {{name}}", { name: null })).toBe(null);
+            expect(module._renderTemplate("My name is {{name}}", {})).toBe(null);
             expect(module._renderTemplate("My name is {{#name}}{{name}}{{/name}}", {})).toBe("My name is ");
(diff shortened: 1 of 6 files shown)


test/specs/print_utils/tests/01.print_utils.js
```diff
@@ -132,5 +133,13 @@ exports.execute = function (options) {
             }
         });
 
+        it('module._renderTemplate() should function correctly', function() {
+            expect(module._renderTemplate("My name is {{name}}", {name: 'John'})).toBe("My name is John");
+            expect(module._renderTemplate("My name is {{name}}", { name: null })).toBe(null);
```
robes: I feel like a test case might be missing here. What about when the environment is `{}` and there is no guard?

`expect(module._renderTemplate("My name is {{name}}", {})).toBe(null);` ???

chiragsanghvi: @robes I will add that testcase. Though the output for above template will be `My name is ` instead of null. This is because the object that we are passing is empty. Mustache replaces null values with `''` (empty string). So it won't return null.

If you want then I can handle this case, to check all properties in the template for their existence in the object and null value.

chiragsanghvi:
@robes  Made changes as requested. Now the template renderer checks for null values as well as for non-existing placeholders.

For example:
```js
module._renderTemplate("My name is {{name}}", {})
```
will return null, as `name` is not passed in the object.

chiragsanghvi:
@robes @hongsudt Should I merge this.

robes:
yes

Issue #224, opened 2016-10-03
Support all columns in the template 


hongsudt:
Currently, the template only allow referencing to the visible column only. We need to allow the annotation to access the formatted value of any columns in the Tuple. 

An option for this is to create another structure e.g `Tuple._data.formatted` that has the same access pattern as `._data` and then the `.value` can just build its value from this structure. 

@chiragsanghvi can you work on this, as I want @RFSH to focus on getting the foreign key feature done. We can converge on the solution tomorrow.

PR #225, opened 2016-10-03
Fixed issue #224

chiragsanghvi:
@hongsudt  Currently, the template only allow referencing to the visible column only. We need to allow the annotation to access the formatted value of any columns in the Tuple.

Instead of referring the `ref.colums`, I am referring the `ref._table.columns` to find the column and format it accordingly.

### DIFF
--- a/js/utilities.js
+++ b/js/utilities.js
@@ -269,12 +269,17 @@ var ERMrest = (function(module) {
     module._getFormattedKeyValues = function(ref, data) {
         var keyValues = {};
 
-        for (var i = 0; i < ref.columns.length; i++) {
-            var col = ref.columns[i];
-            keyValues[col.name] = col.formatvalue(data[col.name], { context: ref._context });
-
+        for (var k in data) {
+            
+            try {
+                var col = ref._table.columns.get(k);
+                keyValues[k] = col.formatvalue(data[k], { context: ref._context });
+            } catch(e) {
+                keyValues[k] = data[k];
+            }
+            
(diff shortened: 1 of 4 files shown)


karlcz:
On Oct 03, Chirag Sanghvi modulated:
> @hongsudt Currently, the template only allow referencing to the visible
> column only. We need to allow the annotation to access the formatted
> value of any columns in the Tuple.
> 
> Instead of referring the ref.colums, I am referring the
> ref._table.columns to find the column and format it accordingly.
> 

This sounds about right to me. Just to be sure: the mustache template
should also be referencing the columns by their real names in the
ERMrest table definition, not influenced by any column renaming done
for display purposes. Right?

robes:
I wonder if there is some additional complexity in this. Like if we are going to use formatted values as the input to markdown, do we need a dependency graph just to determine what order the formatting of values should be processed in?

chiragsanghvi:
@robes Formatting is totally dependent on the type of the column. The `Tuple.values()` getter first gets the formatted values and then starts the templating with markdown following it. So the order wouldn't matter. https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/reference.js#L1261

karlcz:
@robes I think it should be in two phases:

1. pre-formatting, which is single-column only and does not include markdown
2. mustache templates, which can sample all pre-formatted values

there should **not** be any recursion of mustache templates. each one
should see the exact same input environment regardless of which order
they are evaluated.

robes:
@chiragsanghvi @karlcz makes sense.

PR #227, opened 2016-10-06
Allow setting tag to app URL mapping.

shinyichen:
Changes required for https://github.com/informatics-isi-edu/chaise/issues/644
Let chaise set the app-tag to app-url mapping function.

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -837,14 +837,12 @@ var ERMrest = (function(module) {
         },
 
         get appLink() {
-            if (this._context) {
-                var tag = this._table._getAppLink(this._context);
-                if (tag && module._appLinkFn) {
-                    return module._appLinkFn(tag, this._location);
-                }
-            }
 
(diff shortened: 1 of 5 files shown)


js/reference.js
```diff
@@ -828,10 +837,14 @@ var ERMrest = (function(module) {
...
-            else
-                return undefined; // no context
+            if (this._context) {
+                var tag = this._table._getAppLink(this._context);
+                if (tag && module._appLinkFn) {
+                    return module._appLinkFn(tag, this._location);
+                }
+            }
+
+            return undefined; // reference has no context or app link not specified by annotation
```
RFSH: I think we should use the "default" context (if it's available) even if the reference is not contextualized i.e., `this._context` is undefined; instead of returning undefined.

If the default context was not available then we can return undefined.

shinyichen: Ya, I need to add the default case. Thanks.

PR #232, opened 2016-10-07
More encoding and decoding fixes.

shinyichen:
More encoding and decoding fixes. Updated test to use special characters.


RFSH:
These are other lines of codes that I think should use encoding:
  - reference.js: [954](https://github.com/informatics-isi-edu/ermrestjs/blob/encode-decode/js/reference.js#L954,L955), [979](https://github.com/informatics-isi-edu/ermrestjs/blob/encode-decode/js/reference.js#L979), [981](https://github.com/informatics-isi-edu/ermrestjs/blob/encode-decode/js/reference.js#L981), [984](https://github.com/informatics-isi-edu/ermrestjs/blob/encode-decode/js/reference.js#L984,L986), [1051](https://github.com/informatics-isi-edu/ermrestjs/blob/encode-decode/js/reference.js#L1051,L1053)

shinyichen:
It's ready for review now.

js/reference.js
```diff
@@ -1051,8 +1052,8 @@ var ERMrest = (function(module) {
                                         filterString += (j === 0? "" : "&") + mapping[f.column] + "=" + f.value;
```
jrchudy: I think we need to encode `mapping[f.column]` and `f.value` as well.

PR #236, opened 2016-10-12
Improve shortest key algorithm using column names.

shinyichen:
If two keys are equal length and both are the same type, use column names to determine order.


hongsudt:
what if they are both text type?

shinyichen:
my description was wrong.

Issue #230, opened 2016-10-07
a function Column.inputDisabled that incorporate generated and immutable annotation


hongsudt:
provide a function `Column.getInputDisabled(context)` that takes the context as a parameter and depending on the context, return InputDisabled value based on generated and immutable annotation. 

Potential return values:
* `true` : do not allow user input 
* `false` : allow user input 
* `{message:"the message to show in place of the column value"}` : do not allow user input and the `message` should be shown in-place of the column value.  

Rules for setting the value: 
* When the context is create
  * if the column is generated, then inputDisabled={message:"automatically generated by the server"}
* when the context is update
  * if the column is generated, then inputDisabled=true
  * if the column is immutable, then inputDisabled=true

informatics-isi-edu/chaise#432 depends on this function.

howdyjessie:
PR #237 has been merged; closing now.

PR #237, opened 2016-10-13
Implement Column.getInputDisabled()

howdyjessie:
This PR addresses #230.

It adds a function onto the `Column` class called `getInputDisabled`, which determines whether a column instance is editable in RecordEdit based on the current context and the presence of `2016:immutable` and `2016:generated` annotations on that column.

Assigned to @shinyichen but also tagging @jrchudy.


shinyichen:
Everything looks fine. But did you delete the single test runner sample?

howdyjessie:
@shinyichen Sorry I forgot to add it back in; it's back in now!

Issue #221, opened 2016-09-30
implement Reference.update


hongsudt:

hongsudt:
The current version has a limitation that when the unique columns are updated, it will always throw a 409 conflict e.g. `Error 409 Conflict The request conflicts with the state of the server. Input data violates model. ERROR: duplicate key value violates unique constraint "experiment_pkey" DETAIL: Key (id)=(2) already exists.`

@jrchudy please make sure that the new version will support this case. The attribute group API needs to be used for this.

robes:
it will

hongsudt:
Make sure to use etag pre-condition in put and delete.

jrchudy:
This is resolved by PR #238.

PR #238, opened 2016-10-13
Reference update

jrchudy:
This PR includes the necessary changes for `Reference.update`. This function needs to handle updating a single entity as well as updating multiple entities. There are a number of tests that outline updating a single entity. These tests include changing 1 column and multiple columns where those columns cover a range of different cases.

Tests include modifying:
 - an independent key that is the shortest key
 - an independent key that is not the shortest key
 - a column that is part of 1 composite key
 - a column that is part of multiple composite keys
 - a column that is not a key column at all

The tests that include changing multiple columns exhaustively test the combination of the above circumstances.

This PR resolves issue #221. This can be tested from my public directory on `dev.isrd.isi.edu` where I have also made the corresponding changes to Chaise to use the Reference APIs (https://dev.isrd.isi.edu/~jchudy/chaise/recordedit/  I will let you fill out the rest of the url for an entity you want to test against).

Assigning to @howdyjessie (I talked to her about the implementation on a few different days already) and tagging @shinyichen, @robes, and @RFSH.


js/reference.js
```diff
@@ -621,12 +622,113 @@ var ERMrest = (function(module) {
...
+
+                shortestKeyNames = this._shortestKey.map(function (column) {
+                    return column.name;
+                });
+
+                tuples.forEach(function (tuple) {
+                    newData = tuple.data;
+                    oldData = tuple._oldData;
+                    for (var key in newData) {
+                        // if the key is part of the shortest key for the entity, the data needs to be aliased
```
robes: Don't we need to alias all columns so that we avoid possible name collisions? Say an the table definition includes columns named: `filename` and `old_filename` or `new_filename` and they use it as a key.

I would like to see your test schema include key and non-key columns with the `old_` and `new_` prefixes in them.

jrchudy: I addressed in the most recent commit. I am now aliasing columns with "_o" for anything that is represented for the shortest key, and "_n" for all columns that are being updated.

js/reference.js
```diff
@@ -621,12 +626,114 @@ var ERMrest = (function(module) {
...
+
+                // The list of column names to use in the uri
+                columnProjections = this.columns.map(function (col) {
+                    return col.name;
+                });
+
+                // copy all of the tuple update data to the submission data array. Need to submit an array of objects instead of an array of Tuples
+                // tuples.forEach(function(tuple) {
+                //     submissionData.push(tuple.data);
+                // });
```
howdyjessie: Commented out = should be deleted?

howdyjessie:
Looks good to me

PR #239, opened 2016-10-17
Reference delete

jrchudy:
This PR addresses deleting entities via the Reference.delete API. This is currently only an implementation for ermrestJS.

Assigning to @shinyichen and mentioning @howdyjessie to look at it.


howdyjessie:
Looks good to me too :)

PR #243, opened 2016-10-19
Changed tuple getter values to return raw data for edit context

chiragsanghvi:
This **PR** changes `tuple.values` getter to return raw values when the context is "entry/edit".


howdyjessie:
@chiragsanghvi If you could address the current merge conflicts as well as Josh's comment, then this PR should be good to go.

PR #244, opened 2016-10-19
Support Search

shinyichen:
* Reference: Apply search word(s) on a Reference returns a new Reference with the search filter.
* Parser: parse search filter in the URI and add that to the location object.


shinyichen:
Note: Now with search filter in the URL, parser now assumes search filter is at the end of the path (after join). In @RFSH Aref's PR he assumes join in at the end of the path (as it should be before this PR). Just a reminder that has to be updated.

js/reference.js
```diff
@@ -973,6 +973,30 @@ var ERMrest = (function(module) {
...
+        /**
+         * create a new reference with the new search
+         * by copying this reference and clears previous search filters
+         * search term can be:
+         * a) A string with no space: single term or regular expression
+         * b) A single term with space using ""
+         * c) use space for conjunction of terms
+         * @param {string} term - search term, undefined to clear search
+         */
+        search: function(term) {
```
RFSH: Should use `verify` to make sure `term` is as described. At least we should check if it's an string or not; it will throw an error in `term.trim()` if it's not an string.

js/parser.js
```diff
@@ -351,6 +369,74 @@ var ERMrest = (function(module) {
...
+
+                terms.forEach(function(t, index, array) {
+                    filterString += (index === 0? "" : "&") + "*::ciregexp::" + module._fixedEncodeURIComponent(t);
+                });
+
+            } else {
+                this._searchTerm = null;
+            }
+
+            if (this._searchFilter && term) { // replace search filter
```
RFSH: These if statements are a bit confusing.

```javascript
if (this._searchFilter) {
    if (filterString) { // replace search filter

    } else { // remove search filter 

    }
} else if (filterString) { // add search filter

}  
```

This looks easier to understand to me.

RFSH: On a sidenote, do we really need to keep all these `_uri`, `_compactUri` and `_compactPath`?

shinyichen: You mean in location?

js/reference.js
shinyichen: I am adding the check for string only. There is not much to verify for a string.

shinyichen:
Made the changes according to your suggestion.

PR #253, opened 2016-10-28
Add origFK to related references and change contextualize.entryEdit

RFSH:
In this PR, 
- To be able to figure out which foreign key created the related reference, `origFKR` was added.
- Also in `contextualize.entryEdit`, if the reference is actually an association related reference, we contextualize based on the association reference; rather than the current reference.

Assigning to @howdyjessie, since she will use these values in chaise to create the "Add" button in related references.


howdyjessie:
I tested it and it works fine; the code also looks good to me, so I'm merging now.

PR #260, opened 2016-10-29
Reference.update and Reference.create fixed to work with pseudocolumns

jrchudy:
This PR addresses the changes to `reference.columns` to include pseudocolumns that replace the originating column. Update only required a minor change. 

Create needed some changes to the way the defaults are generated. All of the visible columns are now sent as part of the payload from chaise to ermrestJS. This means that values that were not set will have `null` as their value. The defaults function looks through this set and figures out if that column should be included in the list of defaults that ermrest should try to populate.

Assigning to @shinyichen. Mentioning @howdyjessie and @RFSH if that want to look at it as well.

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -481,9 +481,12 @@ var ERMrest = (function(module) {
                      * If not set in any row, add it to defaults.
                      * If 1 row has it set and none of the others, it cannot be part of defaults
                     **/
-                    data.forEach(function (row) {
-                        if (row[columnName]) notSet = false;
-                    });
+                    for (var m = 0; m < data.length; m++) {
+                        if (data[m][columnName]) {
+                            notSet = false;
+                            break;
+                        }
+                    }
 
                     if (notSet) defaults.push(columnName);
                 });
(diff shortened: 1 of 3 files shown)


js/reference.js
```diff
@@ -468,18 +465,48 @@ var ERMrest = (function(module) {
...
+                // loop through the data and add any columns not set to the defaults
+                visibleColumnsNames.forEach(function (columnName) {
+                    var notSet = true;
+                    /**
+                     * Loop through the rows and make sure the current column is not set in any of the rows.
+                     * If not set in any row, add it to defaults.
+                     * If 1 row has it set and none of the others, it cannot be part of defaults
+                    **/
+                    data.forEach(function (row) {
+                        if (row[columnName]) notSet = false;
```
RFSH: This `forEach` will run for all the rows in data, even if it matches the first one. 

Change it to [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) or `for` loop with `break`; so that it will return the `notSet` value as soon as it find one row with `colName`.

Issue #263, opened 2016-11-01
Reference.read() add alias needs to account for search filter in the URL


shinyichen:
This piece of code was written while search capability was being written. Therefore it did not account for the URL having a search filter at the end. 
https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/reference.js#L618-L623

hongsudt:
@shinyichen Do we need to support this for the first round of recordedit?

shinyichen:
It's a very small change. Shouldn't doesn't affect anything right now because recordedit or record don't have search filter.

RFSH:
@shinyichen Does it cause any problems? Urls such as [this](https://dev.isrd.isi.edu/chaise/recordset/#1/legacy:dataset/owner=Axel%20Visel/*::ciregexp::new@sort(id)?limit=15) will work which has both search and filter.

PR #274, opened 2016-11-04
Account for search filter when appending alias to URI

shinyichen:
Address issue #263.

Account for the search filter in the URI when appending alias.


hongsudt:
@RFSH please review the pull request.

RFSH:
Looks good to me. This will also fix [chaise#699](https://github.com/informatics-isi-edu/chaise/issues/696).

Issue #247, opened 2016-10-26
Change Outbound Foreignkey Heuristics


RFSH:
Currently we apply [heuristics for outbound foreignkeys](https://github.com/informatics-isi-edu/ermrestjs/blob/master/doc/heuristics.md#outbound-references) when [`visible-columns` annotation](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2016-visible-columns) is not defined on the Table. We should apply the heuristics even if this annotation is defined on the Table.

PR #277, opened 2016-11-10
Outbound Foreign Key Heuristics

RFSH:
This PR addresses #247.

Merging this PR will break some of the chaise test cases in record as we're validating the value of some foreign key columns.

### DIFF
--- a/js/core.js
+++ b/js/core.js
@@ -1464,6 +1464,8 @@ var ERMrest = (function (module) {
         this._columns = [];
 
         this._table = table;
+
+        this._contextualize_cached = {}; // cache the result for the cases that origFKR is null
     }
 
     Columns.prototype = {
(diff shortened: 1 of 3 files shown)


js/core.js
```diff
@@ -1539,80 +1541,115 @@ var ERMrest = (function (module) {
...
-                                colFound = true;
+                    }
+                } else { // columns
+
+                    // (annotation: column name)
+                    if (typeof col == "string" && !(col in consideredColumns)) {
+                        // find the corresponding column
+                        for (j = 0; j < this.length(); j++) {
+                            if (this.getByPosition(j).name == col) {
+                                col = this.getByPosition(j);
```
shinyichen: Could you have done this with this.get(col) instead of the loop? Just to simplify the code.

shinyichen:
The update looks fine to me.

PR #283, opened 2016-11-15
Expand association table linkage to create and entry contexts

RFSH:
Related Issue in chaise: [chaise-768](https://github.com/informatics-isi-edu/chaise/issues/768).

Currently in `contextualize.entryEdit`, if the reference is actually an association related reference, we contextualize based on the association reference; rather than the current reference. 

Since chaise is using `contextualize.entryCreate` instead, this PR will expand this behaviour to all the entry contexts (`contextualize.entryEdit`, `contextualize.entryCreate`, and `contextualize.entry`).


howdyjessie:
Looks good to me, merging now.

Issue #278, opened 2016-11-14
Allow generated annotation at table-level and schema-level


hongsudt:
Allow generated annotation at table-level and schema-level and expose as generated property on table, disabling creation and edition of new records in GUI.

@robes @karlcz should we add this logic into canCreate, canEdit, and canDelete? I think Chaise already checks this functions before showing the creation/deletion button.

PR #284, opened 2016-11-15
Generated schema and table 

shinyichen:
Address issue #278 
When table or schema has annotation `generated`, permission for create/update/delete should be false.


RFSH:
Looks good to me, I am merging it.

PR #286, opened 2016-11-16
Updated utility function _generateRowName to always return HTML format.

shinyichen:
PR required for Chaise bug https://github.com/informatics-isi-edu/chaise/issues/769
Updated utility function _generateRowName to always return HTML format.

### DIFF
--- a/js/utilities.js
+++ b/js/utilities.js
@@ -300,11 +300,11 @@ var ERMrest = (function(module) {
      * @param {ERMrest.Table} table The table that we want the row name for.
      * @param {String} context Current context.
      * @param {object} data The object which contains key value pairs of data.
-     * @return {object} The display name for the row.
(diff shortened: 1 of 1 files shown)


js/utilities.js
```diff
@@ -300,85 +300,97 @@ var ERMrest = (function(module) {
...
+                    /*
+                     * join all values by ':' to get the display_name
+                     * Eg: displayName for values=["12", "DNA results for human specimen"] would be
+                     * "12:DNA results for human specimen"
+                     */
+                    result = values.join(':');
+                }
+            }
+
+            template = "{{name}}";
```
chiragsanghvi: @shinyichen You might want to change this to `template = {{{name}}}`. The name shouldn't be escaped.

RFSH:
Apart from the comment that @chiragsanghvi made, the changes look good to me.

Issue #305, opened 2016-12-16
tuple.value expose raw values for both edit and create


hongsudt:
tuple.values function to contextualize the values for both edit and create the same way.

PR #306, opened 2016-12-19
tuple.value expose raw values for all entry contexts

RFSH:
Related Issue: #305


hongsudt:
Aref, please make sure to update your test cases as well.

RFSH:
@hongsudt I updated the test cases for create context.

PR #307, opened 2016-12-19
Added default property to referenceColumn

jrchudy:
We want to be able to set default values in recordedit when creating entities. `Column.default` is already a property, so I'm just exposing the default value for the `referenceColumn` object.


RFSH:
@jrchudy What should this default value be in case of composite foreign keys?

jrchudy:
@RFSH, I don't think composite foreign keys should ever have a default value define in the model for them.

RFSH:
@jrchudy So shouldn't the `default` be undefined in those cases?
```javascript
 if (typeof foreignKey === 'undefined' || !foreignKey.simple) {
  this.default = column.default;
}
```

Issue #275, opened 2016-11-08
supporting non-lexical sorting (e.g. allow sorting to be done based on a different columns)


hongsudt:
For example, in RBK and Facebase, there is a column called `age stage` that has values like `3.5 weeks, 10 weeks, etc.` When this column is sorted, it needs to be sorted based on their ordinal ranking e.g. 3.5 comes before 10 weeks. 

@karlcz @robes should we add an annotation to specify the mapping columns to be used for sorting? 

Another related issue.. since each column can be annotated with a markdown.. How do we deal with sorting for those columns. Pesudo-columns have the same issue..

karlcz:
We already have a row_order annotation which could be applied on a domain table.  So maybe what you're asking for is a heuristic to assume that sorting by foreign keys should actually sort by the row_order of the referenced domain table instead of the actual foreign key value?

You would need to form a joined entity path and include the other column in the projection so that you can ask for sort based on that column.  I.e. if your age stage column is a foreign key to an age-stage domain table which includes a ranking column, you need to include this domain table in the join and project the ranking column and mention the ranking column in the sort modifier.

ERMrest only allows sorting by **output** columns in the projection.

karlcz:
After further discussion, I agree that the case with markdown_pattern creates a problem. We cannot realistically determine a semantic sort order for columns where the presentation has been remapped by an arbitrary markdown_pattern expansion.

So, we probably need an annotation to suggest a "column_order" using the same array of sort keys as in the "row_order" annotation option.  With this, the app could apply sort when the user requests sorting on a column. Without this, the app should probably disable sorting on columns that have been transformed by markdown_pattern expansion.

For psuedo columns, I think we could inherit the row_order from the linked domain table instead. We might also allow the column_order annotation on a foreign key in order to explicitly override the generic domain table row_order?

karlcz:
I pushed [column_order annotation options](https://github.com/informatics-isi-edu/ermrest/commit/171f81912ce463e408f305361f4724bfe9d0bba9?short_path=cf73c73#diff-cf73c73d214853f687df4473ad3e234a)

Issue #246, opened 2016-10-21
Sort Reference with PseudoColumn


RFSH:
Currently we are using `Column.name` for sorting the reference. In case of `PseudoColumn`, the `name` property will not return a proper value and therefore we cannot continue using that. We should find a better solution on how to sort reference using `PseudoColumn`.

PR #312, opened 2016-12-22
supporting non-lexical sorting (e.g. allow sorting to be done based on a different columns)

RFSH:
This PR adds `column_order` annotation support for [column](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2016-column-display), [foreignkey](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2016-foreign-key), and [key](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2017-key-display).

It also includes adding `origColumnName` to the related tables, which chaise is will use for prefill. Previously chaise was using `origFKR` to find the pseudo-column, while it could cause some bugs (the pseudo-column name is not necessarily the constraint-name of foreign key).

Related Issues: #246 , #275, #326 

**This PR will break the chaise testcases. [chaise#972](https://github.com/informatics-isi-edu/chaise/pull/972) contains the required fixes.**


js/reference.js
```diff
@@ -881,32 +939,47 @@ var ERMrest = (function(module) {
...
+                        sortCols = sortObject.map(function(sort) {return sort.column;});
+                    }
 
+                    addedCols = {};
+                    for(k = 0; k < sortCols.length; k++) {
+                        if (sortCols[k] in sortMap) {
+                            // sort column has been created via PseudoColumn, so we should use a new alias
+                            sortColumn = module._fixedEncodeURIComponent(sortCols[k]) + ":=" + sortMap[sortCols[k]];
+                        } else {
+                            sortColumn = module._fixedEncodeURIComponent(sortCols[k]);
```
RFSH: It is column name (string).

RFSH:
@jrchudy I made some changes to the PR:
- Fix the typo in function name.
- Fix the issue with encoding while sorting.
- Add some test cases for encoding.

PR #315, opened 2017-01-05
Unalias keys in page returned from reference.update

jrchudy:
The update function handles updating multiple records at a time already. To work with this in the UI, we needed to make sure that the page returned from `reference.update()` had the keys in the data properly named. They were being returned as the aliased column name rather than the column name, so the resultset view for `recordedit` was getting `undefined` for values in `page.tuple.values`.

These changes should fix that. Rather than using the response data from the ermrest request to generate the Page object, I slice the alias off the end of the keys returned (keys are the aliased column names) from ermrest and create a new object to put into the data array used to create a Page.


jrchudy:
This PR is ready for review again. I made changes to the tests to verify that `page.data` does indeed have the submitted data in it when ermrest returns it from the put request.

PR #317, opened 2017-01-07
fixes #311 stringify objects

### LINKED ISSUE #311: Stringify JSON for presentation
This seems like something we can handle in ermrestjs rather than Chaise. We need to `JSON.stringify(...)` objects. This is probably something to handle in Column format values function for columns of type `json`/`jsonb`. 

If this isn't a trivial fix somewhere in that code, then we should have some further discussion.

This issue is related to #146

robes:
This just addressed the issue that when `printText` encounters an Object it uses JSON.stringify() to produce the text rather than object.toString().

### DIFF
--- a/js/utilities.js
+++ b/js/utilities.js
@@ -599,6 +599,9 @@ var ERMrest = (function(module) {
             if (value === null) {
                 return '';
             }
+            if (typeof value === 'object') {
+                return JSON.stringify(value);
+            }
             return value.toString();
         },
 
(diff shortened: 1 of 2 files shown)


jrchudy:
Changes don't effect anything negatively. I looked up the difference between `.stringify` and `.toString` and we absolutely want to avoid `.toString` unless we define a `.toString` method for that object.

Merging this PR.

Issue #320, opened 2017-01-19
Add get reference to table


jrchudy:
`Table` should have a `get reference` method to get a reference for that whole table. 

This is to address [this chaise issue](https://github.com/informatics-isi-edu/chaise/issues/932). This way the apps can get a reference for the table before generating an app link to get a more appropriate link.

PR #321, opened 2017-01-20
Table reference

jrchudy:
This PR addresses issue #320 which is in relation to [chaise issue #932](https://github.com/informatics-isi-edu/chaise/issues/932). There are 3 changes included here:
 - `Table` has a `get reference` function to get a reference to that specific table. `Table` has a private function for getting the `uri` representing it used for generating the reference.
 - `Reference` exposes it's `table` object via a public `get table` function.
 - When contextualizing a reference, the filter is no longer removed when contextualizing to the create case.

This PR must be merged before [this PR](https://github.com/informatics-isi-edu/chaise/pull/938) in chaise is merged.


RFSH:
@jrchudy Please pull the master so the checks pass, then we can merge.

Issue #310, opened 2016-12-21
automatically create linkable column value for unique key constraint


hongsudt:
Ermrestjs should automatically create a linkable value for a unique key constraint, similar to what we have done for the foreign key constraint.
In other words, we want to add new `ReferenceColumn` object to the list of columns that is returned by `Reference.columns`. These new columns will represent a key. 

### How to choose a key
- If `visible-columns` annotation was defined, we should create "linkable column" for all the keys that are in the `visible-columns` annotation. 
  - Duplicate or invalid keys will be ignored.
  - If a key has columns that are html (`markdown` typed or has `markdown_pattern` annotation), instead of creating a `PseudoColumn`, add its constituent columns.
  - If a key has columns that has `nullok=true`, add its constituent columns.
  - In entry contexts (`entry/edit`, `entry/create`, and `entry`), instead of creating a `PseudoColumn` for key, add its constituent columns.
- Otherwise, we should choose the key that (in any context other than entry contexts):
  1. None of its columns are html (`markdown` typed or has `markdown_pattern` annotation).
  2. All of its columns are not null.
  3. Has shorter length than other candidates; i.e. prefer simple key to composite key.
  4. Has better presentational type (e.g. prefer `text` to `serial`).

If we couldn't find a key with these descriptions, we should not create a `ReferenceColumn` for that.

### Attributes
The instructions on how to derive this new column attributes:

- comment: underlying key's comment.
- displayname: key columns display names separated with space.
- formatPresentation (value of column): 
  - If any of the key columns are html (markdown type or has markdown_pattren in column-display) should return "null value".
  - Otherwise, use key columns values separated with space for caption. The URL should refer to the current reference.
- name: `constraint name` of the key (same as foreignkey).
- type: markdown.

RFSH:
@hongsudt @karlcz Should we add this "linkable column" in all the contexts? I don't think we should add them in edit and create context.

karlcz:
I agree they don't make sense in edit or create.

hongsudt:
You are right.. No linkage in the create/edit context. In edit/create
context, they should be treated as other regular columns.

PR #322, opened 2017-01-23
Linkable key

RFSH:
Related Issue: #310 

[chaise#944](https://github.com/informatics-isi-edu/chaise/pull/944) PR should be merged first.


js/reference.js
```diff
@@ -323,16 +324,42 @@ var ERMrest = (function(module) {
...
+                                                        this._referenceColumns.push(new ReferenceColumn(this, cols[j]));
+                                                    }
+                                                }
+                                            } else {
+                                                this._referenceColumns.push(new ReferenceColumn(this, (fk.simple ? fk.colset.columns[0] : null), {"key":fk}));
+                                            }
+                                        }
+                                        break;
+                                    default:
+                                        // unknown behaviour, should not happen
```
jrchudy: Have this case throw an error. Ideally there would be error handling where `fk.subject` is set to avoid this from occurring. Better to have this error out than to fail silently.

js/reference.js
```diff
@@ -2066,7 +2108,11 @@ var ERMrest = (function(module) {
                     for (i = 0; i < this._pageRef.columns.length; i++) {
                         column = this._pageRef.columns[i];
                         if (column.isPseudo) {
-                            presenation = column.formatPresentation(this._linkedData[column._constraintName], {context: this._pageRef._context});
+                            if (column._isKey) {
+                                presenation = column.formatPresentation(this._data, {context: this._pageRef._context});
```
jrchudy: nitpick: `presentation` is misspelled..

js/reference.js
RFSH: The whole logic in the `visible-columns` is to verify and ignore the incorrect one. I changed the comment to reflect this. Instead of saying "unknown behaviour", I will mention that only `key`, `foreignkey` and `columns` are supported and acceptable in the annotation.

PR #327, opened 2017-01-26
Handle 412 Precondition Failed errors

howdyjessie:
This PR partially address informatics-isi-edu/chaise#846 (see corresponding Chaise PR # ...). If Reference.delete() it encounters a 412 Precondition Failed error, the function will check to see if the referenced entity still exists in the DB. If it does, it deletes the entity with a new ETag; otherwise, it passes the error to Chaise like normal.


howdyjessie:
After discussing this issue more in standup, it seems like a larger change and further work is needed to address 412s. Closing this PR for now.

PR #328, opened 2017-01-26
Change displayname

RFSH:
This PR will turn the displayname of `Column`, `Table`, `Schema`, and `Reference` into an object with `isHTML` and `value` attributes.

Merging this PR will break the chaise test-cases. So it should be merged alongside the required changes in chaise.


howdyjessie:
Once work on informatics-isi-edu/chaise#952 has been completed, I will review this and merge both PRs.

howdyjessie:
Is it possible for a tuple displayname to be HTML? If so, tuple.displayname should also be an object with isHTML and value properties.

howdyjessie:
Actually there's no need to answer my previous question. Tuple.displayname is always run through a Markdown parser before its final value is returned, so there's no need for an isHTML flag.

Issue #348, opened 2017-02-09
Tests for searching on ints and floats.


jrchudy:
Changes were made to `location.search` to search using a regular expression for integers and floats. Test cases need to be written for this feature.

PR #350, opened 2017-02-10
tests for the search parser to test integer and float search

jrchudy:
This PR adds the tests that should have been committed with PR #347. I also had to modify the regular expression that is being used for searching in SQL as the previous one did not omit results such as `201` when searching for the integer `1`.

I also added a new schema and changed the search spec to use that new schema instead so it doesn't conflict with any other tests. This will allow for more search tests to be written without affecting other tests that require certain values to maintain a specific order.

This PR resolves issue #348.


howdyjessie:
Hongsuda asked me to review this PR and it looks good to me so I'll be merging now...

Issue #354, opened 2017-02-15
remove serial column from default visible list unless it is used as the unique key


hongsudt:
This only applies to the heuristics for determining visible columns.

PR #355, opened 2017-02-16
update columns heuristics

RFSH:
This PR will change columns heuristics so that serial columns that are part of a simple key, will not be added to list of visible columns (unless they are used for self-link).

Related Issue: #354


RFSH:
I updated the PR to include changes mentioned in #357.

**The chaise test cases are still pending.**

js/reference.js
```diff
@@ -414,6 +417,12 @@ var ERMrest = (function(module) {
                         if (col.name in consideredColumns  || hideColumn(col)) {
                             continue;
                         }
+
+                        // if column is serial and part of a simple key
+                        if (col.type.name.toUpperCase().startsWith("SERIAL") &&
+                            col.memberOfKeys.length === 1 && col.memberOfKeys[0].simple) {
+                            continue;
```
jrchudy: Curious what the point of checking this is and having it call `continue`?

RFSH: We don't want to add "serial columns that are part of a simple key" to list of visible columns. This if statement just takes care of that. If it's serial and part of a simple key, we just ignore it.

Issue #329, opened 2017-01-26
Handling 412 Precondition Failed error


howdyjessie:
cc @hongsudt @jrchudy @robes @karlcz @RFSH @chiragsanghvi 

There are 2 places in ermrestjs that currently use the If-Match request config which could run into a 412 Precondition Failed error if ETags are mismatching:

*Reference.update*
- Already accepts `tuples` as an argument so it already has Old Data (data before changes were made) and New Data (the data that's about to be submitted to DB).
- If update encounters a 412, do a Reference.read to get the tuples of the Current Data (data currently in DB). Compare the Current Data with New Data. If they match, just resolve the promise as usual (there's nothing to update). If they don't match, return a 412 error with the Current Data.
- If a rejected promise with a 412 is sent to Chaise, what should I do with the 412 error and Current Data in Chaise?
    - Show a modal with options to Overwrite and Cancel (and will Cancel refresh the page?)?
    - Try to point out where the Current Data differs from New Data in the RecordEdit UI?
    - Something else?

*Reference.delete*
- Doesn't accept any arguments so a tuples arg should be added here to access tuples.data
- As in update case, we do a Reference.read to get Current Data and compare it with tuples.data. If they match, delete the row and resolve the promise. If not, reject the promise with a 412.
- If a rejected promise with a 412 is sent to Chaise, do we 
    - Show a modal that tells the user something is out of sync with the server and then refresh the page once the user clicks the 1 affirmative button?
    - Show a modal with 2 buttons, Delete (lets the user delete anyway) and Cancel? What does Cancel button do — refresh the page?
    - Something else?

Thanks @RFSH for helping me organize my thoughts on this.

karlcz:
For update the ermrestjs flow should be:

1. Update called with tuples having old and new data
2. PUT request sent with If-Match header and ETag from *old* data
3. 412 error received
4. GET request sent to get current data
5. Compare *old* and *current* data
  - If *old* and *current* are same, go back to step 1 but start using ETag from *current* data
  - If *old* and *current* are different, raise 412 with *old*, *current*, and *new* so client application can decide what to do

howdyjessie:
A sketch from @karlcz on how to deal with 412 errors in ermrestjs:

![](https://cloud.githubusercontent.com/assets/4585431/23228069/19f54304-f8f0-11e6-9ce2-cf69a68afba8.JPG)

howdyjessie:
PR #362 has been merged, closing now.

PR #362, opened 2017-03-01
Handle 412 error on delete

howdyjessie:
This PR updates how Reference.delete should behave in the event of a 412 error. It implements the logic described in #329 (see the DELETE half of photo-diagram in the thread).

In Reference.delete:
- Do a Reference.read to get current data. If the read fails, retry the read.
- If current data is empty, then the rows aren't in the DB anyway -> resolve Reference.delete.
- If current data and old data match, run Reference.delete again.
- If current data and old data do **not** match, raise 412.
- Retry this 412 error workflow a max of 3 times. If no more tries are left, raise 412.

Also
- The etag property has been moved from Reference to Page
- The Tuple class now has a `page` property which stores the Page that a tuple came from.

**This PR should be merged with informatics-isi-edu/chaise#1006.**


js/reference.js
```diff
@@ -1164,26 +1171,89 @@ var ERMrest = (function(module) {
...
+                                        mismatchFound = Object.keys(oldTuple).some(findMismatch);
+                                        if (mismatchFound) {
+                                            break;
+                                        }
+                                    }
+                                }
+                                // If old data matches current data, retry the delete w/ updated ETag
+                                if (!mismatchFound && numTriesLeft > 0) {
+                                    numTriesLeft--;
+                                    return ref.delete(tuples);
```
jrchudy: Should this be `currentData`? If not, it doesn't look like you update the `ETag` on `tuples` here before trying to delete again.

js/reference.js
```diff
@@ -1164,26 +1171,89 @@ var ERMrest = (function(module) {
...
                 this._server._http.delete(this.uri, config).then(function deleteReference(response) {
-
                     defer.resolve();
                 }, function error(response) {
-                    var error = module._responseToError(response);
-                    return defer.reject(error);
+                    var status = response.status || response.statusCode;
+                    // If 412 Precondition Failed it means that ETags don't match
+                    if (status == 412) {
+                        var MAX_TRIES = 3, numTriesLeft = MAX_TRIES;
```
jrchudy: I see how the retries are limited by this counter that you decrement. Should there be a delay on this after each retry?

js/reference.js
howdyjessie: Yes 🙈 I forgot to re-update this when the etag was moved from Reference. I've updated this to be current data.

js/reference.js
howdyjessie: I think that's a good idea. Similar to the retry logic for 500s, I've added a delay that doubles after each retry.

Issue #344, opened 2017-02-08
support domain_filter_pattern in foreign-key annoation


hongsudt:
The annotation should support both specific filtering criteria and the current form values to be used in search filtering. We probably need to discuss the detail in a meeting..

PR #365, opened 2017-03-14
Domain filter pattern support

jrchudy:
This PR addresses issue #344. This PR adds a function to the reference object, `filteredRef`, that returns a reference based on a defined `domain-filter-pattern` for the `foreign-key` annotation or the original reference if there is no such annotation defined. This PR needs to be merged before [this PR](https://github.com/informatics-isi-edu/chaise/pull/1020) in `chaise`.


js/reference.js
```diff
@@ -1516,6 +1516,30 @@ var ERMrest = (function(module) {
...
         /**
+         * This function takes in a tuple and generates a reference that is
+         * constrained based on the domain_filter_pattern annotation. If this
+         * annotation doesn't exist, it returns this (reference)
+         * `this` is the same as column.reference
+         * @param {ERMrest.ReferenceColumn} column - column that `this` is based on
+         * @param {Object} data - tuple data with potential constraints
+         * @returns {ERMrest.Reference} the constrained reference
+         */
+        filteredRef: function(column, data) {
```
RFSH: `filteredRef` should be a property of [`ReferenceColumn`](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/reference.js#L2579) not `Reference`.

js/reference.js
```diff
@@ -1516,6 +1516,30 @@ var ERMrest = (function(module) {
...
+            var filteredRef,
+                uri = this.uri;
+
+            if (column.foreignKey.annotations.contains(module._annotations.FOREIGN_KEY)){
+                var filterPattern = column.foreignKey.annotations.get(module._annotations.FOREIGN_KEY).content.domain_filter_pattern;
+                var uriFilter = module._renderTemplate(filterPattern, data);
+                // NOTE: should we check for (uriFilter.trim() !== '') ?
```
RFSH: From what I remember we don't need to verify the filter. @hongsudt Do we need to have any sanity checks for the `domain_filter_pattern`?

jrchudy:
There are test cases for this in `chaise`. Because it relies on data being defined for the `domain-filter-pattern` to work correctly, I figured it would be more useful to test the functionality from the UI. 

I could write tests for this in `ermrestJS`, but I imagine they would be a near duplicate of the functionality in the tests in `chaise`. The only difference I can imagine is that the test in `ermrestJS` would involve me checking the returned `uri` rather than reading from that set.

As I'm typing this, I realize it wouldn't hurt to have tests for this in `ermrestJS`. I will add these in the morning tomorrow.

js/reference.js
jrchudy: I don't think `module._renderTemplate` can return anything but the defined string from the filter pattern with the variables replaced with values or `null`. But I wanted to leave that comment there and find out because I wasn't sure.

js/reference.js
jrchudy: You are correct. I will move this.

jrchudy:
I updated the PR with a change that moved filterRef to the `ReferenceColumn` object. I also added a  set of tests to verify that the function behaves properly.

RFSH:
Thanks for the changes. It's ready to merge.

PR #368, opened 2017-03-21
Fix API Documentation Related to Errors

RFSH:
Based on what we discussed and Rob documented on [ERMrestJS Dev Guide](https://github.com/informatics-isi-edu/ermrestjs/wiki/ERMrestJS-Dev-Guide), we should make sure that in ERMrestJS Asynchronous APIs only communicate errors via promise. Also the synchronous APIs should only communicate errors via throwing exception.

Based on my review, we followed this rule in our APIs. In this PR, I just changed their documentation to make it more clear that Asynchronous APIs are always returning promises.


jrchudy:
[This line](https://github.com/informatics-isi-edu/ermrestjs/blob/err-handling/js/parser.js#L128) and [this one](https://github.com/informatics-isi-edu/ermrestjs/blob/err-handling/js/parser.js#L134) should throw a `MalformedURIError`.

howdyjessie:
Since [`isUnique()`](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/reference.js#L509-L514) isn't really implemented, should it use `notimplemented()` (which throws an error) instead of `undefined`?

Same with [these `Tuple` methods](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/reference.js#L2229-L2268)?

RFSH:
I addressed the comments and updated the branch with master.

PR #369, opened 2017-03-23
Page size increase after paging back to first page

jrchudy:
This PR addresses [this issue in chaise](https://github.com/informatics-isi-edu/chaise/issues/999). 

The implementation follows @karlcz's comments in the issue. Added a relevant test to the paging spec. Other tests are present there as well.

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -1020,6 +1020,21 @@ var ERMrest = (function(module) {
                     }
                     var page = new Page(ownReference, etag, response.data, hasPrevious, hasNext);
 
+                    // we are paging based on @before (user navigated backwards in the set of data)
+                    // AND there is less data than limit implies (beginning of set) OR we got the right set of data (tuples.length == pageLimit) but there's no previous set (beginning of set)
+                    if ( (ownReference.location.pagingObject && ownReference.location.pagingObject.before) && (page.tuples.length < limit || !page.hasPrevious) ) {
+                        // should I clone the reference then set the paging object to null?
+                        ownReference.location.pagingObject = null;
+                        ownReference.read().then(function rereadReference(page) {
+                            defer.resolve(page);
+                        }, function error(response) {
+                            var error = module._responseToError(response);
+                            return defer.reject(error);
+                        }).catch(function (error) {
+                            return defer.reject(error);
+                        });
+                    }
+
                     defer.resolve(page);
 
                 }, function error(response) {
--- a/test/specs/reference/tests/04.paging.js
+++ b/test/specs/reference/tests/04.paging.js
@@ -68,7 +68,6 @@ exports.execute = function (options) {
                 it('next should return a Reference object that is defined.', function() {
                     reference2 = page1.next;
                     expect(reference2).toEqual(jasmine.any(Object));
-
                 });
 
                 it('read should return a Page object that is defined.', function(done) {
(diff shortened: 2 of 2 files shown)


js/reference.js
```diff
@@ -1020,6 +1020,21 @@ var ERMrest = (function(module) {
                     }
                     var page = new Page(ownReference, etag, response.data, hasPrevious, hasNext);
 
+                    // we are paging based on @before (user navigated backwards in the set of data)
+                    // AND there is less data than limit implies (beginning of set) OR we got the right set of data (tuples.length == pageLimit) but there's no previous set (beginning of set)
+                    if ( (ownReference.location.pagingObject && ownReference.location.pagingObject.before) && (page.tuples.length < limit || !page.hasPrevious) ) {
+                        // should I clone the reference then set the paging object to null?
```
RFSH: Yes, we should not mutate the reference object.

js/reference.js
```diff
@@ -1020,6 +1020,21 @@ var ERMrest = (function(module) {
...
+                        ownReference.location.pagingObject = null;
+                        ownReference.read().then(function rereadReference(page) {
```
RFSH: I think we should pass `limit` to this read too.

js/reference.js
```diff
@@ -1020,6 +1020,21 @@ var ERMrest = (function(module) {
...
+                            defer.resolve(page);
+                        }, function error(response) {
+                            var error = module._responseToError(response);
+                            return defer.reject(error);
+                        }).catch(function (error) {
```
RFSH: Now that we are working on error handling, I don't think this catch would be needed. The [outer catch](https://github.com/informatics-isi-edu/ermrestjs/blob/96a5d11da44c7cc1acc08b46feff015c4f45522e/js/reference.js#L1043) is doing the same.

test/specs/reference/tests/04.paging.js
```diff
@@ -139,17 +138,60 @@ exports.execute = function (options) {
                         expect(tuples[i]._data["value x"]).toBeLessThan(tuples[i+1]._data["value x"]);
                     }
                 });
+
+                // limit was changed after paging back
+                it("read with an increased limit should return a Page object that is defined and doesn't have a previous page.", function(done) {
```
RFSH: Although these are testing the scenario that you led to finding the bug; they don't look like unit tests.
You could simply test this by directly calling read on a reference with before options, when
- Number of returned entities are less than limit.
- It's the beginning of the set.

js/reference.js
jrchudy: You are correct, I forgot to include it.

js/reference.js
```diff
@@ -1020,13 +1020,29 @@ var ERMrest = (function(module) {
...
+                        // set the private values to null because the public functions rely on these
+                        referenceWithoutPaging.location._pagingObject = null;
+                        referenceWithoutPaging.location._paging = null;
+
+                        referenceWithoutPaging.read(limit).then(function rereadReference(rereadPage) {
+                            defer.resolve(rereadPage);
+                        }, function error(response) {
+                            var error = module._responseToError(response);
+                            return defer.reject(error);
+                        }).catch(function (error) {
```
RFSH: Is this `catch` statement needed here?

jrchudy: no it isn't

PR #373, opened 2017-03-28
Shortest key keyset

jrchudy:
This PR makes changes to `Reference.update`. Recently we removed ETag checking for `PUT` requests. To account for data change conflicts we started to use the full list of columns and data for finding the row. This caused precision errors for when timestamp columns and float columns were present in this keyset.

This PR changes the implementation for `Reference.update` back to using the shortest key for the keyset. This simplifies the `GET` request to find the row that needs to be updated. The data submitted for update has also been reduced. Only the data that is part of the visible column list and that is not `generated` is submitted in the `PUT` request to `ermrest`.

This PR fixes issues [#1035](https://github.com/informatics-isi-edu/chaise/issues/1035) and [#1029](https://github.com/informatics-isi-edu/chaise/issues/1029). 

I made sure the chaise tests pass as well still.


js/reference.js
```diff
@@ -1114,30 +1114,80 @@ var ERMrest = (function(module) {
...
+                        if (this.columns[m].isPseudo) {
+                            var keyColumns = [];
+
+                            if (column._isKey) {
+                                keyColumns = column.key.colset.columns;
+                            } else if (column._isForeignKey) {
+                                keyColumns =  column.foreignKey.colset.columns;
+                            }
+
+                            for (var n = 0; n < keyColumns.length; n++) {
```
RFSH: This seems wrong. This whole loop is equivalent of `key = keyColumns[keyColumns.length-1].name` which is wrong. You are ignoring other constituent columns of pseudoColumn.

js/reference.js
```diff
@@ -1114,30 +1114,80 @@ var ERMrest = (function(module) {
...
+                        var column = this.columns[m];
+                        var key;
+
+                        // if the column is disabled (generated or immutable), continue
+                        // no need to submit update data in the submission object
+                        if (column.inputDisabled) {
+                            continue;
+                        }
+
+                        if (this.columns[m].isPseudo) {
```
RFSH: Since you are going to change the code based on my comment below, change `this.columns[m]` to `column`. To make it consistent with the previous statement. (This is not a very important change but can help the readability of the code)

RFSH:
It seems like the code is buggy regarding `pseudocolumns`. Can you please fix that? Also having some test cases for tables with simple and composite foreign key would help.

js/reference.js
```diff
@@ -1114,30 +1114,92 @@ var ERMrest = (function(module) {
...
+                            } else if (column._isForeignKey) {
+                                keyColumns =  column.foreignKey.colset.columns;
+                            }
+
+                            for (var n = 0; n < keyColumns.length; n++) {
+                                var referenceColumn = keyColumns[n];
+                                keyColumnName = referenceColumn.name;
+
+                                // here so each column of the columns in keyColumns set is added
+                                if (columnProjections.indexOf(keyColumnName) === -1) {
```
RFSH: These two if statements are duplicate of two if statements after the else. You can create a function to take care of these.

```javascript
var addProjectionAndKeyData = function(index, colName) {
  // add columnProjection
  if (columnProjections.indexOf(colName) === -1) {
    // the list of column names to use in the uri
    columnProjections.push(colName);
  }

  // if the current keyColumnName is in shortestKeyNames, we already added the data to submissionData
  if (shortestKeyNames.indexOf(colName) === -1) {
    // alias all data to prevent aliasing data to the same name as another column that exists in the table
    submissionData[index][colName + newAlias] = newData[colName];
  }
}
```
Then in code
```javascript
if (column.isPseudo) {
  //..
  for () {
    addProjectionAndKeyData(i, keyColumns[n].name);
  }
} else {
  addProjectionAndKeyData(i, column.name);
}
````

You might need to change the function parameters.

Issue #372, opened 2017-03-28
Add Unformatted Attribute to Displayname


RFSH:
Currently the `Reference`, `Tuple`, `Schema`, `Table`, and `Column`have a displayname attribute which has `value` and `isHTML` in it. We should add a new attribute to these displayname objects that will return the unformatted (without any HTML tags) value. If `isHTML` is false, the `value` and `unformatted` value should be the same.

hongsudt:
@RFSH are you working on this issue or is this one done? Please update the issue status..

RFSH:
Yes, It's done as part of #376. I am closing this issue.

PR #376, opened 2017-04-04
Related filter

### LINKED ISSUE #367: supply a filtering criteria for related table reference
We have decided that instead of rewriting the url for the related table references which only works for the direct in-bound foreign keys, but not the pure and binary association case, we will provide the filtering clue through a URL parameter.. 

What about using `filters` as the parameter name? 

Chaise will then just display what's in the parameter value and provides a reset button..

RFSH:
This issue mainly resolves #367. It will add a query parameter (`subset`) to the uri in case of related entities. Then in chaise this query parameter will be used to show additional information.

Other changes:
- Add `unformatted` attribute to displayname objects ( #372 ).
- Change `appLink` logic to throw an error if `_appLinkFn` is not defined ( #373  ).


js/parser.js
```diff
@@ -369,6 +379,18 @@ var ERMrest = (function(module) {
         },
 
         /**
+         * A dictionary of available query parameters
+         * @returns {Object}
+         */
+        get qureyParams() {
```
jrchudy: typo

PR #390, opened 2017-04-19
Html errors

jrchudy:
This PR addresses [issue #1036](https://github.com/informatics-isi-edu/chaise/issues/1036) reported in chaise.

It was decided that when an HTTP error is returned with `content-type` including HTML, this is an error generated from the server (Apache) rather than our code. `Ermrest` does not produce HTML errors, only JSON errors.

In the case of an HTML error being returned, it had to be because of some server error that `Ermrest` did not catch. This is seen as a server error and changed to a `503 Service Unavailable` error.


jrchudy:
Updated the code to throw a `500 Internal Server Error` instead of a `503 Service Unavailable`.

Issue #379, opened 2017-04-06
incorporate file asset annotation


hongsudt:
- incorporate the new asset annotation
- generate a new type of pseudo column 
  - In the read mode, display the column value
  - In the create/edit mode, all the columns (except the url) mentioned in the annotation should not be editable

@karlcz @robes anything else?

robes:
In read mode we could have a prettier default display:
- generate markdown to display it as a link
- the link should use the `download` class
- the name of the link should conditionally be the asset `filename` if one is specified or the URL itself if not filename is specified.

I think that translate into something like this in markdown but we could do it programmatically in the pseudocolumn:
```
[{{{#filename_column}}}{{{filename_column}}}{{{/filename_column}}}{{{^filename_column}}}{{{this column}}}{{{/filename_column}}}]({{{this column}}}){download}
```

In create/edit mode, yes, I think you're right about that.

RFSH:
```md
[{{{#filename_column}}}{{{filename_column}}}{{{/filename_column}}}{{{^filename_column}}}{{{this column}}}{{{/filename_column}}}]({{{this column}}}){download}
```

PR #400, opened 2017-05-03
Asset Pseudo Column

RFSH:
This PR addresses #379. It includes:

- Refactor `ReferenceColumn` implementation to have a base table that other pusedo-column classes (`KeyPseudoColumn`, `ForeignKeyPseudoColumn`, and `AssetPseudoColumn`) will extend it.
- `AssetPseudoColumn` implementation. Its specific properties are:
  - `useDefault`: It's mainly used internally, it only indicates whether all of other properties are null or not.
  - `filenameColumn`: `Column` object or null
  - `byteCountColumn`: `Column` object or null
  - `md5`: `Column` object, null, or true
  - `sha256`: `Column` object, null, or true
  - `filenameExtFilter`: Array of string or null
- Override `ReferenceColumn` functions for `AssetPseudoColumn`. It only includes the `formatPresentation` function which will take care of returning the value. This is how the value is generated:
  1. If Column has `column-display` annotation use it.
  2. If in edit context, just return the raw value.
  3. Otherwise, return a download link. The caption will be filename column formatted data (if available), or column's formatted data.

- Change `Reference.columns` logic to handle asset columns. The logic is as follows:
  - If we are in edit context, remove `filename`, `bytecount`, `md5`, and `sha256` from visible columns list.
  - When visible-columns annotation is not defined, if column is part of any foreign-keys ignore the asset annotation.

**Chaise test cases are passing with this branch.**

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -2915,9 +2915,7 @@ var ERMrest = (function(module) {
          */
         get default() {
             if (this._default === undefined) {
-                this._default = this._baseCols.reduce(function (res, col, index) {
-                    return res + (index>0 ? ":" : "") + col.default;
-                }, "");
+                this._default = this._simple ? this._baseCols[0].default : null;
             }
             return this._default;
         },
--- a/js/reference.js
+++ b/js/reference.js
@@ -498,21 +498,20 @@ var ERMrest = (function(module) {
                 // if edit context remove filename, bytecount, md5, and sha256 from visible columns
                 if (module._isEntryContext(this._context) && assetColumns.length !== 0) {
                     // given a colName will remove it from visible columns
-                    var removeCol = function(colName) {
+                    var removeCol = function(col) {
                         // property does not exist
-                        if (typeof colName !== 'string') {
+                        if (col === null) {
                             return;
                         }
 
                         // column is not in list of visible columns
-                        if (!(colName in consideredColumns)) {
+                        if (!(col.name in consideredColumns)) {
                             return;
                         }
 
                         // find the column and remove it
                         for (var x = 0; x < self._referenceColumns.length; x++){
-                            col = self._referenceColumns[x];
-                            if (!col.isPseudo && col.name === colName) {
+                            if (!self._referenceColumns[x].isPseudo && self._referenceColumns[x].name === col.name) {
                                 self._referenceColumns.splice(x, 1);
                                 return;
                             }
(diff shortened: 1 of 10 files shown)


js/reference.js
```diff
@@ -473,6 +494,43 @@ var ERMrest = (function(module) {
...
                     }
                 }
+
+                // if edit context remove filename, bytecount, md5, and sha256 from visible columns
+                if (module._isEntryContext(this._context) && assetColumns.length !== 0) {
+                    // given a colName will remove it from visible columns
+                    var removeCol = function(col) {
+                        // property does not exist
+                        if (col === null) {
+                            return;
```
jrchudy: This seems like a case where the API is being used incorrectly and should actually throw an error. Maybe I missed something though.

js/reference.js
```diff
@@ -2769,189 +2828,60 @@ var ERMrest = (function(module) {
...
      * @param {ERMrest.Reference} reference column's reference
-     * @param {?ERMrest.Column} base The column that this reference-column will be created based on.
-     * @param {?Object} kwargs if it's not empty then the column is being created based on foreignkey or key.
+     * @param {ERMrest.Column[]} baseCols List of columns that this reference-column will be created based on.
      * @desc
      * Constructor for ReferenceColumn. This class is a wrapper for {@link ERMrest.Column}.
      */
-    function ReferenceColumn(reference, base, kwargs) {
-
+    function ReferenceColumn(reference, cols) {
```
jrchudy: I thought a `ReferenceColumn` was supposed to be for a single column, not for a set of columns. Unless I need to continue reading to understand the new structure.

Is `cols` an array of columns, but is usually just one? As in, the way you are using the `ReferenceColumn` constructor is by passing in an array of 1 column...? As I look at it more, it looks like `_simple` addresses what I'm asking about. That tells you whether it's composed of more than one column (I think that's how I'm understanding it).

js/utilities.js
```diff
@@ -62,6 +62,26 @@ var ERMrest = (function(module) {
...
+     * @desc
+     * This function should be called to extend a prototype with another one.
+     * Make sure to attach the right constructor to the prototypes after,
+     * and also call `child.superClass.call(this, arguments*)` in frist line of
+     * the child constructor with appropriate arguments.
+     * You can define the extra or overriden functions of child before calling _extends.
+     * This function will take care of copying those functions.
+     * *Must be called after defining parent prototype and child constructor*
+     */
+    module._extends = function (child, parent) {
```
jrchudy: I think this is a good idea to do this so that we don't have to worry about using a polyfill or the implementation changing.

js/reference.js
RFSH: Yes, you're right. At first it was supposed to be for a single column. But in case of ForeignKey and Key, there might be multiple columns. What I was doing before was using just the first column so that it's still just a single column. 

But I changed it to be an array. This way it will be more useful. There are some attributes that don't need to be overridden for case of multiple columns. For instance the `nullok` or `default` attribute. If the `ReferenceColumn` was just wrapper for one Column, then the `ForeignKeyPseudoColumn` and `KeyPseudoColumn` had to implement the same logic just for multiple columns.

js/reference.js
RFSH: You can define an asset column, without defining any `filename`, `bytecolumn`, `md5`, or `sha256` columns for it. It will take care of that case. It think my comments are misleading. I will change them.

PR #405, opened 2017-05-12
Shallow Copy Tuple with de-referenced data

jrchudy:
This PR addresses issue [#1102](https://github.com/informatics-isi-edu/chaise/issues/1102) in chaise. The function I added to the `Tuple` object's prototype allows the user to make a _shallow_ copy of the `Tuple` while de-referencing the private `_data` object. This is so when changes are made to the new `Tuple`, it doesn't modify the originating `Tuple`.

This PR needs to be merged before [this](https://github.com/informatics-isi-edu/chaise/pull/1112) one in `Chaise`.

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -2832,7 +2832,27 @@ var ERMrest = (function(module) {
                 return null;
             }
 
-        }
+        },
+
+        /**
+         * @desc
+         * This function takes the current Tuple (this) and creates a shallow copy of it while de-referencing
+         * the _data attribute. This way _data can be modified in chaise without changing the originating Tuple
+         * @returns {ERMrest.Tuple} a shallow copy of _this_ tuple with it's _data de-referenced
+         */
+         deepCopyData: function() {
+             var newTuple = Object.create(Tuple.prototype);
+             module._clone(newTuple, this);
+             newTuple._data = {};
+
+             //change _data though
+             var keys = Object.keys(this._data);
+             for (var i = 0; i < keys.length; i++) {
+                 newTuple._data[keys[i]] = this._data[keys[i]];
+             }
+
+             return newTuple;
+         }
 
 
     };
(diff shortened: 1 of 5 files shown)


js/reference.js
```diff
@@ -2832,7 +2832,27 @@ var ERMrest = (function(module) {
...
-        }
+        },
+
+        /**
+         * @desc
+         * This function takes the current Tuple (this) and creates a shallow copy of it while de-referencing
+         * the _data attribute. This way _data can be modified in chaise without changing the originating Tuple
+         * @returns {ERMrest.Tuple} a shallow copy of _this_ tuple with it's _data de-referenced
+         */
+         deepCopyData: function() {
```
robes: I find the use of terms too confusing.

Change this function `deepCopyData` to `copy`

So the api will be 
```
var copiedTuple = thisTuple.copy();
```

js/reference.js
```diff
@@ -2832,7 +2832,27 @@ var ERMrest = (function(module) {
...
+             var newTuple = Object.create(Tuple.prototype);
+             module._clone(newTuple, this);
```
robes: I now this isn't part of your diff but `_clone` should be renamed to `_shallowCopy` because its really unclear that 'clone' is supposed to mean shallow copy.

If its a quick and easy change, do it, but if its not easy don't.

test/specs/reference/tests/01.reference.js
```diff
@@ -263,6 +263,18 @@ exports.execute = function (options) {
                 expect(tuple.values).toBeDefined();
             });
 
+            it('tuple.deepCopyData should create a shallow copy of the tuple except for the data.', function() {
+                var newTuple = tuple.deepCopyData();
+
+                // The original and new Tuple should not be the same object, neither should the _data
+                expect(newTuple).not.toBe(tuple);
+                expect(newTuple._data).not.toBe(tuple._data);
```
robes: Right, but your next test should be to `expect(newTuple._data[i].toBe(tuple._data[i])` (in a loop) to test that your function actually did the shallow copy of the keys of `_data`.

PR #407, opened 2017-05-19
This PR adds file upload logic to emrestjs

chiragsanghvi:


js/utilities.js
```diff
@@ -1272,13 +1274,13 @@ var ERMrest = (function(module) {
             });
         }
 
-        var conditionalRegex = /\{\{(#|\^)([\w\d-]+)\}\}/;
+        var conditionalRegex = /\{\{(#|\^)([\w\d-_. ]+)\}\}/;
 
         // If no conditional Mustache statements of the form {{#var}}{{/var}} or {{^var}}{{/var}} not found then do direct null check
-        if (!conditionalRegex.exec(template)) {
+        if (!options.avoidValidation && !conditionalRegex.exec(template)) {
```
robes: does this have anything to do with file upload?

js/http.js
```diff
@@ -240,6 +240,11 @@ var ERMrest = (function (module) {
                                     //throw new Error("httpUnauthorizedFn Event Handler not registered");
                                     deferred.reject(response);
                                 }
+                            } else {
+                                 // Push the current call to _authroizationDefers by calling _onHttpAuthFlowFn
+                                module._onHttpAuthFlowFn().then(function() {
```
robes: this is part of something new that i'm not caught up on

chiragsanghvi: This is the code that we added for handling 401 authentication handling in ermrestjs. Now all unauthorised calls are handled by http layer in ermrestjs and chaise registers a callback which is called in case of http 401 errors. This callback does the login flow and returns a promise. On Successful completion of the flow and resolution of promise, this code will fire all  the calls which were delayed or failed due to the 401 error.

js/utilities.js
chiragsanghvi: Yes. I added it so that we don't validate the template_url for file upload again which would simply return null.

Fileupload has a method to validate the url which is called by Chaise before it begins upload process.

PR #412, opened 2017-06-03
Test added for updating the file asset associated with an entity.

jrchudy:
This PR adds a test to `ermrestJS` that tests editing an entity with an asset. This uploads a new asset for the entity, and then makes sure that the metadata is properly updated as well.


test/specs/reference/tests/10.update.js
```diff
@@ -110,6 +113,138 @@ exports.execute = function (options) {
...
+                		return;
+                	}
+                    done();
+                }, function (err) {
+                    console.dir(err);
+                    done.fail();
+                });
+            });
+
+            it("the metadata for the asset should be updated.", function (done) {
```
RFSH: There are too many `expect` statements here. It's very difficult to understand which one has failed. You have tried to fixed it in two of these expects by adding additional information. We should either do that for other expects here, or break this into different `it` statements.

test/specs/reference/tests/10.update.js
```diff
@@ -110,6 +113,138 @@ exports.execute = function (options) {
...
+                    expect(validRow.bytes).toBe(file.size);
+                    expect(validRow.File_MD5).toBe(file.hash);
+
+                    return uploadObj.createUploadJob();
+                }).then(function(url) {
+                    expect(url).toBeDefined("Chunk url not returned");
+
+                    return uploadObj.fileExists();
+                }).then(function() {
+                    expect(true).toBe(true);
```
RFSH: This is redundant.

jrchudy: So how I started writing this test was by mimicking the tests that @chiragsanghvi wrote specifically for file upload. I wasn't sure what was crucial for the upload task and what wasn't.

This actually looks like a sanity check and can probably be removed under the assumption that the file exists because another spec is testing for that.

RFSH: We can either completely remove the usage of `fileExists` or just remove the `expect` statement. Actually I think it's useful to use the `fileExists` here, since if the promise get rejected it will go to the `catch` and we will understand what the problem is.

test/specs/reference/tests/10.update.js
jrchudy: I added a note of this in the e2e dev guide in the wiki. This is actually something I learned from @chiragsanghvi test cases and something we need to do more often going forward.

test/specs/reference/tests/10.update.js
jrchudy: I decided to remove it because what this is testing for, would be apparent in one of the later expectations.

PR #427, opened 2017-06-16
unify mustache template rendering usage

RFSH:
We had a `_renderTemplate` function that was being used directly in different places. This function would render mustache template with given data. But the data was not following the rules that we had in place for raw and formatted values (by using '_').
Now `_renderTemplate` function will generate the formatted and raw values if the given data is not formatted. Also `_validateTemplate` function has been refactored to support both `url_pattern` and other usages of mustache template.

-----------------
P.S.
- This most probably will break chaise test cases, I will fix those testcases in my chaise PR.

- After merging this, I will update the asset annotation document.

### DIFF
--- a/js/utilities.js
+++ b/js/utilities.js
@@ -419,7 +419,7 @@ var ERMrest = (function(module) {
                 keyValues = module._getFormattedKeyValues(table.columns, context, data);
 
                 // get templated patten after replacing the values using Mustache
-                unformatted = module._renderTemplate(unformattedAnnotation.row_markdown_pattern, keyValues);
+                unformatted = module._renderTemplate(unformattedAnnotation.row_markdown_pattern, keyValues, table, context, {formatted: true});
             }
         }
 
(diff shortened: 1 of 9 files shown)


js/utilities.js
```diff
@@ -1312,31 +1342,81 @@ var ERMrest = (function(module) {
...
+            if (ignoredColumns !== undefined) {
+                ignoredColumns.forEach(function (col) {
+                    ignoredColumns.push("_" + col);
+                });
+            }
+            
+            data = module._getFormattedKeyValues(table.columns, context, data);
+        }
+        
+        // render the template using Mustache
```
jrchudy: Seems like this comment is incorrect. My understanding is that you are validating whether or not it is a valid template and then passing that check off to verify if it's a valid mustache template.

RFSH:
@jrchudy You're correct. Previously we just had one `_renderTemplate` function which wasn't doing any formatting and was using the given data (which I renamed to `_renderMustacheTemplate` in this PR). It was the callers responsibility to format the values. That's why in some cases we forgot to do it.  I just created a wrapper for that function which will take care of formatting values if needed.
(This is also true for `_validateTemplate` and `_validateMustacheTemplate`.)

Issue #419, opened 2017-06-09
File Upload Edit Test


RFSH:
We have unit tests for file upload edit in a table with just one asset column. We should expand the table to have more than 1 columns, to test a more complex scenario.

Issue #418, opened 2017-06-09
File Upload Multi Create Test


RFSH:
We should test file upload in multi create scenario for a table with multiple asset columns. We should make sure that meta data is correctly stored.

PR #434, opened 2017-06-20
File upload tests

jrchudy:
This PR addresses issues #418 and #419. 

The new table in the schema includes 2 different asset columns and the tests verify the data is set properly.


test/utils/utilities.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {string} sortBy - name of column to compare the pageData/tuples by so they are similarly sorted
+ * verifies that the returned data is the same as the submitted data
+ */
+exports.checkPageValues = function(pageData, tuples, sortBy) {
+    pageData.sort(function(a, b) {
+        return a[sortBy] - b[sortBy];
+    });
+
+    tuples.sort(function(a, b) {
+        return a._data[sortBy] - b._data[sortBy];
```
RFSH: This will return `NaN` if any of the `_data` values are not numeric. You can change it to [`localeCompare`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare).

test/specs/upload/tests/03.update_w_upload.js
```diff
@@ -0,0 +1,180 @@
...
+                    hash: "08b46181d7094b5ece88bb389c7499af",
+                    hash_64: "CLRhgdcJS17OiLs4nHSZrw=="
+                }
+            ];
+
+            beforeAll(function (done) {
+
+                for (var i=0; i<files.length; i++) {
+                    var file = files[i];
+                    var filePath = "test/specs/reference/files/" + file.name;
```
RFSH: Change the path to `test/specs/upload/files`. Then you can remove `test/specs/reference/files` folder since we don't have any other file upload test cases in reference.

test/specs/upload/tests/03.update_w_upload.js
```diff
@@ -0,0 +1,180 @@
...
+                    done();
+                }).catch(function (error) {
+                    console.dir(error);
+                    done.fail();
+                });
+            });
+
+            afterAll(function(done) {
+                for (var j=0; j<files.length; j++) {
+                    var filePath = "test/specs/reference/files/" + files[j].name;
```
RFSH: The same comment about the path.

test/specs/upload/conf/upload/schema.json
```diff
@@ -1,152 +1,299 @@
...
+                    "comment": "asset/reference",
+                    "name": "uri",
+                    "default": null,
+                    "nullok": true,
+                    "type": {
+                        "typename": "text"
+                    },
+                    "annotations": {
+                        "tag:isrd.isi.edu,2017:asset": {
+                            "url_pattern" : "/hatrac/ermrestjstest/{{{_fk_id}}}/{{{_uri.md5_hex}}}",
```
RFSH: Should we change this hatrac url_pattern in this PR too? or should it be part of hatrac test cases?

test/specs/upload/conf/upload/schema.json
```diff
@@ -1,152 +1,299 @@
...
+                        "typename": "text"
+                    },
+                    "annotations": {
+                        "comment": [
+                            "thumbnail",
+                            "download"
+                        ],
+                        "tag:isrd.isi.edu,2017:asset": {
+                            "byte_count_column": "file1_bytes",
+                            "url_pattern": "/hatrac/js/ermrestjs/{{{key}}}/{{{file1_MD5}}}",
```
RFSH: change this url_pattern to `/hatrac/js/ermrestjs/{{{_key}}}/{{{_file1_MD5}}}`. This means using the unformatted values for these two. It's true that the unformatted and formatted for the current values are the same, but if someone looks at this schema as an example, it might be misleading for them.

test/specs/upload/conf/upload/schema.json
```diff
@@ -1,152 +1,299 @@
...
+                        "typename": "text"
+                    },
+                    "annotations": {
+                        "comment": [
+                            "thumbnail",
+                            "download"
+                        ],
+                        "tag:isrd.isi.edu,2017:asset": {
+                            "byte_count_column": "file2_bytes",
+                            "url_pattern": "/hatrac/js/ermrestjs/{{{key}}}/{{{file2_MD5}}}",
```
RFSH: Same comment about url_pattern

test/specs/upload/conf/upload/schema.json
```diff
@@ -1,152 +1,299 @@
...
-          "unique_columns": [
-            "id"
-          ]
+    "table_names": [
+        "file",
+        "referee"
+    ],
+    "comment": null,
+    "annotations": {
+        "tag:misd.isi.edu,2015:display": {
```
RFSH: This is something related to the previous schema and not related to your PR. But since you're changing this PR, please remove this unnecessary annotation.

test/utils/utilities.js
jrchudy: This is only ever using the `sortBy` value (which I use an int column always to do it). But I will change it to `localeCompare` for robustness.

test/specs/upload/tests/03.update_w_upload.js
jrchudy: I forgot that should have been adjusted after you changed it in your PR.

test/specs/upload/conf/upload/schema.json
jrchudy: I will adjust the url pattern for the hatrac tests in the PR I make for hatrac tests.

PR #435, opened 2017-06-22
This PR is for issue#1041

nitishksahu:
The PR has the following changes done:

- A new case in core.js for JSON and JSONB

- A fix in if condition to look for null value

- A condition to check for string values and return values enclosed in double quotes to be displayed in the UI.

- A formatJSONpresentation() function to return string appended with pre tag to show the pre-formatted string in the UI.

- And if condition to call the formatJSONpresentation() function based on the type of the columns.

- PrintJSON() function to stringify the JSON and pass a prettiness factor

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -811,8 +811,7 @@ var ERMrest = (function(module) {
                      * If 1 row has it set and none of the others, it cannot be part of defaults
                     **/
                     for (var m = 0; m < data.length; m++) {
-                        console.log(data[m][columnName]);
-                        if (data[m][columnName]===undefined) {
+                        if (data[m][columnName]!== undefined && data[m][columnName] !==null) {
                             notSet = false;
                             break;
                         }
--- a/js/reference.js
+++ b/js/reference.js
@@ -811,13 +811,14 @@ var ERMrest = (function(module) {
                      * If 1 row has it set and none of the others, it cannot be part of defaults
                     **/
                     for (var m = 0; m < data.length; m++) {
-                        if (data[m][columnName]) {
+                        console.log(data[m][columnName]);
+                        if (data[m][columnName]===undefined) {
                             notSet = false;
                             break;
                         }
                     }
 
-                    if (notSet) defaults.push(columnName);
+                    if (!notSet) defaults.push(columnName);
                 });
 
                 return defaults;
--- a/js/reference.js
+++ b/js/reference.js
@@ -2768,7 +2768,7 @@ var ERMrest = (function(module) {
                         else {
                             values[i] = column.formatPresentation(keyValues[column.name], { formattedValues: keyValues , context: this._pageRef._context });
 
-                            if (column.type.name === "json"|| "jsonb") {
+                            if (column.type.name === ("json"|| "jsonb")) {
                                 values[i].isHTML = true;
                             }
                             if (column.type.name === "gene_sequence") {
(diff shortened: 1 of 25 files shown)


test/specs/reference/tests/16.default_value.js
```diff
@@ -0,0 +1,94 @@
+
```
RFSH: These are the changes for your other PR that was merged. Why is this part of this PR too?

js/utilities.js
```diff
@@ -384,7 +384,14 @@ var ERMrest = (function(module) {
 
             try {
                 var col = columns.get(k);
-                keyValues[k] = col.formatvalue(data[k], { context: context });
+                //If the column type is json or jsonb, then insert the raw data instead of formatted data.
+                //Inserting a formatted data would lead to show blank value instead of "null" value in UI
+                if(col.type.name===('json'||'jsonb') && data[k]===null){
```
RFSH: You don't need this check here. It should go to `formatvalue`, so it can call the [`_getNullvalue`](https://github.com/informatics-isi-edu/ermrestjs/blob/8b1f4d3e5f85d7e72a3127aee0ef523c9cd760a7/js/utilities.js#L318) function. This function is using [display annotation](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2015-display) to find the null value. 

JSON column value will be null in two cases:
1. Actual "null" has been inserted into the JSON column in database.
2. The value doesn't exist.

And since we cannot differentiate between these two cases, it's ok to not show exactly "null" and apply the same logic that we have for showing null value.

js/reference.js
```diff
@@ -3111,6 +3129,18 @@ var ERMrest = (function(module) {
         },
 
         /**
+         * Formats the presentation value corresponding to this reference-column definition.
+         * @param {String} data In case of pseudocolumn it's the raw data, otherwise'formatted' data value.
+         * @param {Object} options includes `context` and `formattedValues`
+         * @returns {Object} A key value pair containing value and isHTML that detemrines the presenation.
+         */
+        formatJsonPresentation: function(data, options) {
```
RFSH: What's the purpose of this function? This should be part of your `formatvalue` function. If your `formatvalue` function is different in different contexts, you can use the context that we are passing as an option to figure out which way to format the JSON column.

js/reference.js
```diff
@@ -2723,12 +2721,24 @@ var ERMrest = (function(module) {
...
+                            else {
                                 presentation = column.formatPresentation(this._data, { formattedValues: keyValues, context: this._pageRef._context});
                             }
                             this._values[i] = presentation.value;
                             this._isHTML[i] = presentation.isHTML;
-                        } else {
+                        }
+                        else if(column.type.name === ("json" || "jsonb")){
+                            presentation = column.formatPresentation(keyValues[column.name], { formattedValues: keyValues , context: this._pageRef._context });
+                            if(typeof keyValues['_'+column.name] === "string"){
```
RFSH: I don't understand the purpose of this if statement.

js/reference.js
```diff
@@ -2745,12 +2755,20 @@ var ERMrest = (function(module) {
...
                             if (column.isForeignKey) {
                                 values[i] = column.formatPresentation(this._linkedData[column._constraintName], {context: this._pageRef._context});
-                            } else {
+                            }
+                            else {
                                 values[i] = column.formatPresentation(this._data, { formattedValues: keyValues, context: this._pageRef._context});
                             }
-                        } else {
+                        }
+                        else if (column.type.name === ("json" || "jsonb")){
```
RFSH: As I mentioned in another comment, you don't need this check here. You don't need to call another function.

js/reference.js
nitishksahu: The purpose of the if statement, where it checks for JSON and jsonb, is to pass the formatted value (string representation) of the json.

nitishksahu:
New PR#436 has been raised for these changes, closing this PR as it was showing some old commits which have already been pushed to master.

PR #436, opened 2017-06-23
Changes made for JSON and JSONB column support

nitishksahu:
The PR has the following changes done:

formatValue() function has been changed for JSON and jsonb column types.

A case is added for JSON and jsonb column types.

If condition in reference.js for JSON and jsonB column types.

PrintJSON() function to stringify the JSON and pass a prettiness factor as well as return the pre-formatted string for display.

Test cases added for the JSON support functionality added.

### DIFF
--- a/js/core.js
+++ b/js/core.js
@@ -1711,7 +1711,7 @@ var ERMrest = (function (module) {
         this.formatvalue = function (data, options) {
             //This check has been added to show "null" in all the rows if the user inputs blank string
             //We are opting json out here because we want null in the UI instead of "", so we do not call _getNullValue for json
-            if (data === undefined || (data === null && this.type.name.indexOf('json') !== 0)) {
+            if (data === undefined || (data === null && this.type.name.indexOf('json') !== 0) || (data === null && this.type.name.indexOf('jsonb') !== 0)) {
                     return this._getNullValue(options ? options.context : undefined);
                 }
             if (data === undefined) {
--- a/js/core.js
+++ b/js/core.js
@@ -1709,7 +1709,13 @@ var ERMrest = (function (module) {
          * @returns {string} The formatted value.
          */
         this.formatvalue = function (data, options) {
-            if (data === undefined || data === null) {
+            //This check has been added to show "null" in all the rows if the user inputs blank string
+            if(this.type.name!=='json'){
+                if (data === undefined || data === null) {
+                    return this._getNullValue(options ? options.context : undefined);
+                }
+            }
+            if (data === undefined) {
                 return this._getNullValue(options ? options.context : undefined);
             }
             /* TODO format the raw value based on the column definition
@@ -1742,6 +1748,11 @@ var ERMrest = (function (module) {
                     break;
                 case 'gene_sequence':
                     data = utils.printGeneSeq(data, options);
+                    break;
+                //Cases to support json and jsonb columns
+                case 'json':
+                case 'jsonb':
+                    data = utils.printJSON(data,options);
                     break;
                 default: // includes 'text' and 'longtext' cases
                     data = utils.printText(data, options);
--- a/js/core.js
+++ b/js/core.js
@@ -1710,11 +1710,10 @@ var ERMrest = (function (module) {
          */
         this.formatvalue = function (data, options) {
             //This check has been added to show "null" in all the rows if the user inputs blank string
-            if(this.type.name!=='json'){
-                if (data === undefined || data === null) {
+            //We are opting json out here because we want null in the UI instead of "", so we do not call _getNullValue for json
+            if (data === undefined || (data === null && this.type.name.indexOf('json') !== 0)) {
                     return this._getNullValue(options ? options.context : undefined);
                 }
-            }
             if (data === undefined) {
                 return this._getNullValue(options ? options.context : undefined);
             }
--- a/js/reference.js
+++ b/js/reference.js
@@ -2730,8 +2730,8 @@ var ERMrest = (function(module) {
                             this._isHTML[i] = presentation.isHTML;
                         }
                         //Added this if conditon explicitly for json/jsonb because we need to pass the 
-                        //formatted string representation of JSON values
-                        else if(column.type.name === ("json" || "jsonb")){
+                        //formatted string representation of JSON and JSONBvalues
+                        else if (column.type.name === "json" || column.type.name === "jsonb") {
                             presentation = column.formatPresentation(keyValues[column.name], { formattedValues: keyValues , context: this._pageRef._context });
                             this._values[i] = presentation.value;
                             this._isHTML[i] = presentation.isHTML;
@@ -2758,7 +2758,7 @@ var ERMrest = (function(module) {
                             }
                         } else {
                             values[i] = column.formatPresentation(keyValues[column.name], { formattedValues: keyValues , context: this._pageRef._context });
-
+                            // If the column type is json or jsonB we will send the templated string with <pre> tag
                             if (column.type.name === "gene_sequence"|| column.type.name === "json" || column.type.name === "jsonb") {
                                 values[i].isHTML = true;
                             }
--- a/js/reference.js
+++ b/js/reference.js
@@ -2728,7 +2728,15 @@ var ERMrest = (function(module) {
                             }
                             this._values[i] = presentation.value;
                             this._isHTML[i] = presentation.isHTML;
-                        } else {
+                        }
+                        //Added this if conditon explicitly for json/jsonb because we need to pass the 
+                        //formatted string representation of JSON values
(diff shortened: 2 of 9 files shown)


js/core.js
```diff
@@ -1709,7 +1709,13 @@ var ERMrest = (function (module) {
          * @returns {string} The formatted value.
          */
         this.formatvalue = function (data, options) {
-            if (data === undefined || data === null) {
+            //This check has been added to show "null" in all the rows if the user inputs blank string
+            if(this.type.name!=='json'){
```
RFSH: You can rewrite these two `if` statements like this:
``` javascript
if (data === undefined || (data === null && this.type.name.indexOf('json') !== 0) {
  //...
}
```
It would be easier to understand and also supports `jsonb`. 

Also add a comment that why we're opting `json` out in this case.

test/specs/reference/tests/12.reference_values_edit.js
```diff
@@ -234,6 +234,70 @@ exports.execute = function (options) {
...
+
+            var multipleEntityUri=options.url + "/catalog/" + catalog_id + "/entity/" + schemaName + ":"+ tableName ;
+
+            var reference, page, tuples;
+
+            beforeAll(function(done) {
+                
+                // Fetch the entities beforehand
+                options.ermRest.resolve(multipleEntityUri).then(function (response) {
+                    reference = response;
```
RFSH: the function call that changes the `context` of a reference is `.contextualize.CONTEXTNAME`. so the reference now is not contextualize. You need to change this to `reference = response.contextualize.entryEdit`

js/utilities.js
```diff
@@ -731,6 +731,44 @@ var ERMrest = (function(module) {
...
+        printJSON: function printJSON(value, options) {
+           options = (typeof options === 'undefined') ? {} : options;
+           if(options.context==="entry/edit"){
+               if (value === null) {
+                   return 'null';
+                }
+                else if (typeof value === 'object') {
+                       return JSON.stringify(value,undefined,2);
+                }
+                else if (typeof value === 'string') {
```
RFSH: You don't need any of these `if` statements. `JSON.stringify` returns the correct value in all of the cases. just `return JSON.stringify(value, undefined, 2);`.

Also change this to check for any entry contexts. Although we are never going to call this function in `entry/create`, but it doesn't make sense to add `<pre>` tags to any entry contexts, as these contexts are suppose to show the "raw" value.

```javascript
if (module._isEntry(options.context)) {
  return JSON.stringify(value, undefined, 2);
}
```

js/utilities.js
```diff
@@ -731,6 +731,44 @@ var ERMrest = (function(module) {
...
         },
+        
+        /**
+        * @function		    
+        * @param {Object} value A json value to transform
+        * @param {Object} [options] Configuration options.
+        * @return {string} A string representation of value
+        * @desc Formats a given json value into a string for display.
+        */
+        printJSON: function printJSON(value, options) {
```
RFSH: In the description of the function, comment that what it actually is returning based on different contexts.

test/specs/reference/tests/12.reference_values_edit.js
```diff
@@ -234,6 +234,70 @@ exports.execute = function (options) {
...
+            
+            });
+            
+            it("Testing for EDIT JSON and JSONB tuples values", function() {
+                
+                for( var i=0; i<limit; i++){
+                    let id=jsontuples[i]._data.id;
+                    let josn_col=jsontuples[i]._data.json_col;
+                    let josnb_col=jsontuples[i]._data.jsonb_col;
+                    expect(id).toBe(expectedValues[i].id);
```
RFSH: You were supposed to test the `tuple[i].values` in here not `tuple[i]._data`.

test/specs/reference/tests/05.reference_values.js
```diff
@@ -242,6 +242,70 @@ exports.execute = function (options) {
...
+                console.dir(err);
+                done.fail();
+            });
+        
+        });
+        
+        it("Testing for JSON and JSONB tuples values", function() {
+            
+            for( var i=0; i<limit; i++){
+                let id=jsontuples[i]._data.id;
```
RFSH: The same comment about testing the `values` and not `_data`

test/specs/reference/tests/05.reference_values.js
```diff
@@ -242,6 +242,70 @@ exports.execute = function (options) {
 
             testTupleValidity(6, values, isHTML);
         });
+        
 
     });
+    
+    describe("Testing for JSON AND JSONB Values", function() {
+        var expectedValues=[{"id":"1001","json_col":true,"jsonb_col":true},
```
RFSH: These should be the formatted values. You are testing the `values` here and not `_data`.

test/specs/reference/tests/12.reference_values_edit.js
```diff
@@ -234,6 +234,70 @@ exports.execute = function (options) {
 
             testTupleValidity(6, values);
         });
+        
+        
+        describe("Testing for EDIT JSON and JSONB Values", function() {
+            var expectedValues=[{"id":"1001","json_col":true,"jsonb_col":true},
```
RFSH: These should be formatted values.

js/reference.js
```diff
@@ -2728,7 +2728,15 @@ var ERMrest = (function(module) {
                             }
                             this._values[i] = presentation.value;
                             this._isHTML[i] = presentation.isHTML;
-                        } else {
+                        }
+                        //Added this if conditon explicitly for json/jsonb because we need to pass the 
+                        //formatted string representation of JSON values
+                        else if(column.type.name === ("json" || "jsonb")){
```
RFSH: I don't think that's the right syntax.
```javascript
else if (column.type.name === "json" || column.type.name === "jsonb") {
```

js/reference.js
```diff
@@ -2751,7 +2759,7 @@ var ERMrest = (function(module) {
                         } else {
                             values[i] = column.formatPresentation(keyValues[column.name], { formattedValues: keyValues , context: this._pageRef._context });
 
-                            if (column.type.name === "gene_sequence") {
+                            if (column.type.name === "gene_sequence"|| column.type.name === "json" || column.type.name === "jsonb") {
```
RFSH: add a comment why you're doing this.

test/specs/print_utils/tests/01.print_utils.js
```diff
@@ -61,6 +61,26 @@ exports.execute = function (options) {
...
+            var printJSON = formatUtils.printJSON;
+            var options={context:'entry/edit'};
+            expect(printJSON(null,options)).toBe('null');
+            expect(printJSON('',options)).toBe('""');
+            expect(printJSON(true,options)).toBe('true');
+            expect(printJSON(false,options)).toBe('false');
+            expect(printJSON(2.9,options)).toBe('2.9');
+        });
+        
+        it('printJSON() should format JSON values correctly for Not Edit Context.', function() {
```
RFSH: Also add a JSON object, to see that it's actually returning the beautified version.

test/specs/print_utils/tests/01.print_utils.js
```diff
@@ -61,6 +61,26 @@ exports.execute = function (options) {
             expect(printText({key:123})).toBe('{"key":123}');
             expect(printText({key:123,subkey:{subsubkey:456}})).toBe('{"key":123,"subkey":{"subsubkey":456}}');
         });
+        
+        it('printJSON() should format JSON values correctly for Edit Context.', function() {
```
RFSH: Also add a JSON object, to see that it's actually returning the beautified version.

test/specs/reference/tests/12.reference_values_edit.js
```diff
@@ -234,6 +234,105 @@ exports.execute = function (options) {
...
+                    console.dir(err);
+                    done.fail();
+                }).catch(function(err) {
+                    console.dir(err);
+                    done.fail();
+                });
+                options.ermRest.appLinkFn(appLinkFn);
+            });
+            
+            it("Testing for EDIT JSON and JSONB tuples values", function() {
```
RFSH: Change this to something like "should return the expected values." to be a more meaningful log.

test/specs/reference/tests/05.reference_values.js
```diff
@@ -242,6 +242,106 @@ exports.execute = function (options) {
...
+                console.dir(err);
+                done.fail();
+            }).catch(function(err) {
+                console.dir(err);
+                done.fail();
+            });
+            options.ermRest.appLinkFn(appLinkFn);
+        });
+        
+        it("Testing for JSON and JSONB tuples values", function() {
```
RFSH: Change this message the same to something more meaningful.

nitishksahu:
Changes for JSON and JSONB column support merged to master branch.

PR #438, opened 2017-06-27
Add Inbound Foreignkeys to Reference.columns

RFSH:
If data modelers have mentioned an inbound foreignkey in their visible columns list, ermrestjs will create a pseudo column for that in any non-entry contexts. `column.isInboundForeignKey` indicates whether the given column is inbound fk or not.

Since for related entities we need the tuple, for creating these type of columns we need the tuple too. I didn't wanted to change the whole api of columns, so I added another function that can be called to generate columns list. `Reference.generateColumnsList(tuple)` and `Reference.columns` will return the same list. The only difference being is that for inbound forengkeys, if we don't specify the tuple the [filter criteria ](https://github.com/informatics-isi-edu/ermrestjs/issues/367) won't be added to url.

The value of these columns can be retrieved by calling read on the reference that they return (`column.reference.read()`).

Running test cases locally will fail on upload test cases since we deleted the namespace, but travis build should be successful.


RFSH:
@jrchudy Yes exactly what I was thinking about. The problem with putting classes in different files is that since we're not exposing any of the classes then they don't have access to each other. We should investigate a better way of managing that.

Issue #440, opened 2017-06-28
Column check default values before setting them


jrchudy:
Currently `chaise` wrongly assumes that a default value can be of type function, but in reality, that function might just be a stringified version of that function call (`now()` for instance).

`ermrestJS` should properly check the columns type and then try to coerce the schema defined default value to that type. If that fails, set the default value to null and let PSQL manage setting that value.

This issue is related to [chaise issue #1174](https://github.com/informatics-isi-edu/chaise/issues/1174).

jrchudy:
The corresponding PR was merged into master. Closing this issue.

PR #441, opened 2017-06-28
Defaults are parsed to verify value

jrchudy:
This PR addresses issue #440 and [issue #1174 in chaise](https://github.com/informatics-isi-edu/chaise/issues/1174). 

There was an issue where functions were being parsed and having errors shown in the UI (specifically for the case of timestamps).

### DIFF
--- a/js/core.js
+++ b/js/core.js
@@ -1943,7 +1943,9 @@ var ERMrest = (function (module) {
                         case "timestamptz":
                             // convert using moment, if it doesn't error out, set the value.
                             // try/catch catches this if it does error out and sets it to null
-                            moment(defaultVal);
+                            if (!module._moment(defaultVal).isValid()) {
+                                throw new Error();
+                            }
                             break;
                         case "json":
                         case "jsonb":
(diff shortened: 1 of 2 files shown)


js/core.js
```diff
@@ -1913,6 +1908,55 @@ var ERMrest = (function (module) {
                     module._fixedEncodeURIComponent(this.name)].join(":");
         },
 
+        /**
+         * return the default value for a column after checking whether it's a primitive that can be displayed properly
+         * @return {string}
+         */
+        get default () {
+            if (!this._default) {
```
RFSH: Change this to check for `undefined`.

js/core.js
```diff
@@ -1913,6 +1908,55 @@ var ERMrest = (function (module) {
                     module._fixedEncodeURIComponent(this.name)].join(":");
         },
 
+        /**
+         * return the default value for a column after checking whether it's a primitive that can be displayed properly
+         * @return {string}
```
RFSH: Is the default always an string? Can it be a boolean, integer, or JSON object?

jrchudy: These values are being used to display in the UI. In the end, when this data is submitted to `ermrestJS`, all of this data is in strings. The input fields will properly try to show those values in the UI.

Issue #443, opened 2017-06-29
change update to only send information that was changed by user.


jrchudy:
The update function in `ermrestJS` should only send the values that the user changed to the database for update rather than the full set. This sounds a lot like an issue we had before and functionality we had before.

The reasoning behind this is so we can more easily track what fields user change by looking at database update logs.

This goes along with [chaise issue #1173](https://github.com/informatics-isi-edu/chaise/issues/1173).

karlcz:
To be totally explicit, what's important is to adjust the update
targets projection in the URL so you're only asking ERMrest to apply
updates on the desired columns.

Since you're using JSON input, it's OK to send extra columns of input,
so you could always just send all form data rather than build
different subsets for each request.  ERMrest will just ignore the
extra fields not described in the URL.

PR #444, opened 2017-07-08
Update projections

jrchudy:
This PR fixes issue #443 and chaise issue [#1173](https://github.com/informatics-isi-edu/chaise/issues/1173).

The changes were made to only include the changed columns in the column projections and submission data now. In doing so this uncovered that we were copying the data from the tuples to the page data object improperly because of inconsistent ordering.

Another issue with not submitting all the data is that we can no longer rely on key values to be aliased with the new alias. If a key wasn't changed, we only send the old key value (before we sent both regardless).

I made sure the chaise e2e tests still pass based on the changes I made.

One thing to note:
lines 1175-1184 are meant to address the case when the shortest key is a composite key *AND* those columns were changed in _multiple entities_. It seems to work appropriately, but doesn't actually verify a case like this:
```
[row 1: {
  comp_key1: 1,
  comp_key2: 2
}, row 2: {
  comp_key1: 1,
  comp_key2: 3
}, row 3: {
  comp_key1: 2,
  comp_key2: 3
}]
```
In this case it would set the `duplicate` for both keys and never set the `rowIndexInSubData` (row index in submission data) value properly.

@RFSH: Do you think I should add another test case to `10.update.js` where a composite key is changed to be like the above and then verify the row contents? (I think I already know this answer)

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -3643,8 +3643,7 @@ URI should be to the ERMrest _service_. For example,
 <a name="ERMrest.parse"></a>
 
 ### ERMrest.parse(uri) ⇒ <code>ERMrest.Location</code>
-This is an internal function that parses a URI and constructs an
-internal representation of the URI.
+This function parses a URI and constructs a representation of the URI.
 
 **Kind**: static method of [<code>ERMrest</code>](#ERMrest)  
 **Returns**: <code>ERMrest.Location</code> - Location object created from the URI.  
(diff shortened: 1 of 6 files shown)


jrchudy:
@RFSH: I added a test case that verifies the composite keys work properly. Good thing I did, because I found a minor bug in my logic that I implemented today. Lesson learned that `break` is bad.

PR #446, opened 2017-07-10
added a function to reference to produce a csv download url

jrchudy:
This PR addresses issue #442.


jrchudy:
Yeah we do. I realized that the test would be as simple as verifying the URI. Added an expectation for this now.

Issue #304, opened 2016-12-16
API for retrieving aggregate functions


hongsudt:
We need a way to retrieve aggregate values associated with a reference such as count, min, max, etc. Should this be something like Referece.getAggregate(aggregateList). 

This API is needed for informatics-isi-edu/chaise#878 and will be useful for faceting to retrieve a range of a number/date column. 

check the flow listed on https://github.com/informatics-isi-edu/chaise/wiki/Faceting

@karlcz can a list of aggregate be specified to ermrest or do we have to do this one at a time?

karlcz:
ermrest has both aggregate and attributegroup APIs which support a list of aggregate function "projections".  The aggregate API projects one row of results from an entire table or query set (i.e. grand total count, min, or max) while the attributegroup API projects one row of results for each group key in a table or query set (i.e. group-local count, min, or max for rows sharing the same group key).

karlcz:
In practice, aggregates are implemented as a table scan and it will be cheaper to gather all desired statistics from the same table or query set in a single request rather than via separate requests, each of which has to scan the whole table again.

This whole table scan is also a limiting factor on huge tables since you will eventually run into HTTP request timeouts with millions of rows to process.

RFSH:
Sketch of the proposed api is as follows:
```javascript
Reference.prototype = {
  /**
  * @param{AggregateFn[]} aggregateList list of aggregate functions
  **/
  getAggergate = function(aggregateList) {
    // TODO use the list to make the request and return data
  }

  get rowCount() {
    // TODO return the appropriate aggregateFn corresponding to row count of current reference.
  }
 
  //TODO other aggregate functions
}

ReferenceColumn.prototype = {
  get min () {
    // TODO return the appropriate aggregateFn corresponding to min value of this column
  }

  get min () {
    // TODO return the appropriate aggregateFn corresponding to min value of this column
  }

  //TODO other aggregate functions
}

AggregateFn.prototype = {
 // TODO should have a mapping of function to the actual uri that ermrest understands
}

// usage:
reference.getAggregate(reference.rowCount, reference.columns[0].min);
```

hongsudt:
@RFSH Should the getAggregate take any number of arguments?

jrchudy:
[For aggregate API documentation](https://github.com/informatics-isi-edu/ermrest/blob/master/api-doc/data/rest.md#aggregate-retrieval)
[...]

PR #453, opened 2017-07-13
Simple column aggregates defined

jrchudy:
This PR adds basic functionality for column aggregates (count not null, count distinct, min, and max) and table aggregates (count not null). This is for issue #304.

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -1578,7 +1578,7 @@ var ERMrest = (function(module) {
             var aggAliases = [];
             for (var i = 0; i < aggregateList.length; i++) {
                 var agg = aggregateList[i];
-                var alias = (agg.alias != null ? agg.alias : i);
+                var alias = (agg.alias !== null ? agg.alias : i);
 
                 // this alias is already used, should only be the case when the user defined an alias to be used twice
                 if (aggAliases.indexOf(alias) > -1) {
@@ -1587,7 +1587,7 @@ var ERMrest = (function(module) {
                 aggAliases.push(alias);
 
                 // if this is the first aggregate, begin with the baseUri
-                if (i == 0) {
+                if (i === 0) {
                     url = baseUri;
                 } else {
                     url += ",";
@@ -1596,7 +1596,7 @@ var ERMrest = (function(module) {
                 // if adding the next aggregate to the url will push it past url length limit, push url onto the urlSet and reset the working url
                 if ((url + alias + ":=" + agg.value).length > URL_LENGTH_LIMIT) {
                     // strip off an extra ','
-                    if (url.charAt(url.length-1) == ',') {
+                    if (url.charAt(url.length-1) === ',') {
                         url = url.substring(0, url.length-1);
                     }
 
@@ -1607,7 +1607,7 @@ var ERMrest = (function(module) {
                 url += alias + ":=" + agg.value;
 
                 // We are at the end of the aggregate list
-                if (i+1 == aggregateList.length) {
+                if (i+1 === aggregateList.length) {
                     urlSet.push(url);
                 }
             }
--- a/js/reference.js
+++ b/js/reference.js
@@ -1473,7 +1473,12 @@ var ERMrest = (function(module) {
             return newReference;
         },
 
-        getAggregates: function() {
+        /**
+         *
+         * @param {ERMrest.ColumnAggregateFn[]} aggregateList - list of aggregate functions to apply to GET uri
+         * @return {Promise} TODO: not sure of the structure of this returned promise yet
+         */
+        getAggregates: function(aggregateList) {
 
         },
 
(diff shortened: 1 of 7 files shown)


js/reference.js
```diff
@@ -1473,6 +1478,49 @@ var ERMrest = (function(module) {
...
+
+            // use this index to get the next ascii value in the alphabet starting at `a`
+            var aliasCounter = 0;
+            function generateAlias() {
+                var alias = String.fromCharCode(97 + aliasCounter);
+                aliasCounter++;
+                return alias;
+            }
+
+            var url = this.location.service + "/catalog/" + this.location.catalog + "/aggregate/" + this.location.compactPath + "/";
```
RFSH: I think you're still working on this issue. But this is something that feel like I should have emphasised on. `Location.compactPath`, `Location.path`, `Location.uri`, and `Location.compactUri` return uris that chaise understand. To make a request to ermrest use the "ermrset version" of these apis (`Location.ermrestCompactPath`, `Location.ermrstPath`, `Location.ermrestUri`, and `Location.ermrestCompactUri`).

The only difference right now is that ermrest version of these uris will change the `search` pseudo-operator to appropriate operator that ermrest understands.

jrchudy:
I will be updating this issue with a test case for url length limit being exceeded and forcing the request to be split into multiple requests.

jrchudy:
@RFSH This PR is ready to be reviewed now. I made a few more minor changes.
 - Aliasing is now based on the index of the aggregate in the list (test added for this with some aliases defined)
 - url length limit is handled (and a test case tests this)

js/reference.js
```diff
@@ -1555,6 +1560,83 @@ var ERMrest = (function(module) {
...
+
+            var URL_LENGTH_LIMIT = 2048;
+
+            var urlSet = [];
+            var baseUri = this.location.service + "/catalog/" + this.location.catalog + "/aggregate/" + this.location.ermrestCompactPath + "/";
+            // for matching the aliases after the data returns
+            var aggAliases = [];
+            for (var i = 0; i < aggregateList.length; i++) {
+                var agg = aggregateList[i];
+                var alias = (agg.alias != null ? agg.alias : i);
```
RFSH: lint complains about this `!=` check.

js/reference.js
```diff
@@ -1555,6 +1560,83 @@ var ERMrest = (function(module) {
...
+
+                // this alias is already used, should only be the case when the user defined an alias to be used twice
+                if (aggAliases.indexOf(alias) > -1) {
+                    alias = i;
+                }
+                aggAliases.push(alias);
+
+                // if this is the first aggregate, begin with the baseUri
+                if (i == 0) {
```
RFSH: lint complains about `i == 0` check (should be `===`).

js/reference.js
```diff
@@ -1555,6 +1560,83 @@ var ERMrest = (function(module) {
...
+            var urlSet = [];
+            var baseUri = this.location.service + "/catalog/" + this.location.catalog + "/aggregate/" + this.location.ermrestCompactPath + "/";
+            // for matching the aliases after the data returns
+            var aggAliases = [];
+            for (var i = 0; i < aggregateList.length; i++) {
+                var agg = aggregateList[i];
+                var alias = (agg.alias != null ? agg.alias : i);
+
+                // this alias is already used, should only be the case when the user defined an alias to be used twice
+                if (aggAliases.indexOf(alias) > -1) {
```
RFSH: Instead of just an `if`, it should be recursive. If `aggregateList = [ {'alias': 1, ..}, {<no-alias> ...}]` it's still going to use the same alias for both of them.

js/reference.js
```diff
@@ -4132,8 +4219,95 @@ var ERMrest = (function(module) {
...
+     *  for a specific column
+     * @memberof ERMrest
+     * @class
+     * @param {ERMrest.ReferenceColumn} column - the column that is used for creating column aggregates
+     */
+    function ColumnAggregateFn (column) {
+        this.column = column;
+    }
+
+    ColumnAggregateFn.prototype = {
```
RFSH: These `ColumnAggregateFn`s should either complain when the `column.isPseudo` is correct, or do something extra for those. Because `this.column.name` won't return a name that is understandable by ermrest.

test/specs/reference/tests/17.aggregates.js
```diff
@@ -0,0 +1,322 @@
+exports.execute = function (options) {
+    // Test Cases:
+    describe("for testing default values while creating an entity/entities,", function () {
```
RFSH: You forgot to change the describe message.

js/reference.js
jrchudy: This is searching through an array of strings with only the aggregate aliases. I grab the alias from the aggregate (if provided) then verify if I have added it to my list of aggregate aliases already.

Issue #455, opened 2017-07-14
Don't show regex in the search box on recordset


RFSH:
@hongsudt commented on [Tue May 16 2017](https://github.com/informatics-isi-edu/chaise/issues/1117)

when you do a search in recordset, and then if you drill down to a page and then go back to the recordset, the search you entered get “regex”-ified looking rather technical.

Suggested solution (from the chat room): so you would need to change the fragment syntax to represent some higher-level search primitive that stores the original search term, and only expand to the full regex when sending to ermrest. A simple solution is to use a new operator like :search:  for the fragment..

---

@robes commented on [Thu May 18 2017](https://github.com/informatics-isi-edu/chaise/issues/1117#issuecomment-302534247)

+1 for the simple solution.

RFSH:
This is done as part of #456 PR.

PR #456, opened 2017-07-14
Change Search Operator

RFSH:
Previously we were using actual `::ciregex::` operator for search, the problem with that usage was that we were manipulating the searchTerm internally so ermrest understands it. For instance when we search for `11`, what we need to send to ermrest is `^(.*[^0-9.])?0*11([^0-9].*|$)` regular expression. So we needed to keep two versions of filter, one that users see and one that ermrest understand.

The uri that parser returns (`Location.uri`) is what the client (chaise) understand and other apis for ermrest uri (`Location.ermrsetUri`, `Location.ermrestPath`, etc.) are introduced.

Issue: #455

### DIFF
--- a/js/parser.js
+++ b/js/parser.js
@@ -123,6 +123,7 @@ var ERMrest = (function(module) {
         // compactPath is path without modifiers
         this._compactPath = (modifiers === "" ? this._path : this._path.split(modifiers)[0]);
 
+        // <sort>/<page>
         // sort and paging
         if (modifiers) {
             if (modifiers.indexOf("@sort(") !== -1) {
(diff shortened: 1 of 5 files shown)


js/parser.js
```diff
@@ -145,52 +146,45 @@ var ERMrest = (function(module) {
         }
 
         // Split compact path on '/'
-        // Expected format: "[schema_name:]table_name[/{attribute::op::value}{&attribute::op::value}*]"
+        // Expected format: "<schema:table>/<filter(s)>/<joins(s)/<search>"
```
jrchudy: Not that this matters much but you forgot a `>` after `join(s)`

js/parser.js
```diff
@@ -283,25 +277,144 @@ var ERMrest = (function(module) {
...
         toString: function(){
-            return this._uri;
+            return this.uri;
         },
 
 
         /**
-         *
+         * <service>/catalog/<catalogId>/<api>/<projectionSchema:projectionTable>/<filters>/<joins>/<search>/<sort>/<page>/<queryParams>
          * @returns {String} The full URI of the location
```
jrchudy: How does this URI differ from `ermrestUri`? Could you mention that in the comments here.

js/parser.js
```diff
@@ -283,25 +277,144 @@ var ERMrest = (function(module) {
...
+            if (this._uri === undefined) {
+                this._uri = this.compactUri + this._modifiers;
+            }
             return this._uri;
         },
 
         /**
-         *
+         * <service>/catalog/<catalogId>/<api>/<projectionSchema:projectionTable>/<filters>/<joins>/<search>
          * @returns {String} The URI without modifiers or queries
```
jrchudy: Same comment here. How does this compact URI differ from `ermrestCompactUri`?

RFSH:
@jrchudy Added some comments to the apis. The only difference between `ermrestUri` and `uri` is the search operator right now. `uri` might have `*::search::` in it, but `ermrestUri` shouldn't. However, they will be more and more different by introducing facets.

[There is a test case](https://github.com/informatics-isi-edu/ermrestjs/blob/master/test/specs/reference/tests/13.search.js#L412) for that in ermrestjs and which expects the existing behavior (What recordset is showing in the textbox is `searchTerm` and this expects the `searchTerm` to include the regular expression). We can add some test cases for it in chaise too. We don't need to go to next page and get back, we can just click on `Permalink` and then check the search box and url.

PR #459, opened 2017-07-20
Change asset formatPresentation heuristics

RFSH:
Previously markdown_pattern had priority over raw values in entry context, I changed it to always return the raw value in entry contexts


jrchudy:
I approved this to be merged except a test case failed.. So we should address that first.

RFSH:
@jrchudy I restarted the build and it passed. Merging it.

PR #460, opened 2017-07-21
Changes to avoid regression in JSON value formatting in templates

nitishksahu:
The `<pre>` tags were visible in the UI because the function doing the markdown templating was getting the formatted JSON value to display in the UI. These changes now allow ermrestJS to template a value first and then append a`<pre>` tag to display in the UI.

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -2873,9 +2873,8 @@ var ERMrest = (function(module) {
                         //Added this if conditon explicitly for json/jsonb because we need to pass the
                         //formatted string representation of JSON and JSONBvalues
                         else if (column.type.name === "json" || column.type.name === "jsonb") {
-                            presentation = column.formatPresentation(keyValues[column.name], { formattedValues: keyValues , context: this._pageRef._context });
-                            this._values[i] = presentation.value;
-                            this._isHTML[i] = presentation.isHTML;
+                            this._values[i] = keyValues[column.name];
+                            this._isHTML[i] = false;
                         }
                         else {
                             this._values[i] = this._data[column.name];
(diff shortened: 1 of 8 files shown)


js/reference.js
```diff
@@ -3258,7 +3262,15 @@ var ERMrest = (function(module) {
             }
             return {isHTML: isHTML, value: value};
         },
-
+        /**
+         * Append a <pre> tag to the value so that the beautified JSON can be shown including the whitespaces.
+         * @param {String} value This is the formatted string that we get after stringifying
+         * @returns {String} A string representation of the stringified value enclosed in a <pre> tag
+         */
+        appendPreTag: function appendPreTag(value) {
```
RFSH: You've just used this function one time, and also this function doesn't make any sense for column. Remove it and just append the text instead of calling a function.

Remember whatever apis that you are exposing here might be used by client (chaise), so if the api that you are creating doesn't make sense to client you shouldn't expose it.

js/reference.js
```diff
@@ -2900,7 +2899,12 @@ var ERMrest = (function(module) {
                         } else {
                             values[i] = column.formatPresentation(keyValues[column.name], { formattedValues: keyValues , context: this._pageRef._context });
                             // If the column type is json or jsonB we will send the templated string with <pre> tag
-                            if (column.type.name === "gene_sequence"|| column.type.name === "json" || column.type.name === "jsonb") {
+                            if(column.type.name === "json" || column.type.name === "jsonb"){
```
RFSH: Also add a check for isHTML. We only want to append `pre` tags when output is not html.

test/specs/reference/tests/12.reference_values_edit.js
```diff
@@ -330,6 +330,7 @@ exports.execute = function (options) {
                     var jsonb=JSON.stringify(expectedValues[i].jsonb_col,undefined,2);
                     expect(values[1]).toBe(json);
                     expect(values[2]).toBe(jsonb);
+                    expect(values[3]).toBe(expectedValues[i].json_col_without_markdownpattern);
```
RFSH: Isn't the name of column `json_col_with_markdownpattern`? It shouldn't matter here since it's just a object that you're using internally for tests, but it's confusing that it says "without_markdown" while the column is with markdown.

test/specs/reference/tests/12.reference_values_edit.js
```diff
@@ -330,6 +330,7 @@ exports.execute = function (options) {
                     var jsonb=JSON.stringify(expectedValues[i].jsonb_col,undefined,2);
                     expect(values[1]).toBe(json);
```
RFSH: This is something that I should have mentioned in your previous PR. But when in a single `it` there are multiple `expect`s, add a message for each of them so you can understand why it failed. For example for this one it can be
```javascript
expect(values[1]).toBe(json, "mismatch in tuple with index=" + i + ", column=json_col");
```
Change other test cases accordingly.

Issue #450, opened 2017-07-11
facet APIs with default visible-facet list


hongsudt:
@hongsudt commented on [Tue Jul 11 2017](https://github.com/informatics-isi-edu/chaise/issues/1185)

If annotation is not specified, the default facet list should consists of 
- all visible columns in the compact context
- all the in-bound fks visible in the detailed context

Issue #449, opened 2017-07-11
create applyFilters function 


hongsudt:
Given a list of filters, return a new filtered reference.

hongsudt:
This is no longer relevant since we want the UX to be more adaptive..

Issue #447, opened 2017-07-10
Parser extension to support facets/filters


RFSH:
To make it easier for us to pass the filters and parse them, we decided to encode the filters in chaise, then ermrestjs decodes it and creates a uri that ermrest understand. Therefore filter will be a JSON object. To make it easier to find this filter part, we can introduce a new operator, i.e. `::filters::`.

## Structure

The overall structure of filters is as follows. Each facet term combines a data source and constraint spec, and so all combinations of constraint kind and data source path are possible in the syntax.

```
<FILTERS>:  { <logical-operator>: <TERMSET> }

<TERMSET>: '[' <TERM> [, <TERM>]* ']'

<TERM>:     { <logical-operator>: <TERMSET> } 
            or
            { "source": <data-source>, <constraint(s)> }
```

In the following sections each of location operators, data source, and constraints are explained.

## Logical operators

We want the structure to be as general as possible, so we don't need to redesign the whole structure when we need to support more complex queries. Therefore as the top level, we can have logical operators.

    { "and": [ term... ] }
    { "or": [ term... ]}
    { "not": term }

## Data Source

The difference between base table columns and columns in related entities is just a difference in data source. It shouldn't involve a completely different set of filter structures.  Instead, the single filter structure should conceptually allow various forms of data source specification

    "*"
    "Column1"
    [{"inbound": ["S1", "FK1"]}, "Column2"]
    [{"inbound": ["S1", "FK1"]}, {"outbound": ["S2", "FK2"]}, "Column3"]

The simple form `"Column1"` might be allowed as a short-hand for `["Column1"]` since it isn't ambiguous. But any foreign key path also needs to end with the actual column that will be projected and filtered. The _direction_ field labels `"inbound"` and `"outbound"` remove any ambiguity for self-referencing table navigation scenarios. The constraint name pairs `["S1", "FK1"]` etc. reuse the same names appearing in the ERMrest model introspection document.

Even if we are faceting on a vocabulary concept and just want the user to pick values by displayed *row name* and we substitute the actual entity keys in the ERMrest query, I think we should record this column choice explicitly in the facet spec, e.g. `[{"inbound": ["S1", "Fk1"]}, "id"]` so that the resulting faceting app URL is unambiguous even if there have been subtle model changes in the interim, which might change the default key selection heuristics etc.

> Based on this, we are not supporting filtering on foreign keys with composite keys.

### entity v.s. scalar facet
[...]

PR #464, opened 2017-07-25
Facet Support

RFSH:
This will add faceting support to ermrestjs. This includes:

1. Add `::facets::` pseudo-operator to the parser. Ability to decode to JSON and encode again.
2. Add `FacetColumn` class with `FacetFilters` which has functions to manipulate the filter list.
3. Add `Reference.facetColumns`API which currently is returning list of visible column + related entities.
4. Add `Reference.applyFacets` which will return a new reference with the applied facets.

--------------------------
This PR is just for easing the review process. The task is not complete yet, I need to 
- [x] Use a better decode/encode algorithm for JSON.
- [ ] Add test cases.

Related Issues: #447, #449,  #450


RFSH:
My implementation is flawed. I'm mutating the filters which I shouldn't. I am closing this PR to make the changes and create a new one later.

PR #473, opened 2017-07-28
JSON column default value changes 

nitishksahu:
The JSON column for any table will now store a null value for all the blank string value.

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -551,24 +551,11 @@ var ERMrest = (function(module) {
                             notSet = false;
                             break;
                         }
-                    } 
+                    }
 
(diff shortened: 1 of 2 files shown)


js/reference.js
```diff
@@ -551,11 +551,24 @@ var ERMrest = (function(module) {
                             notSet = false;
                             break;
                         }
-                    }
+                    } 
 
                     if (notSet) defaults.push(columnName);
                 });
-
+
```
RFSH: The following lines of code are just removing any JSON columns from default list. This is incorrect. If it's `undefined` we should add the column to list of defaults. Also it's not doing what the comment suggests (changing "" to null).

PR #474, opened 2017-07-31
Faceting Support

RFSH:
After reviewing #464, we noticed that I was mutating the Reference object. This PR is changing that implementation so that we won't mutate Reference object. Functions that manipulate the facets will return a new Reference with the new set of filters.

To summarize the set of changes, this PR,

1. Adds `::facets::` pseudo-operator to the parser. 
     - Parser understands this pseudo-operator and can decode/encode the value of it. 
     - Parses can parse the simplified version of JSON specified in #447, and create the appropriate ermrest filters.
2. Adds `FacetColumn` class
   - Has knowledge of its `Reference` and `ReferenceColumn`.
   - Has different functions for manipulating the list of `filters`. These function will return a new `Reference` instead of mutating the current `Reference`.
3. Adds `Reference.facetColumns`API which currently is returning list of visible column in compact context + related entities in detailed context.
4. Adds `Reference. removeAllFacetFilters ` to remove all the filters.

Related Issues: #447, #449, #450

P.S.
- The build is failing because of the "defaults" test cases that I mentioned in the standup meeting.
- I created a very limited faceted search in chaise using these apis, which you can access [here](https://dev.isrd.isi.edu/~ashafaei/chaise/recordset/#1/isa:dataset). It's only supporting adding choice filter and removing them. It was solely created for the purpose of making sure these apis are good to go.
- Test cases are missing. I'm not sure whether I should prioritize test cases or chaise support.


js/reference.js
```diff
@@ -4229,6 +4344,375 @@ var ERMrest = (function(module) {
...
+            fc.filters = filters;
+            
+            var newReference = _referenceCopy(this.reference);
+            
+            // clone the location object
+            newReference._location = this.reference._location._clone();
+            
+            // make sure reference.facetColumns and newRef.facetColumns are not referencing the same thing
+            delete newReference._facetColumns;
+            newReference._facetColumns = this.reference.facetColumns.slice();
```
jrchudy: change to _facetColumns

jrchudy:
The changes look good from what we discussed earlier and after I looked over the code some more.

PR #478, opened 2017-08-04
Format presentation changes for JSON column

nitishksahu:
Changes in format presentation function, the formatting logic has been removed from the get value() function and the JSON condition has been added to formatPresentation function()


RFSH:
Please add a test case to make sure the connected issue is resolved. And also always assign the PRs that you create to yourself.

Issue #461, opened 2017-07-21
Add fallback heuristics for domain types


karlcz:
When a column has a domain type like `markdown` there is also a base type like `text`.

When we don't have a presentation preference for the domain type, we should fall back to using any presentation preference for the base type.  When all else fails, we should continue to use a basic text presentation as we do now.

In upcoming changes to deployments, there will be several new domains with base types of `int8` and `timestamptz` and we want to get the appropriate display formatting for those rather than treating these as plain text.

jrchudy:
Another piece of this PR is to add a case in `ermrestJS` in `Column.formatvalue` to handle this.

jrchudy:
The changes for this issue have been merged into master. When a column has a domain defined and we don't know how to format data for that domain type, we check what it's base type is for formatting.

Reassigning to @karlcz to verify the changes are working as he expected for the synapse deployment.

karlcz:
I think this is working now.

PR #483, opened 2017-08-10
Added base type to the type definition.

jrchudy:
This PR addresses issue #461.

When a base type is present on the type in the jsonColumn, it is now set on the type.

### DIFF
--- a/js/core.js
+++ b/js/core.js
@@ -1672,7 +1672,7 @@
                 return this._getNullValue(options ? options.context : undefined);
             }
 
-            return _formatValueByType(this.type, data, options).toString();
+            return _formatValueByType(this.type, data, options);
         };
 
         /**
--- a/js/core.js
+++ b/js/core.js
@@ -2923,7 +2923,7 @@
          */
         this._isDomain = jsonType.is_domain;
 
-        if (this._isDomain || this._isArray) {
+        if (jsonType.base_type !== undefined) {
             /**
              * @type {ERMrest.Type}
              */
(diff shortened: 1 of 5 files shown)


RFSH:
Didn't we wanted to change the `formatvalue` function accordingly based on the basetype? Or that is going to be another PR? 
Also we can have very simple test cases [here](https://github.com/informatics-isi-edu/ermrestjs/blob/master/test/specs/common/tests/02.column.js).

jrchudy:
We are just exposing the base type if it exists right now. 

Any changes to start managing this in `formatvalue` will be handled in another PR when that issue comes up. I will check with @hongsudt before merging these changes.

RFSH:
We started writing unit tests after implementing the core.js. The test cases in the file that I gave you are for any features that we added to core.js. Eventually we might need to test all the features of the core.js, But at least we should test whatever we're adding.

jrchudy:
@RFSH I factored out the switch statement so it can be used recursively. I can't test this functionality with unit tests currently without modifying a schema through `psql` to define a new domain type.

js/core.js
```diff
@@ -1669,52 +1669,10 @@
...
-                case 'jsonb':
-                    data = utils.printJSON(data);
-                    break;
-                default: // includes 'text' and 'longtext' cases
-                    data = utils.printText(data, options);
-                    break;
-            }
-            return data.toString();
+
+            return _formatValueByType(this.type, data, options).toString();
```
RFSH: Why are you calling the `toString` here? If the function is expected to return string (based on jsdoc its return type is string), it should be inside that function and not here.

Issue #482, opened 2017-08-09
Reference For Attributegroup API


RFSH:
We are going to introduce `aggregateGroup` APIs which will return a paginated, sortable and searchable result. We cannot reuse the same `Page` object because it's based on Reference. Since current implementation of Reference is based on ermrest entity api. We need a new type of Reference that is based on ermrest attributegroup api.

We might need to implement new Page and Tuple prototypes based on this new Reference too.

PR #484, opened 2017-08-10
attributegroup reference skeleton

RFSH:
Related Issue: #482

### DIFF
--- a/js/ag_reference.js
+++ b/js/ag_reference.js
@@ -1,0 +1,242 @@
+/*
+ * Column and Refernce has a lot of arguments, I don't know how to fix that
+ * The same way we have the location object, we can have something like that here,
+ * which has the service, catalog, path, sort, paging, and search in it. but should we?
+ */
+
+function AttributeGroupReference(service, catalog, path, keyColumns, aggregateColumns, sort, paging) {
+    
+    /**
+     * The uri to ermrest service
+     * @type {string}
+     */
+    this._service = service;
+    
+    /**
+     * id of the catalog
+     * @private
+     * @type {stirng}
+     */
+    this._catalogId = catalog;
+    
+    /**
+     * The path that will be used for generating the uri in read.
+     * @type {string}
+     */
+    this._path = path;
+    
+    /**
+     * Array of AttributeGroupColumn that will be used as the key columns
+     * @type {ERMrest.AttributeGroupColumn[]}
+     */
+    this._keyColumns = keyColumns;
+    
+    /**
+     * Array of AttributeGroupColumn that will be used for the aggregate results
+     * @type {ERMrest.AttributeGroupColumn[]}
+     */
+    this._aggregateColumns = aggregateColumns;
+    
+    /**
+     * The sort object. It will be an array of object with the following format:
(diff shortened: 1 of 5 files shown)


js/ag_reference.js
```diff
@@ -0,0 +1,299 @@
...
+    
+    constructor: AttributeGroupReference,
+    
+    get location () {
+        return this._location;
+    },
+    
+    /**
+     * the displayname of the reference
+     * TODO not sure if this sis needed
```
jrchudy: We will need it for when there is a picker popup in the facet.

Maybe use the `table.name` and if that's not defined (chances are it won't), we should have fallback heuristics. Maybe displayname should be related to what aggregate was called?

js/ag_reference.js
```diff
@@ -0,0 +1,299 @@
...
+        if (this._isHTML === undefined) {
+            var value = this.values;
+        }
+        return this._isHTML;
+    },
+
+    get values() {
+        //TODO 
+        // returned value should be the same as current tuple.values 
+        // or we should make this (https://github.com/informatics-isi-edu/ermrestjs/issues/189) change now?
```
jrchudy: I think we do want to move toward this approach with `tuple.values` being an array of objects like:
```
{
  value: <value>,
  isHtml: boolean,
  isMarkdown: boolean
}
```

Karl commented that we should add `isMarkdown`. I'm not sure whether or not he meant to replace `isHtml` with `isMarkdown` or if they should both be defined. I'm trying to think of a case where we define a value as HTML that is not also Markdown.

js/ag_reference.js
```diff
@@ -0,0 +1,299 @@
...
+function AttributeGroupColumn(alias, name, displayname, type, comment, sortable, visible) {
+    /**
+     * @private
+     * @type {string}
+     */
+    this._alias = alias;
+    
+    /**
+     * This might include the aggregate functions,
+     * should we seperate those?
```
jrchudy: Not sure about this.

js/ag_reference.js
```diff
@@ -0,0 +1,299 @@
...
+        return this._name;
+    },
+    
+    formatvalue: function (data, options) {
+        //TODO should be the same as Column.formatvalue
+        notimplemented();
+    },
+    
+    formatPresentation: function (data, options) {
+        //TODO take cares of markdown_pattern and all, is it needed?
```
jrchudy: I don't know how the user would define a markdown pattern for one of these columns. We would have to design and annotation for it as well.

js/ag_reference.js
```diff
@@ -0,0 +1,299 @@
...
+    /**
+     * Represents the paging. It will be in the following format:
+     * {"before":boolean, "row":[v1, v2, v3...]}
+     * v1, v2, v3.. are in the same order of columns in the sortObject
+     * 
+     * @type {?Object}
+     */
+    this.pagingObject = paging;
+}
+AttributeGroupLocation.prototype = {
```
jrchudy: Maybe add `get url()`?

js/ag_reference.js
RFSH: Use term instead of name.

Issue #476, opened 2017-08-02
Faceting Changes


RFSH:
The first version for Faceting support (#474) has been merged. That implementation included some incorrect assumptions that we should address. 

I assumed that `FacetColumn`s will wrap `ReferenceColumn`s. That means a FacetColumn can be a normal column, inbound related or outbound related (I assumed this is the requirement for the first iteration of faceting). But we want Faceting to support more than one or two joins. The following are the properties of `FaceColumn`:
 - `filters` The same as it is.
 - `reference` The reference that this `FacetColumn` belongs to.
 - `sourceReference` The reference that has all the joins specified in the `source` and also all the filters of other `FacetColumn`s in the reference. This reference will be used for domain filtering in the facet (getting the aggregated values so user can pick based on those).
 - `column`: This will be a `ReferenceColumn` but it only will capsulate the last column in the `source` array. It should be created based on `sourceReference` (it will be used for showing the displayname and also column level aggregates).
- `display`: This will show the prefered display mode. It will eventually be something that annotation dictates and chaise can use. First implementation should only return `entity` or `scalar`. It will be `entity` if the column is a key column.

RFSH:
Based on the discussion that I had with @hongsudt, we decided to add a `displayname` to the `FacetColumn`. This will eventually use annotation to get the displayname, but currently this is going to be the heuristic:

1. Use `from_name` and `to_name` of  last foreign key based on direction (`from_name` for inbound and `to_name` for outbound).
2. If it's not defined, use the table name.

> If it's in `scalar` mode append the column name (`table_name (col_name)` ).

RFSH:
We also should have a `comment` attribute which will be used for tooltip. It most probably should be the "comment" of last foreign key.

PR #485, opened 2017-08-15
Faceting

RFSH:
This PR includes, 
 - Changes described in #476.
-  Changes to the structure of `source` as described in #447. It will also add more APIs for getting the preferred ux mode, and `isEntityMode`.

These changes require a bit of refactoring and eventually testing.

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -4426,11 +4426,15 @@
      * Based on facets JSON structure we can have joins that result in facets
      * on columns that are not part of reference column.
      *
+     * TODO This is just experimental, the arguments might change eventually.
+     * 
      * If the ReferenceColumn is not provided, then the FacetColumn is for reference
      *
      * @param {ERMrest.Reference} reference the reference that this FacetColumn blongs to.
      * @param {int} index The index of this FacetColumn in the list of facetColumns
-     * @param {obj|string} json the filter object
+     * @param {ERMrest.Column} column the column that filters will be based on.
+     * @param {object} json The filter object that this FacetColumn will be created based on
+     * @param {ERMrest.FacetFilter[]} filters Array of filters
      * @constructor
      */
     function FacetColumn (reference, index, column, json, filters) {
@@ -4571,7 +4575,7 @@
                         self = this;
                     
                     // create a path from this facetColumn to the base reference
-                    if (this._hasJoin) {
+                    if (Array.isArray(this.dataSource)) {
                         this.dataSource.forEach(function (ds, index, arr) {
                             // last elemenet is the column name
                             if (index !== arr.length -1 ) {
--- a/js/reference.js
+++ b/js/reference.js
@@ -4514,6 +4514,12 @@
          * This should be used if we're not in entity mode.
          * TODO: what should be the default? This will eventually change,
          * currently we are not using multi facet mode, so it won't be used.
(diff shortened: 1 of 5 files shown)


js/reference.js
```diff
@@ -4710,6 +4987,30 @@
...
+    }
+    module._extends(EntityFacetFilter, FacetFilter);
+    
+    /**
+     * String representation of entity filter. It will be the tuple displayname
+     * 
+     * @return {string}
+     */
+    EntityFacetFilter.prototype.toString = function () {
+        //TODO: should it be unformatted or not?
```
jrchudy: This will depend on what the `toString` is going to be used for. If it's for display purposes then yes it should be formatted.

js/reference.js
```diff
@@ -4350,75 +4428,266 @@
...
+        },
+        
         /**
-         * data source of this facet. Can be used to find filters for this facet
-         * @return {String} [description]
+         * The Preferred ux mode.
+         * Any of:
+         * `choices`, `range`, or `search`
+         * This should be used if we're not in entity mode.
+         * TODO: what should be the default? This will eventually change,
```
jrchudy: If we want to consider a default mode for facets for any column type, we might want to keep it simple and have the default mode show as `choices`. `Search` mode would imply that the user needs to be aware of the whole set of values they are searching through. Choices provides some of that information for them.

js/reference.js
```diff
@@ -4350,75 +4428,266 @@
...
+         * @type {ERMrest.Reference}
          */
-        this.column = refColumn;
-
         this.reference = reference;
-
+        
+        /**
+         * The index of facetColumn in the list of facetColumns
+         * NOTE: Might not be needed
```
jrchudy: I don't think we ever use the index on the `Column` object in `core.js`.

js/reference.js
```diff
@@ -4350,75 +4428,266 @@
...
+     * @param {ERMrest.Reference} reference the reference that this FacetColumn blongs to.
+     * @param {int} index The index of this FacetColumn in the list of facetColumns
+     * @param {obj|string} json the filter object
      * @constructor
      */
-    function FacetColumn (reference, refColumn, index) {
-        verify(!refColumn.isAsset && !refColumn.isKey, "Facet Column cannot be asset or key pseudo-column.");
-        verify(!refColumn.isForeignKey || refColumn.foreignKey.simple, "Facet Column does not support composite foreign keys.");
-
+    function FacetColumn (reference, index, column, json, filters) {
```
jrchudy: You are missing documentation info for `column` and `filters`.

js/reference.js
RFSH: This is something that I have used in the implementation of `FacetColumn`. The Note is just for when I get to refactor it, we might not need to pass this index around

js/reference.js
RFSH: Yeah exactly. Most of these TODOs for me to not forget to change it when I get to actually use it in chaise. That was one of the reasons that I didn't wanted to create this PR. There are a bunch for TODOs that should be addressed but cannot be addressed until we get to integrate it with chaise.

RFSH:
I added your comments to the code and also a little more documentation. We will address these TODOs when we get to integration part with chaise.

PR #496, opened 2017-09-13
Added support for es6 using Babel

chiragsanghvi:
This PR adds support to use basic ES6 features like arrow functions, spread operators and parameter destructuring.


Makefile
```diff
@@ -75,19 +75,19 @@ $(BUILD)/$(VER): $(SOURCE)
 	mkdir -p $(BUILD)
 	git log --pretty=format:'%H' -n 1 > $(BUILD)/$(VER)
 
-# Rule to build the un-minified library
+# Rule to build the un-minified library and convert es6 code to es5
```
robes: I don't see any change to the recipe for un-minified code. I would expect that it needs to be compiled here by babel.

chiragsanghvi: I missed the line. I have added it back. In addition if we don't plan on having Node installed then we might need to get rid of the `jshint` library too. As it is also run when we build the code.

Just having Node as part of our build process would be great. New `es6` features are pretty useful and reduces effort spent in handling scope and parameters.

RFSH:
Closing this PR because our codebase has changed a lot ever since and also new and better ways of allowing es6+ are available.

Issue #487, opened 2017-08-21
row-level ermrest ACL support


hongsudt:
@karlcz what are the steps do we need to do to migrate to the new ACL in ermrest?

karlcz:
If you are on latest ermrest master you already have the fine-grained
ACL features.  You can see evidence of them as various "rights"
documents embedded inside the main /ermrest/catalog/N/schema/ response
document.

To make the front-end adapt to fine-grained rights, it should start to
consult those "rights" documents on each part of the model. At the
table level:

- select: is client allowed to query the table
- insert: is client allowed to insert new rows
- update: is client allowed to edit existing rows
- delete: is client allowed to delete existing rows

The answers can be true/false/null.  A null means that the access
decision is data-dependent (i.e. row-level policy) and cannot be
advertised statically for the entire table.  It may succeed or fail
on specific rows.

The column-level version of these rights says whether that particular
column can be addressed in select, insert, or update operations. You
might map this to ermrestjs concepts like:

- insert false: column is "generated" during row insert
- update false: column is "immutable" during row edits

Similarly, if column-level select rights are false, the column should
be dropped from any visible columns list since it will cause a
forbidden error if included.

hongsudt:
From the meeting today, we decided to: 
- [ ] handle table right policy for canCreate, canUpdate, canDelete, canRead
  - null value should be treated as true.
- [ ] column level policy
  - insert false: column is "generated" during row insert 
  - update false: column is "immutable" during row edits
  - select false: show `--hidden--` for the formatted value
- [ ] rename canCreate to canInsert to be consistent with ermrest terms
- [ ] rename canRead to canSelect to be consistent with ermrest terms

hongsudt:
take a look at the epic associated with this issue. Karl has provided information on the new right in the epic.

chiragsanghvi:
Select for columns will be taken care by ermrest as mentioned in this issue https://github.com/informatics-isi-edu/ermrest/issues/167

PR #512, opened 2017-10-06
Acls

chiragsanghvi:
This PR adds support for ACL's in ermrestjs as mentioned in this issue https://github.com/informatics-isi-edu/ermrestjs/issues/487. 

For now we don't handle `select` case for columns. This is something that will be handled in ermrest.


js/core.js
```diff
@@ -1780,6 +1796,30 @@
...
+         *
+         * @type {Object}
+         */
+        this.rights = jsonColumn.rights;
+
+        /**
+         * Mentions whether we should hide the value for this column
+         * @type {Boolean}
+         */
+        this.isHidden = this.rights.select === false;
```
RFSH: It seems like you are not using this variable in the implementation of ermrestjs. It's only being used in test cases.

js/utilities.js
```diff
@@ -1780,3 +1780,5 @@
     ];
 
     module._systemColumns = ['RID', 'RCB', 'RMB', 'RCT', 'RMT'];
+
+    module._HIDDEN_TEXT = "--hidden--";
```
RFSH: This is not used anywhere.

test/specs/reference/tests/06.permissions.js
```diff
@@ -1,8 +1,44 @@
-var nock = require("nock");
+var ermrestImport = require(process.env.PWD + '/test/utils/ermrest-import.js');
```
RFSH: I think it's easier if I ask these questions in person, and If you can walk me through these test cases. But I'm writing my questions here just for a reference:

- Why are you changing ACLs for in `afterAll` and `beforeAll` for old test cases? Aren't those test cases only working based on annotation?
- This test spec is difficult to read. Can you break it into multiple specs? or at least have a top level `describe` that differentiates the test cases that are only based on annotation, and the test cases that are based on new ACLs?
- You've used arrow functions in this file which is ES6, I have nothing against it. But we should discuss about it in the group. It's odd that one file is using ES6.

test/utils/ermrest-import.js
```diff
@@ -43,6 +43,20 @@ exports.importSchemas = function(configFilePaths, catalogId) {
 	return defer.promise;
 };
 
+exports.importAcls = function(params) {
```
RFSH: I know it's obvious what this function is for, but can you add some comments to it. These utils functions are lacking documentation. Also the indentation in line 50 and 51 are off.

test/specs/reference/tests/06.permissions.js
```diff
@@ -1,8 +1,44 @@
...
+
+    var removeCachedCatalog = (catalogId) => {
+        var server = options.ermRest.ermrestFactory.getServer(process.env.ERMREST_URL);
+        delete server.catalogs._catalogs[catalogId];
+    };
+
+    describe("For determining reference object permissions,", () => {
 
-exports.execute = function (options) {
+        var restrictedUserId = process.env.RESTRICTED_AUTH_COOKIE_ID,
```
RFSH: Please comment that `RESTRICTED_AUTH_COOKIE_ID` is something that the test framework is generating (I was confused at first that I might have to set it up myself).

test/specs/reference/tests/06.permissions.js
chiragsanghvi: We need to change the permissions as the restrictedUser doesn't has permissions to do anything. So when we call `canCreate` the permissions are also checked with the annotations, thats the reason I had to change them. Moreover I didn't want to test it in context of admin user. Admin already has all the rights.

Secondly,  I used the arrow functions only in test-specs. These tests are run in `node.js`. We should gradually change all test specs as we make changes to them to make development easier.

js/core.js
chiragsanghvi: I was using these variables before when I had the select=false handling. But after discussing things with Karl I removed the logic and forgot to remove these variables.

This variable will be useful when I need to remove the column during introspection.

js/utilities.js
chiragsanghvi: Same reason as mentioned above for isHidden. I will get rid of this variable though.

test/specs/reference/tests/06.permissions.js
chiragsanghvi: Yes, I will add the comment.

test/utils/ermrest-import.js
chiragsanghvi: They're just proxy functions. We already have documentation for them in ermrestDataUtils. Thats the reason they don't have any documentation.

chiragsanghvi:
Made the changes as requested

PR #518, opened 2017-10-18
Renamed canRead to canSelect and canCreate to canInsert

chiragsanghvi:
This PR renames `canRead` to `canSelect` and `canCreate` to `canInsert` to be more close to actual ACL policies as mentioned [here](https://github.com/informatics-isi-edu/ermrestjs/issues/487).

Merge Chaise [PR](https://github.com/informatics-isi-edu/chaise/pull/1341) immediately after merging this PR to avoid build problems.


RFSH:
I know the discussion was only about permissions. But shouldn't we change the `create` and `read` functions aswell? So we are using the same terminology all over the place.

PR #520, opened 2017-10-19
Added logic to disable system Columns input

chiragsanghvi:
This PR disables system columns in create and update context

https://dev.isrd.isi.edu/~chirag/chaise/recordedit/#35534/reference_schema:table_with_system_columns


test/specs/reference/tests/01.reference.js
```diff
@@ -6,6 +6,7 @@ exports.execute = function (options) {
             tableName = "reference_table",
             tableNameWithSlash = "table_w_slash",
             tableNameWithCompKey = "table_w_only_composite_key",
+            systemColumnTable = "table_with_system_columns",
```
RFSH: This is not needed since you added your test cases in another file.

chiragsanghvi: Removed

Issue #522, opened 2017-10-19
Make attribute date available to Mustache


chiragsanghvi:
Add date object to Mustache template variables. This object can be access as `$moment` in Mustache template.

It will contain following properties for date Thu Oct 19 2017 16:04:46 GMT-0700 (PDT)

* date: 19
* day: 4
* month: 10
* year: 2017
* dateString: Thu Oct 19 2017
* hours: 16
* minutes: 4
* seconds: 14
* milliseconds: 873
* timeString: 16:04:46 GMT-0700 (PDT)
* ISOString: 2017-10-19T23:04:46.873Z
* GTMString: Thu, 19 Oct 2017 23:04:46 GMT
* UTCString: Thu, 19 Oct 2017 23:04:46 GMT
* LocaleDateString: 10/19/2017
* LocaleTimeString: 4:04:46 PM
* LocalString: 10/19/2017, 4:04:46 PM

We discussed about exposing date values in `$moment` object to the Mustache template, which can be referred to as `{{$moment.month}}` for month in the template.

hongsudt:
Can you add this to the markdown template document?

PR #523, opened 2017-10-19
Added date support in Mustache template exposing it as $moment

chiragsanghvi:
This PR adds date support in Mustache templates as mentioned [here](https://github.com/informatics-isi-edu/ermrestjs/issues/522) by exposing `$moment` object which can be easily accessed in templates for current date and time.

### DIFF
--- a/js/utilities.js
+++ b/js/utilities.js
@@ -1510,6 +1510,54 @@
         };
     };
 
+    /**
+     * @function
+     * @desc
+     * Gets currDate object once the page loads for future access in templates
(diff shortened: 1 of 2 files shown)


jrchudy:
This looks good and the test looks right. But I don't fully understand the logic enough to approve it..

chiragsanghvi:
@jrchudy The logic is simple. We are injecting a new object `$moment` in the template environment similar to the `encode` and `decode` functions.

PR #532, opened 2017-11-02
Made columns with asset annotation that dont have hatrac set in  url pattern hidden in edit context

chiragsanghvi:
This PR removes the asset column from table in entry context if the url_pattern doesn't has hatrac  in it.


js/reference.js
```diff
@@ -2189,15 +2189,27 @@
...
-                        // add asset annotation
-                        var assetCol = new AssetPseudoColumn(self, col);
-                        assetColumns.push(assetCol);
-                        self._referenceColumns.push(assetCol);
-                        return;
+                        
+                        var urlPattern = col.annotations.get(module._annotations.ASSET).content.url_pattern;
+                        
+                        // If url_pattenr doesn't has hatrac in it then consider the column as a plain text column and proceed
+                        if ((typeof urlPattern !== 'string') || (urlPattern.indexOf('/hatrac') === -1)) {
```
RFSH: Please use `console.debug` instead, and there are some grammatical errors. (doesn't has/ doesn't contains)

chiragsanghvi: Node.js has no `console.debug` method, so I used console.log instead.

RFSH: oh, I didn't know that.

chiragsanghvi:
I confirmed with Hongsuda, we should be ignoring the column in entry context. I have made the relevant changes.

PR #537, opened 2017-11-08
Added support to expose file mimetype in asset url_pattern

chiragsanghvi:
This PR adds support for file mimetype in asset url_pattern.

### DIFF
--- a/js/hatrac.js
+++ b/js/hatrac.js
@@ -646,6 +646,7 @@ var ERMrest = (function(module) {
 
         row[this.column.name].filename = this.file.name;
         row[this.column.name].size = this.file.size;
+        row[this.column.name].mimetype = this.file.type;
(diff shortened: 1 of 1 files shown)


RFSH:
I know that in the issue I only mentioned those two attributes, but is there any other file related attribute that can be exposed?
And please add this new attribute to the [annotation wiki](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2017-asset) after we merge the PR.

chiragsanghvi:
There aren't any more attributes apart from name, size and type which are supported in all browsers.

RFSH:
I think you need to add the same line for mimetype as [here](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/hatrac.js#L260,L267) to ignore that value in the validateURL.

Issue #545, opened 2017-11-13
facet list merging heuristics (url vs. annotation)


hongsudt:
Currently, if defaults are specified in annotation, they are included in the facet list (the results are combined e.g both defaults and url facets are applied). 

We shouldn't combine them. The url facets should take precedence.

In other words, if you have facets in your url we should ignore the default values for filter facets specified in the annotation.

PR #553, opened 2017-11-16
Facet List Merging Heuristics (url vs. annotation)

RFSH:
If chaise testcases  are based on previous heuristics we need to change them.

Issue: #545


RFSH:
Both chaise and ermrestjs test cases are passing locally. But they are failing on "sudo apt-get update" in travis. It looks like there are [some issues with travis build image](https://github.com/travis-ci/travis-ci/issues/8774).

RFSH:
@jrchudy It seems like travis issue has been resolved and tests are passing.

js/reference.js
```diff
@@ -610,11 +620,10 @@
                 // NOTE: current assumption: annotation is correct
                 if (annotationCols !== -1) {
                     usedAnnotation = true;
-                    //TODO should check for :
-                    // 1. duplicates ?! (not sure)
-                    // 2. correct values for choices, range, search
+                    //TODO should we check for duplciates? we're allowing having same sources but with differnt ux_modes
```
jrchudy: I think we should allow duplicates if they define it in their annotation.

PR #555, opened 2017-11-17
Formatted error

amitjha21:
This PR creates objects for **409** type error and provides `chaise` with the prototype to check them at front-end.


js/utilities.js
```diff
@@ -836,6 +836,57 @@
...
+                return word[0].toUpperCase() + word.slice(1);
+              }).join(' ');
+    };
+    /**
+     * @function
+     * @param  {string} errorStatusText    http error status text
+     * @param  {string} generatedErrMessage response data returned by http request
+     * @return {object}                    error object
+     */
+    module._conflictErrorMapping = function(errorStatusText, generatedErrMessage) {
```
RFSH: As karl mentioned in one of the meetings, these error messages might change.  Can you change some of the logics that if they are not as what you are expecting, they still produce meaningful messages?

Also add the expected format for each of those as a comment. So we can refer to that.

js/utilities.js
```diff
@@ -836,6 +836,57 @@
...
+      var mappedErrMessage;
+
+      if(generatedErrMessage.indexOf("violates foreign key constraint") > -1){
+          referenceTable = generatedErrMessage.match(/(?:^|\W)dataset(\w+)(?!\w)/g);
```
RFSH: You cannot assume that table name starts with `dataset`.

js/utilities.js
amitjha21: Made few changes so that in case of string literals not showing p; we can still have useful messages. However we are still reliant on message thrown for error processing.

RFSH:
10 days ago [I updated the description of what we need](https://github.com/informatics-isi-edu/chaise/issues/716#issuecomment-345412427). This is still not what we need. 
Remove the `CustomConstraintConflictError`. And the else case must be the default case of just showing the generic message.

RFSH:
Also make sure that your test cases are passing.

test/specs/errors/tests/07.http_content_error.js
```diff
@@ -44,6 +81,40 @@ exports.execute = function (options) {
...
+
+            server.catalogs.get(id).then(null, function(err) {
+
+                expect(err.code).toBe(409);
+                expect(err.message).toBe(deleteConflict);
+
+                done();
+            }).catch(function() {
+                expect(false).toBe(true);
+                done();
```
RFSH: These should be `done.fail()`

test/specs/errors/tests/07.http_content_error.js
```diff
@@ -18,7 +18,44 @@ exports.execute = function (options) {
...
+                    + '<head>'
+                        + '<title>409 Conflict</title>'
+                    + '</head>'
+                    + '<body>'
+                        + '<h1>Conflict</h1>'
+                        + '<p>409 Conflict The request conflicts with the state of the server. Input data violates model. ERROR:  duplicate key value violates unique constraint "dataset_pkey" DETAIL:  Key (id)=(269) already exists.</p>'
+                    + '</body>'
+                + '</html>';
+
+            var htmlCustomConstraintResponseMessage = '<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">'
```
RFSH: Change this to general response message.

test/specs/errors/tests/07.http_content_error.js
```diff
@@ -18,7 +18,44 @@ exports.execute = function (options) {
...
+                        + '<title>409 Conflict</title>'
+                    + '</head>'
+                    + '<body>'
+                        + '<h1>Conflict</h1>'
+                        + '<p>Error 409 Conflict The request conflicts with the state of the server. ERROR: the provided site_name is not consistent with your login profile. Please enter an appropriate site CONTEXT: PL/pgSQL function experiments.userid_update() line 25 at RAISE</p>'
+                    + '</body>'
+                + '</html>';
+
+
+        var terminalErrorMessage = "An unexpected error has occurred. Please report this problem to your system administrators.",
```
RFSH: This is not needed.

test/specs/errors/tests/07.http_content_error.js
RFSH: Also add a test case for this one.

RFSH:
@amitjha21 Some of the logic that you have added still had some assumptions about the database. I removed those and also added some extra cases to make sure that if the error message changes we're still going to handle it. As I mentioned earlier I moved the test cases to make it more readable, and also adding the `persist` solved the issue of nock behaving unexpectedly.

Issue #536, opened 2017-11-07
Add client context header for logging


karlcz:
See informatics-isi-edu/ermrest#177 for more discussion.

We want to replace our previous bag of query parameters for logging properties with a header containing a bag of attribute-value pairs.

1. Have some mechanism to collect the bag of parameters as an object
2. Encode the object in JSON w/o any pretty-printing (compact out all optional whitespace).
3. Encode the JSON string with a modified version of url-encoding to make it safe for HTTP headers
4. Send the encoded string as the body of a new `Deriva-Client-Context` header

The url-encoding will be similar to that in RFC 3986 except we can leave some "reserved" characters in their native ASCII form.  Consider these char classes safe to pass unescaped:
- alpha: `A` - `Z` and `a` - `z`
- digit: `0` - `9`
- non-reserved punctuation: `-`, `_`, `.`, `~`
- JSON and HTTP safe reserved chars: `{`, `}`, `"`, `,`, `:`

Everything else will be encoded byte-by-byte into the `%hh` hex representation, including any existing `%` character in the original JSON string.

It is OK to use a regular url-encoding function in a pinch.  The server will decode it all correctly and end up with the same JSON. However, the intermediate HTTP header will be much uglier and longer due to changing the frequently occuring JSON syntax characters into 3-byte `%hh` forms when it isn't necessary.

karlcz:
The initial properties for the client context would be what we had before in the query URL:
- cid
- pid

We may over time want to add more app-specific context parameters, e.g. for Chaise recordset to distinguish different kinds of query it is using in one app. We should deal with those as future extensions rather than trying to do everything at once.

karlcz:
The small change in ermrest, hatrac, and webauthn to include this header in its log output is already pushed to each of those respective repos.

So, once dev servers have updated, it should be possible to dev and test the client-side mechanism. You'll see the new header content logged into a field called `dcctx` in the existing JSON-formatted request log events we emit in `/var/log/messages` etc.

chiragsanghvi:
This task is divided into 3 phases:

- [x] Add support to http.js in ermrestjs to send headers as sent by user
- [ ] Change methods in ermrestjs that're exposed to user to start accepting header parameters.
- [ ] Make changes in Chaise to start sending logging headers to ermrestjs methods.

PR #556, opened 2017-11-17
Added client context header for logging in HTTP

chiragsanghvi:
This PR adds client context parameters like 'cid' etc to explicit HTTP header `Deriva-Client-Context` as mentioned in https://github.com/informatics-isi-edu/ermrestjs/issues/536. 

You can check the HTTP headers in the network tab of this URL

https://dev.isrd.isi.edu/~chirag/chaise/recordset/#1/isa:dataset@sort(release_date::desc::,id)

In addition you can check the messages log on the machine to be sure.

1. SSH to `dev.isrd` machine.
2. Switch user to root `su -`. 
3. Run `grep "dcctx" /var/log/messages`. You should see your last call entry in the log with name `dcctx`.

### DIFF
--- a/js/core.js
+++ b/js/core.js
@@ -48,7 +48,7 @@
      * @memberof ERMrest
      * @function
      * @param {string} uri URI of the ERMrest service.
-     * @param {Object} [params={cid:'null'}] An optional server query parameter
+     * @param {Object} [contextHeaderParams={cid:'null'}] An optional server header parameters for context logging
      * appended to the end of any request to the server.
      * @return {ERMrest.Server} Returns a server instance.
      * @throws {ERMrest.InvalidInputError} URI is missing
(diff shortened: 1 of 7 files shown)


js/core.js
```diff
@@ -57,21 +57,21 @@
...
      * `https://www.example.org/ermrest`.
      */
-    function getServer(uri, params) {
+    function getServer(uri, contextHeaderParams) {
 
         if (uri === undefined || uri === null)
             throw new module.InvalidInputError("URI undefined or null");
 
-        if (typeof params === 'undefined' || params === null) {
+        if (typeof contextHeaderParams === 'undefined' || contextHeaderParams === null) {
```
RFSH: This `if` statement seems unnecessary, we're setting `cid` to null in the `Server` constructor.

chiragsanghvi: I am not sure why we're setting it as null. This was there from before. If we don't want to send the value then I can simply remove it.

RFSH: I think we're setting it to null so we can just take a look at the log and see if we have any cid=null or not, just to see if there are some cases that we forgot to pass cid. My point was that this logic is duplicated. [The constructor itself is doing the same](https://github.com/informatics-isi-edu/ermrestjs/pull/556/files#diff-d85c7ecdc144ac4aec3ee7a0b24c3f9fR104).

It's not an important change. We can leave it the way it is since it's a legacy code.

PR #561, opened 2017-12-01
Add moment-timezone

jrchudy:
This PR adds moment timezone to the project for use in `Chaise`. This will properly convert `timestamptz` values to the timezone of the user's browser.


RFSH:
Can you please add a comment on top of the moment.min.js file that you created that says it includes those two libraries and states their version numbers?

Issue #299, opened 2016-12-03
number format pattern


hongsudt:
@karl: I need more than two digits of fraction in the floating point display. My project has small values which are being rounded and truncated too much and giving poor UX.

In the long run, the pre_format display option is meant to allow this to be customized. I know there are some concerns about which formatting language and library to use here.

@hongsudt commented on [Wed Nov 02 2016](https://github.com/informatics-isi-edu/ermrest/issues/115)

We need to start working on the template. I hear quite a few complaints about the GUI presenting an ID column in a number format (e.g. with , separating thousands). 

@karlcz @robes is there any js number formatting template that we can use? 

@hongsudt not sure what you mean by js number formatting template.

Sounds like if the concern is not formatting the number when it is a ID column, maybe we either need a heuristic for that, or we need a turn off formatting annotation?

@karlcz commented on [Fri Dec 02 2016](https://github.com/informatics-isi-edu/ermrest/issues/115#issuecomment-264608146)

I think this issue is mostly an ermrestjs issue?  We need to pick a posix-style formatting library so we can implement the pre_format display annotation. Then we can update the annotations doc with more precise dialect info for the format directives.

chiragsanghvi:
@hongsudt @karlcz Will this format work

The placeholders in the format string are marked by `%` and are followed by one or more of these elements, in this order:
- An optional `+` sign that forces to preceed the result with a plus or minus sign on numeric values. By default, only the - sign is used on negative numbers.
- An optional padding specifier that says what character to use for padding (if specified). Possible values are `0` or any other character precedeed by a `'` (single quote). The default is to pad with spaces.
- An optional `-` sign, that causes `printf` to left-align the result of this placeholder. The default is to right-align the result.
- An optional number, that says how many characters the result should have. If the value to be returned is shorter than this number, the result will be padded. When used with the `j` (JSON) type specifier, the padding length specifies the tab size used for indentation.
- An optional precision modifier, consisting of a `.` (dot) followed by a number, that says how many digits should be displayed for floating point numbers. When used with the `g` type specifier, it specifies the number of significant digits. When used on a string, it causes the result to be truncated.
- A type specifier that can be any of:
```
% — yields a literal % character
b — yields an integer as a binary number
c — yields an integer as the character with that ASCII value
d or i — yields an integer as a signed decimal number
e — yields a float using scientific notation
u — yields an integer as an unsigned decimal number
f — yields a float as is; see notes on precision above
[...]

PR #564, opened 2017-12-01
Added support for pre_format annotation

chiragsanghvi:
This PR adds support for pre_format annotation as mentioned in the issue https://github.com/informatics-isi-edu/ermrestjs/issues/299. 

The format for the printf function is as follows
```js

/**
      * printf function allows you to format strings using POSIX statndard
      *
      * %[flags][width][.precision]type
      *
      * flags field: The Flags field can be zero or more (in any order) of:
    
           - (minus) Left-align the output of this placeholder. (The default is to right-align the output.). 
                     This is used in conjunction with width field.
           + (plus)  Prepends a plus for positive signed-numeric types. positive = +, negative = -
                     (The default doesn't prepend anything in front of positive numbers.)
           0 (zero)  When the 'width' option is specified, prepends zeros for numeric types. (The default prepends spaces.)
                     For example, printf("%2X",3) produces  3, while printf("%02X",3) produces in 03.
           ' (quote) It separates numeric types by thousands using a comman(,) or other characters accroding to localization 
           #{char}   An optional padding specifier that says what character to use for padding (if specified). 
                     Possible values are any other character precedeed by a # (hash). print("%#_5d",10) produces ___10
    
      * Width field: The Width field specifies a minimum number of characters to output, and is typically used to pad fixed-width fields, 
                     where the fields would otherwise be smaller, although it does not cause truncation of oversized fields.
                     For example, printf("%*d", 5, 10) will result in "   10" being printed, with a total width of 5 characters.
                     Though not part of the width field, a leading zero is interpreted as the zero-padding flag mentioned above, 
                     and a negative value is treated as the positive value in conjunction with the left-alignment - flag also mentioned above.
                     When used with the j (JSON) type specifier, the padding length specifies the tab size used for indentation.
      
      * Precision field: The Precision field usually specifies a maximum limit on the output, depending on the particular formatting type.
                    For floating point numeric types, it specifies the number of digits to the right of the decimal point that the output should be rounded.
                    For the string type, it limits the number of characters that should be output, after which the string is truncated.
                    For example, printf("%.3s", 3.42134) will result in 3.421 being printed.
        
      * Type field: It can be any of:
                    % — yields a literal % character
                    b — yields an integer as a binary number
                    c — yields an integer as the character with that ASCII value
                    d or i — yields an integer as a signed decimal number
                    e — yields a float using scientific notation
                    u — yields an integer as an unsigned decimal number
                    f — yields a float as is; see notes on precision above
                    o — yields an integer as an octal number
                    s — yields a string as is
                    t — yields true or false
                    X — yields an integer as a hexadecimal number (upper-case)
                    j — yields a JavaScript object or array as a JSON encoded string
      
      * Named arguments:
                    Format strings may contain replacement fields rather than positional placeholders. Instead of referring to a certain argument, 
                    you can now refer to a certain key within an object. Replacement fields are surrounded by rounded parentheses - `(` and `)` - 
                    and begin with a keyword that refers to a key:

                        var user = { name: 'Dolly' }
                        sprintf('Hello %(name)s', user) // Hello Dolly

                    Keywords in replacement fields can be optionally followed by any number of keywords or indexes:

                        var users = [
                            {name: 'Dolly'},
                            {name: 'Molly'},
                            {name: 'Polly'},
                        ]
                        sprintf('Hello %(users[0].name)s, %(users[1].name)s and %(users[2].name)s', {users: users}) // Hello Dolly, Molly and Polly

                    Note: mixing positional and named placeholders is not (yet) supported
                    

       * @param{Object} options A javascript object that contains format property and optional bool_true_value
                                and bool_false_value. These will be used to replace true and false in `t` type fields.
       * @param{String|Number|Object} value This would be the value that needs to be formatted.

       * @returns {String}
       **/
```


js/core.js
```diff
@@ -1719,6 +1722,16 @@
                 return this._getNullValue(context);
             }
 
+            if (this.annotations.contains(module._annotations.COLUMN_DISPLAY)) {
```
RFSH: This is not necessary. Please remove it.

js/core.js
```diff
@@ -2039,6 +2047,8 @@
                 }
 
                 this._display[context] = {
+                    "isPreformat": (typeof annotation.pre_format === 'string'),
```
RFSH: This seems wrong. You should get the format before this line and in here check if that is string or not.
```javascript
var preformatConfig = typeof annotation.pre_format === 'object') ? ((typeof annotation.pre_format.format === 'string') ? annotation.pre_format : null ) : null;

this._display[context] = {
  "isPreformat": (typeof preformatConfig === 'string'),
  "preformatConfig": preformatConfig,
   //..
}
```

RFSH:
Please add some test cases that are testing the formatValue function and the fact that is actually using the annotations (Current implementation is not using the annotation because of the bug that I mentioned in the comments). You're only testing the internal function (`printf`). You can add them to the reference_values spec that you have created before, or any other specs that you think is more appropriate.

chiragsanghvi:
@RFSH I have added testcases for the annotation to be used in reference.

js/core.js
```diff
@@ -2046,9 +2044,16 @@
                     columnOrder = annotation.column_order;
                 }
 
+                if (typeof annotation.pre_format === 'object') {
+                    if (typeof annotation.pre_format.format !== 'string') {
```
RFSH: You don't need to throw an error, just ignore the annotation as if it didn't exist. You can just do a `console.log`.

chiragsanghvi: Alright I have removed it.

js/core.js
```diff
@@ -1723,8 +1726,16 @@
...
+            if (data === undefined || (data === null && this.type.name.indexOf('json') === -1)) {
                 return this._getNullValue(context);
+            } else if (data === null && this.type.name.indexOf('json') !== 0) {
+                return data;
+            }
+
+            var display = this.getDisplay(context);
+
+            if (display.isPreformat) {
+                return module._printf(display.preformatConfig, data);
```
RFSH: This function might throw an error which we don't want to raise to the users. Wrap this in a try-catch to ignore the pre_format annotation in that case.

PR #566, opened 2017-12-07
Temp handlebars

chiragsanghvi:
This PR adds support for [Handlebars](http://handlebarsjs.com/) as an alternated template engine to Mustache.

Default template engine is Mustache but you can always specify `handlebars` as a template engine in your annotation.

`template_engine` can be specified as a property in following annotations 

**[tag:isrd.isi.edu,2016:column-display](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2016-column-display)** : 
```js
{ 
  context: { 
     template_engine: 'handlebars',
     markdown_pattern: '...pattern...'
      ...
   }
}
```
**[tag:isrd.isi.edu,2016:table-display](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2016-table-display)** : 
```js
{ 
  context: { 
     template_engine: 'handlebars',
     row_markdown_pattern: '...pattern...'
      ...
   }
}
```
**[tag:isrd.isi.edu,2016:key-display](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2016-key-display)** : 
```js
{ 
  context: { 
     template_engine: 'handlebars',
     markdown_pattern: '...pattern...'
      ...
   }
}
```
**[tag:isrd.isi.edu,2016:foreign-key](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2016-foreign-key)** : 
```js
{ 
  context: { 
     template_engine: 'handlebars',
     domain_filter_pattern: '...pattern...'
      ...
   }
}
```


RFSH:
These are the first two things that came to my mind:

1. We need test cases to test actual value of columns. Currently although test cases are testing all the features of handlerbars that we want to use, but it's not testing them in the context of column value and only the actual function responsible for the rendering.

2. We need to add this to assets too. Which is just passing the templateEngine to the renderTemplate in the `_generateURL` function in hatrac.js

jrchudy:
@RFSH  I added an options object to the compile function for handlebars that has 2 parameters, `knownHelpers` and `knownHelpersOnly`. `knownHelpers` is an object with keys matching helper function names (builtin or registered with handlebars) and true/false as the value (which is how you turn off built in helpers). `knownHelpersOnly` is a boolean value that determines whether handlebars checks the defined helper(s) in the current template against the `knownHelpers` hash.

Documentation for these 2 properties is [here](http://handlebarsjs.com/reference.html) (explained poorly). All I could fine for a list of builtin helpers was [this](http://handlebarsjs.com/builtin_helpers.html).

The tests look like we are only testing a few of the multiple defined helpers in `handlebar_helpers.js`. I'm thinking we should add more tests for these added helpers to make sure they function as expected (even though they are pretty straightforward functions). I also think there should be a test against the negative case (one of the builtin helpers being disallowed).

RFSH:
@jrchudy The first thing that I pointed in my comment is still missing. We're testing the render functions directly but we're not testing the annotation and the integration of handlebars with the `tuple.values`, hatrac `url_pattern`, or other places that we're using template rendering.
We need test cases for all the different places that the `template_engine` setting has been added to.

jrchudy:
@RFSH  once again I think this is now finished.. 😄

RFSH:
I noticed you changed some specs that were not directly testing the template engines. For example the change that you made in faceting. It's testing that the returned row-names are correct or not, but it's not supposed to actually test the logic of row-name. Or the change that you made in columns_list. Having these test cases won't harm so we can just leave them since you went into trouble of adding them.

The only thing that I think we should change is what we're defining for the markdown_pattern value to test handlebars. To make sure it's actually using handlebars, we should modify those to be using a specific handlebars syntax.

test/specs/faceting/tests/01.faceting.js
```diff
@@ -1339,7 +1339,33 @@ exports.execute = function (options) {
                     });
                 });
 
-                it ("should return the list of rownames in entity-mode.", function (done) {
+                it ("should return the list of rownames (moustache templates) in entity-mode.", function (done) {
```
RFSH: As we discussed since this is not needed, if it's easy please remove the changes that you made here.

test/specs/reference/tests/05.reference_values.js
```diff
@@ -50,9 +52,159 @@ exports.execute = function (options) {
             return url;
         };
 
+        var moment = options.ermRest._currDate;
+        var expectedMomentValue = "<p>" + moment.day + " " + moment.date + "/" + moment.month + "/" + moment.year + "</p>\n";
+        var testObjects ={
```
RFSH: This whole spec has got to the point that I feel like we should just refactor it later. Just to help with readability of it, can you please add a note in those lines of `expectedValue` that their actual value will be attached with the `setLinkRID` function?

test/specs/reference/tests/05.reference_values.js
```diff
@@ -50,9 +52,159 @@ exports.execute = function (options) {
...
+                                '<p><img src="http://example.com/4006.png" alt="image"></p>\n',
+                                '<p><img src="/4006.png" alt="image with size" width="400" height="400"></p>\n',
+                                '<p><a href="" download="">download link</a></p>\n',
+                                '<figure class="embed-block" style=""><figcaption class="embed-caption" style="">&lt;script&gt;alert();&lt;/script&gt; caption</figcaption><iframe src="http://example.com/iframe" width="300" ></iframe></figure>',
+                                '<p><strong>This is some markdown</strong> with some <code>code</code> and a <a href="http://www.example.com">link</a></p>\n',
+                                '<p><strong>Name is :</strong> &lt;script&gt;alert();&lt;/script&gt;<br>\n<strong>This is some markdown</strong> with some <code>code</code> and a <a href="http://www.example.com">link</a></p>\n',
+                                '<code>GATCGATCGC GTATT</code>',
+                                '<p><a href="http://example.com/Sophomore">Sophomore</a></p>\n',
+                                '',
+                                '',
```
RFSH: This line for example. You can add a comment after it that suggests the actual value will be attached later.

Makefile
```diff
@@ -144,6 +145,16 @@ distclean: clean
 test: $(BUILD) ../ErmrestDataUtils
 	node test/jasmine-runner.js
 
+# Rule to run the unit tests
+.PHONY: testsingle
+testsingle: $(BUILD) ../ErmrestDataUtils
+	node test/single-test-runner.js
+
+# Rule to run the unit tests
+.PHONY: testsingledebug
```
RFSH: Please remove this. node-debug package has been deprecated.

RFSH: If you want you can change it to `node --inspect-break` but I don't think this is useful and I don't see the point of having a debug make target.

jrchudy:
Added other comments and removed make command.

Issue #562, opened 2017-12-01
Handle RID default function


hongsudt:
In RBK, all the RID columns have the following value as default:
`_ermrest.urlb32_encode(nextval('_ermrest.rid_seq'::regclass))`

I think this is true for all deployments that use RID. 

When we see `/_ermrest.*\(\.)*/`, maybe we can treat that as a function and not pass that to the default value.

jrchudy:
My thoughts were that if we know a column of domain type RID will always be generated, then we can ignore setting the "default value" for that column.

hongsudt:
Yes, that works.. RID will always be generated by the server.. So we can mark that as "auto-generated" and ignore default.

PR #570, opened 2017-12-14
System columns don't set column.default

jrchudy:
This PR addresses issue #562. Other system columns are addressed the same way (RCT, RMT, RCB, RMB).


RFSH:
I think you have referenced to the wrong issue, #562 is the related issue to this PR.

RFSH:
If the assumption is that ermrest is always returning UTC, you can just change the expected value for the `timestamptz_proper` to UTC and that test case should pass.

This is not really important and it's more of an observation. Since we're logging the entire error object, it is printing the whole stack trace of error too. And it's printing it in the test output too. Isn't it better to just log the error message? Saying `VALUE is not a valid DateTime default value.` should be sufficient to figure out the issue and stack trace seems unnecessary.

Issue #571, opened 2017-12-14
Error messages with table/schema level details


amitjha21:
We need to have table details for users to take corrective action when they see errors.
- [ ] Table name along with column details
- [ ] Handle invisible columns names

**Summary**
- We should change this to show the displayname of related reference.
- If we cannot find the displayname then should show the name that ermrest is returning
- use flag to make sure it is being passed from `delete` function
- In the case that we couldn't figure out the related reference's displayname we should append a message that says "If you have trouble removing dependencies please contact the site administrator." to the error message that we generate.

- Test case for pure and binary association. An example of table definition can be found here

-  Test case for the case that you couldn't find the table displayname given the constraint name and table name. In the mocked response use an arbitary constraint name.

- Test case for when the error is being generated from somewhere else. This might not happen in ermrest but we can mock the server response.

- Test the actual displayname of the related entities:

     - In case of pure and binary association, define a to_name on the foreignkey to the related entity. Displayname should be that one.
    - In case of related, define a from_name on the foreignkey.

RFSH:
This is only related to the `IntegrityConflictError`, which we are throwing when user is deleting a table that has some foreign key to:

 Currently if we can parse the message, we are showing the table name that ermrest is returning. We should change this to show the displayname of related reference. We have access to the table and also the constraint name of the foreign key. If given these two we can find the displayname we should change it, otherwise we should show the name that ermrest is returning.

Also we are assuming that this error will only going to happen in delete case, but we're not checking for it. We should change the `_responseToError` to accept a parameter that makes sure that we are doing this logic only in case of delete (You need to pass that argument in the error case in delete function in ermrestjs).

RFSH:
In the case that we couldn't figure out the related reference's displayname we should append a message that says "If you have trouble removing dependencies please contact the site administrator." to the error message that we generate.

amitjha21:
This has been merged. Assigning to @hongsudt  to verify and close it.

PR #574, opened 2017-12-16
Referenced table name should be displayed in case of Deletion [Integrity] Error

amitjha21:
This PR addresses #571.
In case of Delete operation if we can find the display name of the table then return it else show the name that `ermrest` is returning.
Can be tested  [here](dev.isrd.isi.edu/~amit/chaise/record/#1/isa:dataset/id=269).


js/reference.js
```diff
@@ -1754,13 +1754,14 @@
                  *
                  * github issue: #425
                  */
-
+                var self = this, delFlag = 'DEL';
```
RFSH: Instead of just defining this string here. Define an object the same way we have contexts defined in the utilities.js file. I'm not sure what is the best way to name that object. But eventually it should include `update` and `create`  too.

```javascript
module._operations = Object.freeze({
  DELETE: "delete",
  CREATE: "create",
  UPDATE: "update",
  READ: "read"
});
```

js/utilities.js
```diff
@@ -894,31 +894,49 @@
...
           }
 
+          // if(){
+            var fkConstraint = generatedErrMessage.match(/foreign key constraint \"(.*?)\"/)[1];    //get constraintName
+            if(fkConstraint != 'undefined' && fkConstraint != ''){
+              var relatedRef = ref.related(); //get all related references
+
+              for(i = 0; i < relatedRef.length; i++){
+                  key  = relatedRef[i];
+                  if(key.origFKR.constraint_names["0"][1] == fkConstraint && key.origFKR._table.name == refTable){
```
RFSH: change this from `"0"` to `0`

js/utilities.js
```diff
@@ -894,31 +894,49 @@
...
             detail = generatedErrMessage.substring(detail, generatedErrMessage.length);
             referenceTable = detail.match(/referenced from table \"(.*)\"(.*)/);
             if(referenceTable && referenceTable.length > 1){
-                referenceTable =  "the <code>"+ referenceTable[1] +"</code>";
+                refTable = referenceTable[1];
+                referenceTable =  refTable;
             }
           }
 
+          // if(){
```
RFSH: Remove this comment.

js/utilities.js
RFSH: Add a flag that you shows whether you found the referenceTable or not. If not. append the "If you have trouble removing dependencies please contact the site administrator." message.

**Please read all the comments in the issue before creating PR.**

test/specs/errors/conf/error_schema/error_schema.json
```diff
@@ -47,8 +47,114 @@
           "name": "Valid Table Name"
         }
       }
+    },
+    "main_delete_table": {
```
RFSH: Rename this. It's confusing, you're not testing delete feature here. Name it `main` or something.

test/specs/errors/conf/error_schema/error_schema.json
```diff
@@ -47,8 +47,114 @@
...
+          "default": null,
+          "nullok": false,
+          "type": {
+            "typename": "text"
+          },
+          "annotations": {}
+        }
+      ],
+      "annotations": {
+        "tag:misd.isi.edu,2015:display": {
```
RFSH: Remove this annotation. Keep your schema as minimal as you can for your test so it can be easily readable.

test/specs/errors/conf/error_schema/error_schema.json
```diff
@@ -47,8 +47,114 @@
...
+          "name": "Delete"
+        }
+      }
+    },
+    "association_table": {
```
RFSH: This is not an association table here. It's a simple related table.

RFSH:
Don't forget to assign the PRs that you create to yourself.

js/utilities.js
```diff
@@ -894,31 +894,49 @@
...
+            var fkConstraint = generatedErrMessage.match(/foreign key constraint \"(.*?)\"/)[1];    //get constraintName
+            if(fkConstraint != 'undefined' && fkConstraint != ''){
+              var relatedRef = ref.related(); //get all related references
+
+              for(i = 0; i < relatedRef.length; i++){
```
RFSH: Make sure that you have declared all the variables that you are using in the scope of the function. (`var i;`)

js/utilities.js
amitjha21: Added site Admin message by default and if the table name is found in `displayname`then it will be replaced by blank string in the main message.

test/specs/errors/tests/07.http_409_error.js
```diff
@@ -6,51 +6,143 @@ exports.execute = function (options) {
...
+
+        it("if it's an integrity error without deletion flag passed then we should allow the ermrest error to the client.", function (done) {
+           nock(url, ops)
+               .get("/ermrest/catalog/1234/schema")
+               .reply(409, integrityErrorServerResponse)
+               .persist();
+
+           server.catalogs.get("1234").then(null, function(err) {
+               expect(err.code).toBe(409, "invalid error code");
+               expect(err.message).toBe(integrityErrorWithoutDelMessage, "invalid error message");
```
RFSH: If this is going to the general (third) case, it should not have the prefix. But in your expected value the prefix is there.

amitjha21: This is the case when there is no DELETE flag.  In this case everything that is coming from the server would be passed.

Issue #568, opened 2017-12-12
remove RID from row name hueristics


hongsudt:
I heard a few complaints after we added RID in our heuristics.. 

We should leave that out and let the key heristics (for display purpose) deals with it..

RFSH:
We also decided to change the logic of rowname, to instead of falling back to using shortestkey, it should use displaykey.

PR #575, opened 2017-12-18
change rowname heuristics

RFSH:
1. removed RID from list of specific columns for rowname.

2. Change rowname to use displaykey instead of shortestkey (prefer text to other types).

[ermrestjs#568](https://github.com/informatics-isi-edu/ermrestjs/issues/568), [chaise#1403](https://github.com/informatics-isi-edu/chaise/issues/1403)


js/core.js
```diff
@@ -857,14 +857,54 @@
...
+        /**
+         * The columns that create the shortest key that can be used for display purposes.
+         *
+         * @type {ERMrest.Column[]}
+         */
+        get displayKey () {
+            if (this._displayKey === undefined) {
+                if (this.keys.length() !== 0) {
+                    var countTextColumns = function(key) {
+                        for (var i = 0, res = 0; i < key.colset.columns.length; i++) {
```
jrchudy: Didn't know you could instantiate multiple variables here. Good way to communicate the scope of a variable that isn't just the looping variable.

Issue #394, opened 2017-04-22
null support in faceting/searching


hongsudt:
when a value is null, the syntax for filtering/faceting is different from checking for an empty string (which is what we are currently doing).

hongsudt:
I heard some comments about the need to filter NOT NULL for some columns.

hongsudt:
The scalar facet supports null filtering, but not the entity facet..

PR #576, opened 2017-12-19
Not Null Support In Faceting APIs

RFSH:
This PR includes:

1. Adding not_null to the faceting syntax. The parser will accept having not_null and other filters, but in the FacetColumn if it has a not_null filter:
   - And also =null, then we will ignore all the filters.
   - otherwise we will just add the not_null filter.
2. Change faceting joins to left join, so we would produce null for entity pickers too.

issues: #394 

**I will create a PR in chaise that is using these features, we should merge this PR before the one in chaise.**


js/core.js
```diff
@@ -2933,7 +2934,9 @@
                 rightString += separator;
 
             }
-            return "(" + leftString + ")=(" + rightString + ")";
+
+            var joinType = ((typeof isLeft === "boolean") ? (isLeft ? "left": "right") : "");
```
jrchudy: If it is not a `left` join, it's a `right` join. But if the variable is not a boolean type, what type of join is it? Maybe add a comment about that here? I'm sure that's written in the `ermrest` spec though..

karlcz:
This seems kind of confused to me.

There are four types of join in ermrest:
- inner join (the default w/ just using foreign keys etc)
- left outer join (specified with the `left` keyword)
- right outer join (specified with the `right` keyword)
- full outer join (specified with the `full` keyword)

I don't know of any case in chaise/ermrestjs where we would be trying
to use right or full outer joins right now.

So, I would think the ermrestjs code should only be deciding between
a normal inner join or left outer join.

js/core.js
jrchudy: Based on Karl's comment, maybe we want to use a left join if specified and don't specify anything else for the other cases.

RFSH: I have documented it in the jsdoc `false: right join, true: left join, other values: inner join`. Even though right now only left join and inner join are being used. Although I'm going to change it to `true: left join, other values: inner join`

PR #579, opened 2017-12-22
Error super class ErmrestError created 

amitjha21:
A superclass of all ermrest error has been created in `errors.js`. The purpose of this class is to match errors at one place in `chaise`.
Test cases tweaked to have `instanceOf` check for `ErmrestError`.


js/errors.js
```diff
@@ -1,4 +1,5 @@
 
+    module.ErmrestError = ErmrestError;
```
RFSH: This is not really important. But the module that we are exposing is `ERMrset`, so just replace all these `ErmrestError` with `ERMrestError.`

hongsudt:
Did you mean ermrestjs error or ermrest error?

js/errors.js
```diff
@@ -17,19 +18,35 @@
...
      * @memberof ERMrest
      * @param {string} status the network error code
      * @param {string} message error message
      * @constructor
      */
     function TimedOutError(status, message) {
-        this.code = 0;
-        this.status = status;
-        this.message = message;
+        var code = 0;
```
RFSH: You don't need to create a variable here. You could just call it with the right code. Now that I am thinking about it, it might be better if you create a object of HTTPErrorCodes.

```
{
  BAD_REQUEST: 400,
  UNAUTHORIZED: 401,
  FORBIDDEN: 403,
 ...,
}
```
For these cases that the http code doesn't make sense just pass 0.

Although I'm not sure why for timeout error we're using 0, we can use it's appropriate code (408). We should also change the switch-case for the timeout to use it's own status code.

js/utilities.js
```diff
@@ -2251,3 +2251,18 @@
         UPDATE: "UPDT",   //update
         READ: "READ"        //read
       });
+
+    module._errorStatus = Object.freeze({
+      forbidden : "Forbidden",
+      itemNotFound : "Item Not Found",
+      facetingError: "Faceting Error",
```
RFSH: Please change this to "Invalid Facet Filters",  so it would be in the same format as others.

RFSH:
@hongsudt Even the object that we are exposing to chaise is just called `ERMrest` and not `ERMrestjs`. So we should stick with the same idioms. To chaise (or any other clients), it's just `ERMrset`.

js/errors.js
RFSH: Don't change the code for timeout yet, let's discuss about this first.

amitjha21:
Added api.md. Running all chaise cases.

Issue #578, opened 2017-12-19
fake a put request to log terminal error cases 


hongsudt:
We will be using the same header mechanism to log the detail..

hongsudt:
@karlcz did you say we should use a put request? Do you have a specific ermrest path to use? 
`/ermrest/error/terminal`?

RFSH:
We're faking a `put` request to `/ermrest/terminal_error` for now.

Issue #558, opened 2017-11-20
Refactor parser to convert filters to facets


jrchudy:
Change ermrestjs parser to convert filter into a facet. In the process we should remove all the filter related APIs.

This requires making changes to the `_processSingleFilterString` and `_processMultiFilterString` functions in `parser.js`.

RFSH:
A function has been added to filter that will turn it into facet. But we might not be able to use this facet, therefore this issue should still remain open. That function is now only being used for log purposes only.

RFSH:
I implemented this in #598. We don't want to always change filter to facet. Only in case of faceting APIs this change has uses. Therefore it has been added to the `Reference.facetColumns`. If facet equivalent of filter was understandable by current implementation of faceting, we will change the filters to facet.

PR #589, opened 2018-01-23
Add Extra Functionality to Pass Object to Context Header

RFSH:
This PR includes 
- (#558) Changing filter to facet for log purposes. It's only now for the log purposes because it  produces facets that the current faceting implementation might not understand.
- ([chaise#1389](https://github.com/informatics-isi-edu/chaise/issues/1389), #536) Adding functionality so chaise can pass extra objects to be logged by each request to ermrest.
- (#578) A method to log the errors in ermrest by creating a fake request which will result in 400 but will be logged anyways.

### DIFF
--- a/js/errors.js
+++ b/js/errors.js
@@ -282,3 +282,32 @@
 
     NoConnectionError.prototype = Object.create(Error.prototype);
     NoConnectionError.prototype.constructor = NoConnectionError;
+
+
+    /**
+     * Log the error object to the given ermrest location.
+     * It will generate a put request to the /terminal_error with the correct headers.
+     * ermrset will return a 400 page, but will log the message.
+     * @param  {object} err             the error object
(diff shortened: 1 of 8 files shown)


js/errors.js
```diff
@@ -282,3 +282,32 @@
...
+     * @param  {object} err             the error object
+     * @param  {string} ermrestLocation the ermrest location
+     */
+    module.logError = function (err, ermrestLocation) {
+        var defer = module._q.defer();
+        var http = module._wrap_http(module._http);
+
+        var headers = {};
+        headers[module._contextHeaderName] = {
+            e: 1,
```
jrchudy: What does `e: 1` mean? Is that the error code attached to the terminal error?

Will that always be the error code or do we want it to be the one returned for that request?

Issue #590, opened 2018-01-25
Change range parser


jrchudy:
After some discussion the range parser should be changed to use `::geq::` and `::leq::`. This way, the results shown in the UI is more consistent with what the user sees based on the histogram data.

This will become part of the changes to `ermrestJS` including with the other changes for histogram support (chaise [issue#1265](https://github.com/informatics-isi-edu/chaise/issues/1265)).

PR #593, opened 2018-01-27
Bucket ag ref

jrchudy:
This PR includes the reference API implementation for histogram support (chaise issue [#1265](https://github.com/informatics-isi-edu/chaise/issues/1265)). 

This allows for users to include histograms in their facets for columns with the following base types:
 - integer
 - float
 - date
 - timestamp

A new property (`barPlot`) was added to the facet definition inside of the `visible-columns` annotation under the `filter` context. This property, if set to false, the histogram will not show. 

It is expected to be an object with one property that `ermrestJS` and `chaise` understand, `nBins`. This property allows data modelers to control how many bins are used in the bin API which controls the number of visible bars in the graph.

This PR also addresses issue #590. To have consistent behavior between what data is displayed by the histograms and the recordset table, range queries are now inclusive with the min and max values instead of exclusive.

Note: In the tests, instead of verifying the exact contents of the response of `BucketAttributeGroupReference.read()`, I am instead verifying the integrity of the returned data. So in the for loop I'm testing 3 things. 
 1. Is the current `label.max` the next `label.min` 
    - this is to ensure the data I use to generate zoomed requests is working as expected and labels are the data used to calculate the `response.x` values
 2. Is the current min value (`response.x`) less than the next min value
 3. Is the difference between the next min value and the current one within a certain margin of error of the calculated width of each bucket. 
    - in some cases (floats, dates, timestamps) the labels get calculated as close to what the next one should be as possible. 
    - when calculating the difference between two dates or two timestamps, I noticed that from February to March and October to November, there is not a "whole" number of days.

This PR should be merged when the corresponding `chaise` PR is ready to be merged.


js/ag_reference.js
```diff
@@ -905,3 +921,181 @@ AttributeGroupReferenceAggregateFn.prototype = {
...
+                min: [],
+                max: []
+            };
+
+            var min, max;
+
+            for (var i=0; i<response.data.length; i++) {
+                var index = response.data[i].c1[0];
+                if (index !== null) {
+                    min = response.data[i].c1[1];
```
RFSH: You're assuming c1 and c2 are the name of columns here, but in the reference constructor your accepting any list of columns. Since this is completely written for this specific binning api in ermrest, you can change the Reference constructor variables to be the actual column name that you want binning for, then constructor the column object in the constructor.

js/ag_reference.js
```diff
@@ -905,3 +921,181 @@ AttributeGroupReferenceAggregateFn.prototype = {
...
+    verify(true == false, "Invalid function");
+};
+
+/**
+ * @return {ERMRest.BucketAttributeGroupPage}
+ */
+BucketAttributeGroupReference.prototype.read = function () {
+    var moment = module._moment;
+
+    // function decimals_s(num){
```
RFSH: Are these commented functions needed?

js/ag_reference.js
```diff
@@ -905,3 +921,181 @@ AttributeGroupReferenceAggregateFn.prototype = {
...
+
+                    data.x[index] = min;
+                    data.y[index] = response.data[i].c2;
+                }
+                // else if null (this is the null bin)
+                // we currently don't want to do anything with the null values
+            }
+
+            // This should be set to 12 to include the # of bins we want to display + the above max and below min bucket
+            for (var j=0; j<currRef._options.bucketCount+2; j++) {
```
RFSH: If bucketCount and the other two options are actually required and not optional just flat that list out. Instead of having them in a options object, list them as the parameters of the constructor.

js/reference.js
```diff
@@ -5401,6 +5401,36 @@
...
         },
 
+        /**
+         * Returns true if the plotly histogram graph should be shown in the UI
+         * If _facetObject.barPlot is not defined, the value is true. By default
+         * the histogram should be shown unless specified otherwise
+         *
+         * @type {Boolean}
+         */
+        get showHistogram() {
```
RFSH: I think it's better if we change the name of the api to be `barPlot` (same as the attribute name). Also the name of attributes in the annotation are usually snake case and not camel case. I think we should change barPlot to bar_plot and nBins to n_bins to be consistent.

js/ag_reference.js
```diff
@@ -905,3 +921,181 @@ AttributeGroupReferenceAggregateFn.prototype = {
...
+};
+
+BucketAttributeGroupReference.prototype.search = function (term) {
+    verify(true == false, "Invalid function");
+};
+
+/**
+ * @return {ERMRest.BucketAttributeGroupPage}
+ */
+BucketAttributeGroupReference.prototype.read = function () {
```
RFSH: Can you add a little explanation about the logic of read. The extra logic that you have added and also the expected format of output.

js/reference.js
```diff
@@ -6389,12 +6425,60 @@
         histogram: function (bucketCount, min, max) {
             verify(typeof bucketCount === "number", "Invalid bucket count type.");
             verify(min !== undefined && max !== undefined, "Minimum and maximum are required.");
+            var width, range, minMoment, maxMoment;
+
+            var absMax = max,
+                moment = module._moment;
+
+            if (this.column.type.rootName.indexOf("date") > -1) {
```
RFSH: Please add comments of why you are doing these changes, it makes it easier to follow the code.

js/utilities.js
```diff
@@ -1027,7 +1027,11 @@
             case 503:
                 return new module.ServiceUnavailableError(response.statusText, response.data);
             default:
-                return new Error(response.statusText, response.data);
+                if (response.statusText || response.data) {
```
RFSH: Was there an error response that didn't have statusText and data in it?

jrchudy: In some cases `_responseToError` is being called as part of the `.catch()` callback for promises. It's used correctly there when an http request errors out and returns. But if there is an error in code execution, the `.catch()` will catch that and then call `_responseToError` with no data in it (sometimes it's `response.message`).

js/ag_reference.js
jrchudy: They aren't required or used by `chaise` in any way. Those properties are required for calculating the labels to return to chaise though and padding out the x and y axes.

I figured it would be easier to add one object (maybe `_options` is the wrong name) that holds the values I need for calculations so the constructor only includes one value instead of 4 seemingly extra ones.

jrchudy:
@RFSH This PR is ready for review again.

RFSH:
Just to wrap this PR, can you please add the bar_plot to some of the existing facet, and make sure we're returning the correct `barPlot` and `histogramBucketCount`?

PR #597, opened 2018-02-02
Faceting errors

amitjha21:
This is related to faceting error issue in `chaise`. Its chaise counterpart will be created once this is approved. 
Detailed discussion on the issue is listed [here](https://github.com/informatics-isi-edu/chaise/issues/1332)
- Following error classes have been introduced:
   `InvalidFacetSorting` and `InvalidPageCriteria`.
- Changes were made to [parser.js](https://github.com/informatics-isi-edu/ermrestjs/blob/facetingErrors/js/parser.js), [utilities.js](https://github.com/informatics-isi-edu/ermrestjs/blob/facetingErrors/js/utilities.js) [reference.js](https://github.com/informatics-isi-edu/ermrestjs/blob/facetingErrors/js/reference.js) to throw these newly created errors and pass their path to generate new redirect links. 
- [error.js](https://github.com/informatics-isi-edu/ermrestjs/blob/facetingErrors/js/errors.js) has logic to trim invalid conditions and pass along the new link via error object.

Depending upon [which error](https://github.com/informatics-isi-edu/chaise/issues/1332#issuecomment-359974794) is being thrown the modal header changes and asks user to click on the OK button to reload the page without the faulty condition.
You can play around with the condition [here](https://dev.isrd.isi.edu/~amit/chaise/record/#1/isa:dataset/id=269@after(id)) to see the behavior in error modal.


js/errors.js
```diff
@@ -26,11 +28,15 @@
      * @param  {string} subMessage  technical details about the error. Appear in collapsible span in the modal box
      * @constructor
      */
-    function ERMrestError(code, status, message, subMessage) {
+    function ERMrestError(code, status, message, subMessage, redirectPath) {
```
RFSH: Comment what this `redirectPath` is. And I don't think you need the `errorData` object. Just flatten the attribute (`this.redirectPath = redirectPath`)

js/filters.js
```diff
@@ -137,7 +137,7 @@
      */
     function UnaryPredicate (column, operator) {
         if (!module.isValidOperator(operator)) {
-            throw new module.InvalidFilterOperatorError("'" + operator + "' is not a valid operator");
+            throw new module.InvalidFilterOperatorError("'" + operator + "' is not a valid operator", this._uri, '');
```
RFSH: This file is not being used by ermrsetjs at all. Just change it back. We should eventually remove this file.

js/parser.js
```diff
@@ -121,6 +121,7 @@
 
         // <sort>/<page>
         // sort and paging
+        var  pathWithCatalog = '#' + this._catalog + '/' + this._path;
```
RFSH: Remove this and just pass the path whenever you're using `pathWithCatalog`. Chaise should create the url.

js/parser.js
```diff
@@ -447,7 +448,7 @@
 
                 if (this.facets) {
                     facetFilter = _JSONToErmrestFilter(this.facets.decoded, mainTableAlias, mainTableName, this.catalog, module._constraintNames);
-                    if (!facetFilter) throw new module.InvalidFacetOperatorError();
+                    if (!facetFilter) throw new module.InvalidFacetOperatorError('', uri);
```
RFSH: Instead of `uri`, you should pass a path that chaise understands. This function is turning a chaise/ermrestjs url into ermrest. Some of the notions that are being used here is not understandable by chaise/ermrsetjs. You want to use a logic similar to the `get path()` function in here.

js/parser.js
```diff
@@ -435,7 +436,7 @@
 
                 if (this.projectionFacets) {
                     facetFilter = _JSONToErmrestFilter(this.projectionFacets.decoded, this.projectionTableAlias, this.projectionTableName, this.catalog, module._constraintNames);
-                    if (!facetFilter) throw new module.InvalidFacetOperatorError();
+                    if (!facetFilter) throw new module.InvalidFacetOperatorError('', uri);
```
RFSH: Take a look at the comment below.

js/errors.js
```diff
@@ -217,9 +227,21 @@
      * @constructor
      * @desc An invalid filter operator
      */
-    function InvalidFilterOperatorError(message) {
+    function InvalidFilterOperatorError(message, fullURI, invalidFilter) {
```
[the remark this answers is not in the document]
RFSH: Also You can just pass the correct path (path without filter) to this function, since you have access to that information when you're throwing this error and you don't have to do this check again.

js/parser.js
```diff
@@ -1337,9 +1342,10 @@
      * https://github.com/informatics-isi-edu/ermrestjs/issues/447
      *
      * @param       {String|Object} str Can be blob or json (object).
+     * @param       {String|Object} hash to generate rediretUrl in error module.
      * @constructor
      */
-    function ParsedFacets (str) {
+    function ParsedFacets (str, hash) {
```
RFSH: Change `hash` to `path`. It's miss-leading. Hash makes sense in chaise, but not here.

js/errors.js
```diff
@@ -203,9 +209,13 @@
      * @constructor
      * @desc An invalid facet operator
      */
-    function InvalidFacetOperatorError(message) {
+    function InvalidFacetOperatorError(message, hash) {
+
         message = message ? message : module._errorMessage.facetingError;
-        ERMrestError.call(this, '', module._errorStatus.facetingError, message);
+        currentUrlHash = hash;
+        facetFilter = currentUrlHash.slice(currentUrlHash.search('\\*::'), currentUrlHash.search('@'));
```
RFSH: Change this to search for `*::facets::`

js/errors.js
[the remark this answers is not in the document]
jrchudy: We should do all the processing of data for errors outside of the error constructor. The data should be provided to the constructor and attached to the error object there.

js/errors.js
```diff
@@ -273,7 +295,7 @@
      * @memberof ERMrest
      * @param {string} message error message
      * @constructor
-     * @desc A no internert was passed to the API.
+     * @desc A no internet was passed to the API.
```
jrchudy: This comment should be changed. This error can be for both no internet connection or no connection to the server. Sometimes there are intermittent errors when trying to communicate to the server.

js/errors.js
amitjha21: errorData is an object that is being used in [`Chaise`](https://github.com/informatics-isi-edu/chaise/blob/facetErrors/common/errors.js#L249:L261) to contain extra information regarding errors. I added here to align it with chaise and also make it more readable at chaise repo when extracting redirect link.

js/filters.js
amitjha21: Reverting all changes from `js/filters.js`.

amitjha21:
Incorporated all the comments for the review.

js/errors.js
```diff
@@ -283,7 +314,58 @@
     NoConnectionError.prototype = Object.create(Error.prototype);
     NoConnectionError.prototype.constructor = NoConnectionError;
 
+    function removePageCondition(path){
+      var newPath;
+      if (path != undefined && path.indexOf("@before") !== -1) {
+        beforerLiteral = path.match(/(@before\([^\)]*\))/)[1];
```
RFSH: You can use regex in replace, so you can just change these remove page and sort functions to just 

`return path.replace(REGEX, "")`

js/parser.js
```diff
@@ -121,6 +121,7 @@
 
         // <sort>/<page>
         // sort and paging
+        var  pathString = this._path;
```
RFSH: This variable seems unnecessary, just pass `this._path` in two places that are using this.

js/utilities.js
```diff
@@ -2463,7 +2463,9 @@
       invalidInput : "Invalid Input",
       invalidURI : "Invalid URI",
       noDataChanged : "No Data Changed",
-      noConnectionError : "No Connection Error"
+      noConnectionError : "No Connection Error",
+      InvalidSortCriteria : "Invalid Facet Sorting",
```
RFSH: Change this to be the same as the other error "Invalid Sort Criteria"

RFSH:
More test cases are needed for this PR. You need to test the actual properties of the error object to make sure your logic of modifying the path is working as expected.

test/specs/parser/tests/01.parser.js
```diff
@@ -320,6 +322,22 @@ exports.execute = function(options) {
...
 
+        describe("when uri have invalid paging Criteria", function() {
+            it("it should throw an error.", function() {
+                expect(function () {
+                    options.ermRest.parse(baseUri + "@after()");
+                }).toThrow(invalidPageCondition);
+            });
+        });
+
+        describe("when uri have invalid paging Criteria", function() {
```
jrchudy: This describe doesn't describe the test inside it. This should be something like "when uri has an invalid filter".

amitjha21:
Added `try..catch()` to return once parse() fails. This helps avoid going diretly to `exceptionHandler`.
```
module._onload().then(function () {
              try {
                location = module.parse(uri);
              } catch (error){
                return defer.reject(error);
              }
...
```

js/errors.js
```diff
@@ -195,31 +202,55 @@
...
     InvalidFacetOperatorError.prototype.constructor = InvalidFacetOperatorError;
 
+    // path consits of facet filter alongwith table and schemaName
+    // invalidFilter is removed from the path
+    function removeInvalidFilter(path, invalidFilter){
+      var newPath;
+      if (invalidFilter != ''){
+        newPath = path.slice(path, path.search(invalidFilter));
+      } else{
+        path = path.slice(path);
```
RFSH: Can you explain what are these three lines doing? I cannot exactly understand it.

test/specs/parser/tests/01.parser.js
```diff
@@ -441,6 +441,22 @@ exports.execute = function(options) {
         var baseUri = options.url + "/catalog/" + catalogId + "/entity/" + schemaName + ":" + tableName;
         var facetError = "Given encoded string for facets is not valid.";
         var location, uri;
+        var invalidPageConditionErrorObj = {
```
RFSH: Make these changes to your test cases:

1. Add a test case for Invalid Sort criteria. Sort based on a column that is not sortable. You need to define [column-display](https://github.com/informatics-isi-edu/ermrest/blob/master/user-doc/annotation.md#2016-column-display) annotation for that column with `column_order: false`

2. For invalid page, change the url to have filter. To make sure that you're only removing the page.

3. For invalid filter and facet, add two test cases for each. One with sort/page and one without any modifiers.

test/specs/parser/tests/01.parser.js
```diff
@@ -455,6 +471,54 @@ exports.execute = function(options) {
                     options.ermRest.parse(baseUri + "/*::facets::invalidblob");
                 }).toThrow(facetError);
             });
+
+            it("should throw an error for invalid facet filter errors.", function() {
+                try{
+                      options.ermRest.parse(baseUri + "/*::facets::invalidblob@sort()");
+                } catch(e){
+                  expect(e).toEqual(jasmine.any(Object));
```
RFSH: These `expect(e).toEqual(jasmine.any(Object));`s are not necessary.

RFSH: Also if this doesn't throw any errors, the `it` will succeed.  Add a dummy `expect(false).toBe(true, "didn't throw any errors".)` after the parse inside the try-catch.

js/errors.js
amitjha21: It was meant for cases where a filter is not supplied to error module and the path needs to be sanitized. Earlier `path` used to have `/schema:table/` format with two `/`. Since that has been changed to `schema:table/`, I just need to get everything before filter starts(before `/`). 

e.g. In this case  `isa:dataset/id::eq::269&;`, I should get `isa:dataset` as `Location()` throws only the `path`.
Made this change.

js/reference.js
```diff
@@ -1162,7 +1164,8 @@
 
                         // column is not sortable
                         if (typeof sortCols === 'undefined') {
-                            throw new module.BadRequestError("", "Column " + sortObject[i].column + " is not sortable.");
+                            var path = parts = this.uri.match(/(.*)\/catalog\/([^\/]*)\/(entity|attribute|aggregate|attributegroup)\/(.*)/)[4];
```
RFSH: `this.location` returns the location object here, and then you have access to path (`this.location.path`).

test/specs/errors/tests/09.modifier_error.js
```diff
@@ -0,0 +1,164 @@
...
+    };
+
+      beforeAll(function (done) {
+        options.ermRest.resolve(baseUri + "@sort(id)", {cid:"test"}).then(function(response) {
+          references = response;
+          done();
+        });
+      });
+
+        describe("when uri doesn't have any facets, ", function() {
```
RFSH: This is unnecessary here. Just remove it.

test/specs/errors/tests/09.modifier_error.js
```diff
@@ -0,0 +1,164 @@
...
+            it("Location.facets should be undefined.", function() {
+                location = options.ermRest.parse(baseUri);
+                expect(location.facets).toBeUndefined();
+            });
+        });
+
+        describe("when uri have invalid facet blob", function() {
+            it("parser should throw an error.", function() {
```
RFSH: This is not needed as you are testing the same thing in depth few lines further.

Issue #607, opened 2018-03-07
Sorting Based on Invalid Column


RFSH:
If you sort based on a column that doesn't exist in the table, the read function is throwing 404 error while it should throw `InvalidSortCriteria`.

Issue #592, opened 2018-01-26
App error if facet source is a list w/ just a column


karlcz:
The syntax for the facet `source` field is based on that used in ermrest policies. In ermrest, the source `"RID"` is a short-hand for the generalized source list `["RID"]` i.e. a projection with no joins. In the facet list, using the general form with an array causes an app error instead of working the same way as a bare column name.

RFSH:
I didn't know this was the behavior in ermrest policies. I assumed if it's an array, it will be a path. I need to update the Facet wiki and also the implementation to behave the same way.

PR #610, opened 2018-03-13
Pseudo column

RFSH:
This PR will add `PseudoColumn` support to ermrestjs. We should be able to use the same syntax that we have for facets in all of the `visible-foreignkeys` and `visible-columns`  contexts. Based on the [chaise#1262](https://github.com/informatics-isi-edu/chaise/issues/1262) issue, this is the implementation of first two tasks in ermrestjs side. For the second bullet, I need to make a small set of changes in chaise.

This a summary of changes that I made:

1. Changed the structure of code since it was getting messier. I created an extra file for anything related to column implementation. Also modified some of the test spec structures for the same reason.

2. Add support for facet syntax in visible-columns list. This also includes changing the previous logic of generating name for specific pseudo-columns that we had (Key, ForeignKey, InboundForeignKey). With this change, the generated name is a md5 of the path. To make sure we are backward compatible, I created the `getColumnByName` so that it look for previous syntax of generating names too.

3. Add support for facet syntax in visible-foreignkeys. `PseudoColumn` objects have an `InboundForeignKey` flag that if its true, its reference will behave the same way as Inbound Related References (has all the extra attributes).

4. Modify sort, page, and tuple to make sure everything is working with the new syntax for columns.

Related Issues: #607 #592 [chaise#1262](https://github.com/informatics-isi-edu/chaise/issues/1262)

Notes: 
1. The only test case in chaise that needs modifying is the prefill test which is manually creating a cookie. 
2. I need to update the annotation.
3. [This](https://github.com/informatics-isi-edu/ermrestjs/wiki/Pseudo-Column-Logic-&-Heuristics) is the logic wiki that I wrote. I'm not sure if this has enough information or not.


jrchudy:
I don't see any issues with this PR. 

A number of private functions were added to parser that include `Facet` in the names. Are these functions specific to faceting or general functions associated with a source reference/table?

Issue #604, opened 2018-03-05
Preformat support for times, dates, timestamps,...


robes:
I.e., I'd like to be able to preformat a timestamp to display it like a date (e.g., YYYY-MM-DD only).

Why? In a compact view showing a fully timestamptz is a bit verbose and honestly not that necessary however, the time parts still seem useful to track internally if not also to display in the record page (detail view) where there is more room.

jrchudy:
When a "preformat" annotation exists for columns of type `Date`, `Timestamp`, or `Timestamptz`, the moment library will be used to then format the current value based on the value provided by the "preformat" annotation. If this creates an invalid moment object (the format string was invalid), format the value as if no "preformat" was defined.

Test cases:
Test Date/Timestamp format working as expected.
Test invalid Date/Timestamp format returns raw value per our defined print functions when "preformat" doesn't exist

jrchudy:
Changes to the functionality that @RFSH and I discussed:
 - Removed the error case
   - whatever the format is, it will be displayed that way.
   - moment doesn't have an exposed function to verify format

jrchudy:
The corresponding PR was merged into master, reassigning this issue to @robes.

PR #613, opened 2018-03-14
Preformat datetime

jrchudy:
This issue resolves issue #604.


jrchudy:
Documentation for valid tokens in the string provided to the "preformat" annotation in the case of "Dates" and "Timestamp[tz]s" can be found in the [Pre Format Annotation document](https://github.com/informatics-isi-edu/ermrestjs/wiki/Pre-Format-Annotation).

Issue #618, opened 2018-03-29
Inconsistent behavior when column_order has multiple columns


RFSH:
@karlcz commented on [Thu Mar 29 2018](https://github.com/informatics-isi-edu/chaise/issues/1475)

@BDemps commented on [Tue Mar 27 2018](https://github.com/informatics-isi-edu/synspy/issues/61)

When I click the arrow at the bottom of the page on say the "zebrafish subjects" page, or when I click "show details" on one of the searchable fields on the left hand side of the page, the system crashes, giving a terminal error, which I've also taken a screenshot of.
<img width="1008" alt="screen shot 2018-03-27 at 4 26 18 pm" src="https://user-images.githubusercontent.com/28578487/38000330-dbddf482-31db-11e8-9041-6ed826081f31.png">

---

@karlcz commented on [Thu Mar 29 2018](https://github.com/informatics-isi-edu/chaise/issues/1475#issuecomment-377355195)

When I tried this on the prod server with firefox, I get a different error for the next page button:

Bad Request

400 Bad Request The request is malformed. The "after" page key of length 3 does not match the "sort" key of length 4.

---

@karlcz commented on [Thu Mar 29 2018](https://github.com/informatics-isi-edu/chaise/issues/1475#issuecomment-377356549)

I can reproduce this on synapse-dev too with the same result.  I had to make sure my annotations were in the same state, so I think it has to do with a customized sort order.

You can go to https://synapse-dev.isrd.isi.edu/chaise/recordset/#1/Zebrafish:Subject  and then try to use the next-page button at the bottom of the page.

RFSH:
This has been fixed.

PR #620, opened 2018-04-03
fix issues related to paging

RFSH:
In this PR I fixed the issue reported in #618, which was related to having multiple columns in `column_order`. While fixing that, I made some other changes to the APIs related to sort and paging.

The set of changes are:

- Make sure sort and page length are the same right before creating
the http request in read. Previously we were doing this check in the parser, but that is not necessarily true since the modifiers that are defined in the parser (location object) are not necessarily the modifiers that we are using to make ermrest request.

- Make sure Page.next and Page.previous are working properly in cases
that the column_order length is more than one. We were not using the proper `sortColumns` and were just creating the list of `after` and `before` based on the column in non-pseudo-column mode.

- Avoid adding duplicate columns in sort and page. To do this, I had to guard against duplicate columns in the `next`, `previous`, and `read` functions while looking at the sort columns. We might be able to have a central place for processing the sort modifier, which can be done as part of factoring the read function later.

- We were always appending the shortestkey to the sort columns, but that is not needed as long as all columns of a key are in the sort columns (it doesn't have to be the shortestkey, it can be any keys). I've fixed this in the system-columns branch and now just moved it to this branch too.


js/core.js
```diff
@@ -2486,7 +2491,19 @@
                     for (var i = 0 ; i < annotation.column_order.length; i++) {
                         try {
                             // column-order is just a list of column names
-                            columnOrder.push(this.table.columns.get(annotation.column_order[i]));
+                            var col = this.table.columns.get(annotation.column_order[i]);
+
+                            // json and jsonb are not sortable.
+                            if (["json", "jsonb"].indexOf(col.type.name) !== -1) {
```
jrchudy: Should we have an enumerated list of "non-sortable" column base types?

js/core.js
```diff
@@ -3049,7 +3066,19 @@
                     for (var i = 0 ; i < annotation.column_order.length; i++) {
                         try {
                             // column-order is just a list of column names
-                            columnOrder.push(this.key.table.columns.get(annotation.column_order[i]));
+                            var col = this.key.table.columns.get(annotation.column_order[i]);
+
+                            // json and jsonb are not sortable.
+                            if (["json", "jsonb"].indexOf(col.type.name) !== -1) {
```
jrchudy: same comment as above

Issue #625, opened 2018-04-13
Inconsistent visible columns processing


karlcz:
We need to use a single, common visible column parsing routine that can handle all the syntactic variants and convert them (or reject if invalid) for the given application context.  A DBA should be able to set a single `*` context visible column list that functions correctly for all applications, if they don't want to bother customizing each context:

-  `"Column"`, `{"source": "Column"}`, and `{"source": ["Column"]}` are semantically equivalent and should be supported everywhere that regular columns are supported.  There should be no difference in ermrest query formulation regardless of which notation is used.
- `["S", "FK"]` and `{"source": [{"outbound": ["S", "FK"]}, "RID"]}` are semantically equivalent and should be supported everywhere that a single outbound foreign key is supported. There should be no difference in ermrest query formulation regardless of which notation is used.
- The generalized pseudo column syntax should be acceptable everywhere. If the path is too complex for the individual context, that pseudo column should be discarded. It should never cause an application failure.  (Right now, the "entry" contexts violate this rule.)
- The `"markdown_name"` and `"comment"` features of the pseudo column notation should be accepted and used for point-of-use customization in every application context, overriding any other column or foreign key annotation preference, even for the simple column and foreign key cases.

RFSH:
The only contexts that are treated differently are `entry` (`entry/edit`, `entry/create`), and `filter`.

The reason that I'm not allowing the pseudo-columns in `entry` context is because I wasn't sure to what extend we actually want to support this. Only pseudo-columns that have a source without any foreignkey or just one outbound foreignkey? Anything other than that should be ignored? Or do we actually want to support the all-outbound pseudo-columns too?

And for `filter`, currently the way it works is that if you haven't defined the specific context, we're creating facets based on visible-columns in `compact` and related entities in `detailed`. Which we can change obviously. But I think the reason behind having specific `filter` context was to make sure that those are specifically defined for facet. There are some extra attributes that facet objects can have that are not applicable to columns (`choices`, `ranges`, etc.).

karlcz:
I'm not suggesting any changes to the core functionality as far as
which pseudo-column types work in each app context. I'm only
requesting that we treat the different annotation variants as
equivalent where they have the same semantic meaning:

1. a local column
   - `"Column Name"`
   - `{"source": "Column Name"}`
   - `{"source": ["Column Name"]}`
2. a single outbound foreign key
   - `["Schema", "Constraint"]`
[...]

Issue #624, opened 2018-04-12
Visible Columns Changes Regarding Aggregate


RFSH:
If pseudo column has inbound foreign key and aggregate is not defined on it, it should be ignored unless it's in entity mode and context is `detailed`.

Issue #623, opened 2018-04-12
Wrong tooltip on entity-mode pseudo columns


karlcz:
When a pseudo column is in entity mode, the tooltip should come from the table's comment rather than the key column's comment.

I also suspect we should add support for a "comment" field in the pseudo column (sibling to "markdown_name") so that a DBA can customize the tooltip to better express the meaning of the whole source path.

RFSH:
Based on the discussion that we had the following should be the logic for both displayname and comment of pseudo columns (first rule applicable).


### Displayname
1. Use the defined `markdown_name`.
2. If aggregate is defined, use the following:
    - min: `Min {COL_NAME}`
    - max: `Max {COL_NAME}`
    - cnt/cnt_d: `# {COL_NAME}`
3. If there's no path (column from same table)
   - If `entity` mode: It's the same as `KeyPseudoColumn`, and therefore it should have the same logic.
  - Otherwise return column's displayname.
4. If it's inbound foreign key (pure and binary, or just a single inbound fk), use the same logic as `InboundForeignKeyPseudoColumn`.
5. Otherwise return the end table's displayname.

### Comment
It should align with the displayname, therefore:
1. Use the defined `comment`.
2. If aggregate is defined, use the following:
    - min: `Minimum of {COL_NAME}`
    - max: `Maximum of {COL_NAME}`
    - cnt: `Number of {COL_NAME}`
    - cnt_d: `Number of distinct {COL_NAME}`
3. If there's no path (column from same table)
   - If `entity` mode: It's the same as `KeyPseudoColumn`, and therefore it should have the same logic.
  - Otherwise return column's comment.
4. If it's inbound foreign key (pure and binary, or just a single inbound fk), use the same logic as `InboundForeignKeyPseudoColumn` (comment of fk, or comment of association table).
5. Otherwise return the end table's comment.

RFSH:
This has been done as part of #626. Both displayname and comment have been changed based on the logic described on the previous comment.

PR #626, opened 2018-04-18
Pseudo column changes

RFSH:
This PR includes:

- Modifying the `Reference.columns` logic in case of pseudo-columns. Previously as long as the definition of column is in the form of source-object I was creating the general `PseudoColumn` object. And that prototype would act the same way as `KeyPseudoColumn`, `ForeignKeyPseudoColumn`, or `InboundForeignKeyPseudoColumn` for special cases. Since we wanted to expand this way of defining columns to other context (mainly `entry`), we couldn't continue doing that since in the `entry` context each of these prototypes have their own special APIs. Therefore, I created the `_createPseudoColumn` wrapper object that will make sure that we are creating the correct object type (#625).

- Modify the visible columns logic: If pseudo column has inbound foreign key and aggregate is not defined on it, it should be ignored unless it's in entity mode and context is detailed (#624).

- Modify the displayname and comment logic (#623).

### DIFF
--- a/js/column.js
+++ b/js/column.js
@@ -3,6 +3,89 @@
  */
 
 /**
+ * Will create the appropriate ReferenceColumn object based on the given sourceObject.
+ * It will return the follwing objects:
+ * - If aggregate: PseudoColumn
+ * - If no path, scalar, asset annot:AssetPseudoColumn
+ * - If no path, scalar (or entity in entry context): ReferenceColumn
+ * - If no path, entity, non-entry context: KeyPseudoColumn
+ * - If path, entity, outbound length 1: ForeignKeyPseudoColumn
+ * - If path, entity, inbound length 1: InboundForeignKeyPseudoColumn
+ * - If path, entity, p&b association: InboundForeignKeyPseudoColumn
+ * - Otherwise: PseudoColumn
+ *
+ * @private
+ * @param  {ERMrest.Reference}  reference    [description]
+ * @param  {ERMrest.Column}  column       [description]
+ * @param  {Object}  sourceObject the column definition
+ * @param  {String}  name         the name to avoid computing it again.
+ * @param  {ERMrest.Tuple}  mainTuple    the main tuple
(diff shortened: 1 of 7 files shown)


js/utilities.js
```diff
@@ -1105,7 +1105,7 @@
 
     /**
      * returns the displayname that should be used for a facetObject
-     *
+     * TODO update this!
```
jrchudy: Is this to remind you to update before merging the PR or later?

js/column.js
```diff
@@ -324,8 +421,13 @@ ReferenceColumn.prototype = {
 };
 
 /**
- * TODO aggregate is currently ignored since it's not being used.
- * We should add suppoort for aggregate to pseudo-columns later
+ * If you want to create an object of this type, use the `module._createPseudoColumn` method.
```
jrchudy: I know there isn't a way to make anything `private`, but we could preface `private` constructors (maybe using `_` still) so that it is more clear looking at the code.

RFSH: All these prototypes are internal and are not available on the `ERMrest` module. It's true about almost all of the prototypes that we have in ermestjs (`Reference`, `Page`, `Tuple`, `Column`, etc.). So the notaion of making it `private` wouldn't do much in this case since it's always being used within ermrestjs. This is more of a note that if in other parts of ermrsetjs we wanted use this prototype, we should call that function instead. I was looking at other ways of changing a constructor of a prototype to produce another type of prototype, but this method seemed more feasible.

js/utilities.js
RFSH: I completely forgot about this. Thanks for pointing it out. This is the function that was used in both `PseudoColumn` and `FacetColumn`, but since I changed the logic in `PseudoColumn`, it's now only being used in `FacetColumn`. Ultimately both of these should have the same logic, but I didn't change the `FacetColumn` logic because I saw that some deployment are actually relying on this logic. I'll add more info for the `TODO`.

PR #630, opened 2018-05-04
remove encoding of the catalog id

jrchudy:
It is returned from `ermrest` as encoded so it should still be decoded. But there's no need to re-encode the catalog id before using it in the urls.


RFSH:
The [getAssociationRef](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/reference.js#L4264) is the only other place that I found that we're doing catalog encoding and is missing in this PR.

This actually shows that we (mainly me), should have used just one central place for creating the URLs. It would have made it a lot easier to do this task.

PR #641, opened 2018-05-21
fix facetFilter merge logic

RFSH:
The logic wasn't taking care of having the same source in the filter.


jrchudy:
That makes sense. So before it would iterate and see that a new source was added to the list and run the loop again (causing this buggy behavior) when in reality it should have terminated?

Issue #643, opened 2018-05-22
Allow Array Aggregate For Columns


RFSH:
Currently we're only allowing `cnt`, `cnt_d`, `min`, and `max`. We should allow `array` too. It should return the values joined by ` ,` (`a, b, c`).

Default name: `{COLUMN_NAME}`
Default comment: `List of {COLUMN_NAME}`

RFSH:
This has been merged with master. I updated the annotation document regarding the usage of `array` aggregate

PR #644, opened 2018-05-22
Array agg

RFSH:
Allow `array` aggregate on pseudo-columns. One of the test cases will fail because the array aggregate function in ermrest is not deterministic and is not sorting the values. I wasn't sure if I should sort the values, or modify the test case to just look for the values and not the actual order (We should discuss this during the meeting).

#643


RFSH:
We decided to rely on the javascript sort function for now and simply sort the raw values that we're getting from ermrset.

PR #663, opened 2018-06-16
Empty hatrac files

jrchudy:
Another change that was discussed. I do not think this change should be made as it causes some further issues down the line that need to be fixed in chaise. Not doing this means that the error would be returned from the file exists check and never create any extra upload jobs (in the case where 2 files are trying to be uploaded but one namespace is 403 and the other doesn't exist yet).

This means cleanup would not need to be done and therefore the version info that chaise holds won't need to be fixed. But maybe that is a bug that should actually be fixed. See PR [#1556](https://github.com/informatics-isi-edu/chaise/pull/1556) in `chaise` for more info.


jrchudy:
After some discussion with @karlcz I realized I had configured my ACLs in hatrac improperly. I had `subtree-create` defined for all the sub directories (namespaces) but never changed the permissions on the individual job that was created by another user. 

Adding the `update` ACL to the job allowed me to reproduce the 403 response for the HEAD request again and then to properly create a new job (same name  but with a different version).

Issue #629, opened 2018-05-02
Basic Array Support


RFSH:
These are the minimum set of changes that we need in ermrestjs to make sure we're supporting arrays:

- [x] Ignore the facet definitions based that are based on array columns. You should add extra case to [this](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/reference.js#L384,L394) and [this](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/reference.js#L396,L421) function. If column is an array, you should return false.

- [x] Make sure [`formatvalue`](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/core.js#L1754) supports array columns. If it's array instead just calling `module._printf` or `_formatValueByType` on the data, you should loop through data and apply them to each value in the array. This should return the values separated by comma (`FORMATED_VAL1, FORMATED_VAL2`).

PR #664, opened 2018-06-19
Array support

RFSH:
With this change, we can support presentation of array column types. This only includes primitive types (number, date types, and text). It does not include `json`, or `markdown` or any other custom domain types.

The following are a summary of changes in this PR:

- Made sure facet is ignoring column array types (we're not supporting faceting for arrays).

- Changed the `formatvalue` implementation to support arrays. This function in case of arrays, will return an array of formatted values. I had to do this, because we wanted to change the value of `null` to `*No Value*` and `""` (empty string) to `*Empty*`. `formatvalue` has no notion of markdown. It's the caller's responsibility to go over the returned values and process markdown appropriately.

- Changed `formatPresentation` function to accept the raw values of table instead of formatted value of the column.

*Currently I am adding more comments to the implementation of `values` to explain how exactly these formatPresentation and formatvalue are used together. I think having them in code is better than here*

Issue: #629


js/utilities.js
```diff
@@ -2220,6 +2255,14 @@
         }, text);
     };
 
+    module._escapeHTMLCharacters = function (text) {
```
jrchudy: We have this function in chaise:
```
function makeSafeIdAttr(string) {
            return String(string)
                .replace(/&/g, '&amp;')
                .replace(/\s/g, '&nbsp;') // any whitespace
                .replace(/</g, '&lt;')
                .replace(/>/g, '&gt;')
                .replace(/"/g, '&quot;')
                .replace(/'/g, '&#39;');
        }
```

Maybe this one you wrote should consider whitespace characters too.

PR #668, opened 2018-06-22
Delete tuples

jrchudy:
This PR adds tuples to the parameters of the delete request. This should be merged along with the corresponding `chaise` PR.

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -1425,7 +1425,7 @@
 
         /**
          * Updates a set of resources.
-         * @param {Array} tuples array of tuple objects so that the new data nd old data can be used to determine key changes.
+         * @param {Array} tuples array of tuple objects so that the new data and old data can be used to determine key changes.
          * tuple.data has the new data
          * tuple._oldData has the data before changes were made
          * @param {Object} contextHeaderParams the object that we want to log.
@@ -1807,17 +1807,18 @@
         },
 
         /**
-         * Deletes the referenced resources.
+         * Deletes the referenced resources. Should be called on an unfiltered reference
+         * tuples passed to create the exact set to be deleted
          * NOTE This will ignore the provided sort and paging on the reference, make
          * sure you are calling this on specific set or rows (filtered).
          *
-         * @param {Object} contextHeaderParams the object that we want to log.
+         * @param {Array} tuples optional array of tuple objects used to construct the uri, if not-defined use the reference uri (should only be empty for association reference)
+         * @param {Object} contextHeaderParams optional object that we want to log.
          * @returns {Promise} A promise resolved with empty object or rejected with any of these errors:
          * - ERMrestjs corresponding http errors, if ERMrest returns http error.
          */
-        delete: function(contextHeaderParams) {
+        delete: function(tuples, contextHeaderParams) {
             try {
-
                 var defer = module._q.defer();
 
                 /**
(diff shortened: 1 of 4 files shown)


js/reference.js
```diff
@@ -1807,17 +1807,17 @@
         },
 
         /**
-         * Deletes the referenced resources.
+         * Deletes the referenced resource's tuples.
          * NOTE This will ignore the provided sort and paging on the reference, make
          * sure you are calling this on specific set or rows (filtered).
          *
-         * @param {Object} contextHeaderParams the object that we want to log.
+         * @param {Array} tuples optional array of tuple objects used to construct the uri, if not-defined use the reference ermrestCompactUri (should only be empty for association reference)
```
RFSH: Change this to `@param{ERMrest.Tuple[]}` if you actually want tuple objects.

test/specs/reference/tests/11.delete.js
```diff
@@ -4,14 +4,14 @@ exports.execute = function (options) {
         var catalogId = process.env.DEFAULT_CATALOG,
             schemaName = "delete_schema",
             tableName = "delete_table",
+            tuples = [{data: {key_col: 1}}],
```
RFSH: It's better if you actually read and pass `page.tuples` to the function (instead of hacking the tuple object). 

We also need two more test cases. One for the case of passing multiple tuples and another for calling delete without any parameters.
We could also test this with a shortestkey of length >1, but I don't think that's needed (soon all deployments will enforce RID).

jrchudy:
Made some changes that @RFSH and I discussed that made using tuples much more difficult in the case of association references. 

@RFSH  There is one test case still failing that I updated in the error specs. I'm not exactly sure what I need to change to fix it besides adding the array of data to the `delete` request.

js/reference.js
```diff
@@ -1807,18 +1808,21 @@
         },
 
         /**
-         * Deletes the referenced resources.
+         * Deletes the referenced resource's tuples.
          * NOTE This will ignore the provided sort and paging on the reference, make
          * sure you are calling this on specific set or rows (filtered).
          *
-         * @param {Object} contextHeaderParams the object that we want to log.
+         * @param {Object[]} tuples optional array of objects with just the data needed to construct the uri, if not-defined use the reference ermrestCompactUri (should only be empty for association reference)
```
RFSH: I don't think it's optional anymore. you might want to remove the association reference part too.

js/reference.js
```diff
@@ -1807,18 +1808,21 @@
         },
 
         /**
-         * Deletes the referenced resources.
+         * Deletes the referenced resource's tuples.
          * NOTE This will ignore the provided sort and paging on the reference, make
```
RFSH: This note is not needed anymore. We're creating the url using the given tuples.

js/reference.js
```diff
@@ -1839,7 +1843,26 @@
...
+                var uri = this.table.uri;
+
+                var keys = [];
+                //create key for each tuple and push it into the key array
+                for (var i=0; i<tuples.length; i++) {
+                    // NOTE: should probably be factored out into a function of table
+                    var keyPair = "";
+                    for (var j=0; j<this.table.shortestKey.length; j++) {
+                        var colName = this.table.shortestKey[j].name;
+                        keyPair += module._fixedEncodeURIComponent(colName) + "=" + module._fixedEncodeURIComponent(tuples[i][colName]);
```
RFSH: Since it's now an array of objects, I think we should check for tuples[i][colName] to be available. You might want to throw an error if it's not available.

test/specs/reference/tests/11.delete.js
```diff
@@ -4,14 +4,18 @@ exports.execute = function (options) {
...
         var baseUri = options.url + "/catalog/" + catalogId + "/entity/" + schemaName + ':' + tableName;
 
-        describe('when deleting regularly', function() {
+        var associationUri = options.url + "/catalog/" + catalogId + "/entity/" + schemaName + ':' + assocTableName;
+
+        describe('when deleting one row', function() {
+            var reference, tuples;
+
             beforeAll(function (done) {
                 var uri = baseUri + "/key_col=1";
```
RFSH: Just to make sure we're actually creating the url and not using the reference url, we could just change this test case to not have the filter. Also make sure that the table has more than one row, then delete one row and see if the other rows exist or not. All the test cases are currently deleting all the rows that are available on the table (which is not testing the actual change in this PR).

jrchudy:
@RFSH  I made changes based on your suggestions. Should I change the error message that I throw if the submitted data doesn't match the shortest key column name? Should I test this error case here in `ermrestJS`?

js/reference.js
```diff
@@ -1839,7 +1841,27 @@
...
+                var uri = this.table.uri;
+
+                var keys = [];
+                //create key for each tuple and push it into the key array
+                for (var i=0; i<tuples.length; i++) {
+                    // NOTE: should probably be factored out into a function of table
+                    var keyPair = "";
+                    for (var j=0; j<this.table.shortestKey.length; j++) {
+                        var colName = this.table.shortestKey[j].name;
+                        if (!tuples[i][colName]) throw new module.BadRequestError(400, "tuples[" + i + "] does not have any data for shortest key column: " + colName);
```
RFSH: As we discussed this should be a `verify` statement, so we're sending an ermrestjs error and not ermrest. We don't need to mask this as an ermrest error.

PR #670, opened 2018-07-03
Update markdown-it-attrs

jrchudy:
Updated the `markdown-it-attrs` dependency for `markdown-it`. The version we were using before was `v0.6.3`. The new version is `v2.2.0`. There have been 2 major releases that focused mostly on fixing bugs and adding support to new elements. With the release of `1.0.0`, a complete rewrite of the library was done

Some changes to note:
 - [Better support](https://github.com/arve0/markdown-it-attrs/issues/32) for attributes on `<ul>`
 - [A fix](https://github.com/arve0/markdown-it-attrs/issues/36) for issues with blockquote syntax
 - [Support](https://github.com/arve0/markdown-it-attrs/issues/39) for attributes on markdown tables
 - [Support](https://github.com/arve0/markdown-it-attrs/pull/48) for CSS-modules (`..green` -> `css-module="green")
 - ability to configure what we want to use as a delimiter (instead of `{` and `}`)
 - [Support](https://github.com/arve0/markdown-it-attrs/issues/67) for the horizontal line element

This allows us to attach attributes to markdown tables. The following template is how it will behave now:

This template:
`|heading|\n|-|\n|text|\n{.class-name}` 
produces:
```html
<table class="class-name">
  <thead>
    <tr>
      <th>heading</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>text</td>
    </tr>
  </tbody>
</table>
```

Whereas before this update was added, the same template would produce:
```html
<table>
  <thead>
    <tr>
      <th>heading</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>text</td>
    </tr>
  </tbody>
</table>
<p class="class-name"></p>
```
With the class name attached to an extra `<p>` tag after the content.

NOTE: There is one issue with this updated content. For the case of `dropdown`, if a class is supplied to the first caption of the tag:
`::: dropdown MyCaption{.some-class} [Caption1](some link) ...`

The class (`.some-class`) is applied appropriately to the first button element we draw. But it isn't properly parsed/stripped off of the text that is displayed. The button displays `MyCaption{.some-class}` as the the text. Some changes were made to strip off the last attribute block from the content so it behaves the same as before.


npm-shrinkwrap.json
```diff
@@ -1,11 +1,1823 @@
 {
```
RFSH: I don't think we want to update this file. Just revert it back to what it was. I think this happened because of your nodejs version.

RFSH:
Also please add the table example to the mustache/markdown document that we have:
https://github.com/informatics-isi-edu/ermrestjs/wiki/Template-and-Markdown-Guide

PR #682, opened 2018-07-31
create entity w/ non-insertable columns

robes:
Before creating an entity, check if the user does not have permission to
insert into any column. Add non-insertable columns to the 'defaults'
argument for the request.

This addresses bug documented in the Chaise repo.
https://github.com/informatics-isi-edu/chaise/issues/1610


robes:
I have tested this fix on dev.facebase.org catalog #5

RFSH:
We decided to change the implementation a little bit. These are the summary of changes:

- Instead of checking for just the `insert` access, changed it to check for the `inputDisabled` attribute that the client will use. This will take care of annotations too.
- Changed the `Automatically generated` to `Not Allowed` if the user doesn't have `insert` access to the column.
- Added test cases. To do so, I had to move the utility functions that we had for permission test cases to a common file that I can reuse.

Issue #666, opened 2018-06-20
Change Default Heuristics for Int/serial Key Columns


hongsudt:
We want these changes to be made for heuristics fo int/serial key columns:
- [x] Use choice facet by default for int/serial key columns.
- [x] Turn off the value formatting (no thousand separator).

hongsudt:
We should also turn off the default int formatting (no thousand separator).

Issue #637, opened 2018-05-17
Support pre_format in Facet Values


RFSH:
We should support `pre_format` in the facet values. Currently it's directly calling the `_formatValueByType` function, we might want to call the `formatvalue` of the column object instead. Although we should be cautious of the extra logic that is in that function (the logic to deal with `json` for example).

RFSH:
AggregateGroupReference and Faceting APIs are now using the same `formatvalue` function of column, therefore they should now support pre-format annotation.

PR #686, opened 2018-08-14
formatvalue changes

RFSH:
This PR changes the formatvalue logic throughout ermrsetjs. When I implemented the faceting, I started using `_formatValueByType` directly in different places. In this PR, I tried eliminating those usages. This includes:

- Turning off the value formatting (no thousand separator) for int/serial simple key and foreignkey columns.
- Making sure that facet filter functions are using the `column.formatvalue` instead of calling the format function directly.
- Passing the base column to `AggregateGroupColumn` and using the `baseColumn.formatvalue` to return the formatted value.
- Adding context to `AggregateGroupReference`. For now I'm just passing the context string. At first I was doing the same thing that we're doing in reference (creating a contextualize object and attaching it to reference), but I decided to avoid over-engineering this for now.
- Adding the base column parameter to entityCounts and entityValues as well as the histogram APIs.

Issues: #666  #637


RFSH:
Since I was changing the `formatvalue` logic, I decided to do the #637 as part of this PR too. 
Now we just have one central place for getting the formatted value of columns. Faceting and aggregate group APIs are going to use that function from now on.
@jrchudy Can you please take another look at these new changes?

PR #693, opened 2018-08-21
add catalog snapshot, version, and id to mustache and handlebars environments

jrchudy:
This PR includes the first set of changes for [history support](https://github.com/informatics-isi-edu/chaise/issues/1502).

The changes included add 3 new variables to the markdown environment. These new variables are all nested under the `$catalog` object. These variables allow the data modeler to specify a full `snapshot` value for a catalog or each individual part, `id` and `version`.

For example, a catalog with snapshot `1@1234` will be defined as the following:
```js
$catalog: {
  snapshot: "1@1234",
  id: "1",
  version: "1234"
}
```

### DIFF
--- a/test/specs/print_utils/tests/01.print_utils.js
+++ b/test/specs/print_utils/tests/01.print_utils.js
@@ -2,7 +2,6 @@ exports.execute = function (options) {
     var module = options.includes.ermRest;
     var formatUtils = module._formatUtils;
     describe("Print utils, For pretty printing values based on a value's type, ", function () {
-        var catalog = {id: process.env.DEFAULT_CATALOG};
         // Test Cases:
         it('printFloat() should format floats correctly.', function () {
             var printFloat = formatUtils.printFloat;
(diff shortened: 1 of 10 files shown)


jrchudy:
@RFSH I made the changes you suggested to include only the catalog rather than passing the table as well. I didn't add a test for `page.content` because that was just testing the combination of all the `rownames` for the fetched tuples which is being tested by the test case I added.

As I was fixing the test cases in the `print_utils` spec, I noticed that passing just the catalog id instead of the catalog might be better. But if we extend the API later to need something from the catalog, this would have to be changed. For the tests in that spec I "faked" a catalog object by passing an object with only an `id` property defined.

test/specs/print_utils/tests/01.print_utils.js
```diff
@@ -2,6 +2,7 @@ exports.execute = function (options) {
     var module = options.includes.ermRest;
     var formatUtils = module._formatUtils;
     describe("Print utils, For pretty printing values based on a value's type, ", function () {
+        var catalog = {id: process.env.DEFAULT_CATALOG};
```
RFSH: I know currently you are only using `id`, but it's better if you pass the actual catalog object. You can access the current catalog using `options.catalog`.

RFSH: Or you can even change the catalog parameter of those functions to be optional. So if you pass the catalog we're going to attach `$catalog` and if you don't, we're not. I think this might be the better solution. You just need to add a check to `_addErmrestVarsToTemplate` function to make sure catalog is defined and then use it.

RFSH:
When I added `$fkeys` I forgot to add the support to `page.content` and also forgot about the test case.  Can you please modify the current test case to test `$catalog` too? [This](https://github.com/informatics-isi-edu/ermrestjs/blob/catalog-variables/test/specs/annotation/tests/03.table_display.js#L520,L533) is the test case, and [this](https://github.com/informatics-isi-edu/ermrestjs/blob/catalog-variables/test/specs/annotation/conf/table_display/schema.json#L433) is the annotation that you can add a `$catalog.id` to it.

You don't need to add a extra test case, just modifying the annotation and expected value should be enough.

jrchudy:
@RFSH I did the following:
 - made catalog an optional parameter to `_addErmrestVarsToTemplate`
 - modified the annotation and `page.content` test case
 - added a test case in `print_utils` for when `_renderMustacheTemplate()` and `_renderHandlebarsTemplate()` are called without a `catalog` parameter to test for completeness for those functions

Issue #690, opened 2018-08-16
Allow Custom Sort/Display In Scalar Facets


RFSH:
We want data modelers to be able to have more control over what the facet column is showing. These are the new set of changes that we thing could be useful (These are only for **scalar** facets in **choices** UX mode):

- [x] **Custom sort**: Currently we're always sorting the values by descending value of frequency (and tie breaking with ascending value of the column). We want data-modelers to be able to change that. To do so, they can provide `column_order` in the facet object (same syntax as `column_order` defined in annotation). For example you can define your facet this way:

```json
{
  "source": [{"inbound": ["schema", "constraint"]}, "column"],
  "order": [{"column": "sort_col", "descending": true}]
}
```

To also give them the ability to add the frequency in their `column_order`, we're proposing special `$frequency` column name. This way you can have the frequency in your own custom `column_order`

- [x] **Hide Frequency Column In Modal**: In most cases, we don't want to show the frequency column at all in faceting modal. Currently we're showing it just so users can sort based on that. To remove that we need to add a new attribute to facet JSON.

```
{
  "source": [{"inbound": ["schema", "constraint"]}, "column"],
  "hide_num_occurrences": true
}
```

I am going to implement the first suggestion, but I feel like we need to discuss more about the second one.

@hongsudt @karlcz @robes Can you please review this? Do you think these features would be useful?

Issue #497, opened 2017-09-15
entityCounts and entityValues should honor column_order


RFSH:
Currently we're sorting based on raw values of the column. We can change the implementation to honor column_order. This should be possible by adding those columns to the projection list.

PR #694, opened 2018-08-23
Add columnOrder support to AttributeGroup and Faceting

RFSH:
These are the summary of changes in this PR:

`ReferenceColumn.AggregateGroup.entityCounts`:

This function is currently used to return the scalar values and their frequency. I added two parameters to this function. 
  - `sortColumns`: The columns that should be used for sorting the reference.
  - `hideNumOccurrences`: Whether we should add the frequency column or not. Since now this is an option in this function, I removed the `entityValues` which is the same as passing `true` for this parameter.

`FacetColumn`:
  - `sortColumns`: This is coming from the `sort` attribute defined on the facet. This will be used for having custom sort while showing the values of a scalar facet.
  - `hideNumOccurrences`: The `hide_num_occurrences` attribute in facet JSON dictates the value of this attribute.
  - `scalarValuesReference`: Currently in chaise we're calling the `facetColumn.column.entityCounts`. Now that I have added two variables to this function, I felt like it is better if I do this in ermrestjs. This will return an `AttributeGroupReference` that has the scalar values.

With these changes, the "custom scalar facet sort" has been added to ermrestjs. The next change is to add `column_order` support to the columns of `AttributeGroupReference`. What this means to chaise is that if you open the facet modal and sort based on a scalar column, that will honor the `column_order` and won't necessarily be sorted based on the raw value.

`AttributeGroup` APIs:
  - Add context  and sorting APIs to the column object.
  - Add `ermrestSortObject` API which will return the sort object that will be used for constructing the url. Modified the `ermrestPath` to use this API. If there's a column in `ermrestSortObject` that is missing from the list of columns, we have to add it to the key columns. That's why I'm passing the "term" too (I added a few more comments in the code).

Other changes:
- Fixed the "occurrences" typo (it was "occurences").

Issues: #690 #497


js/ag_reference.js
```diff
@@ -92,8 +106,14 @@ AttributeGroupReference.prototype = {
         return this._columns;
     },
 
+    /**
+     * Returns the visible key columns
```
jrchudy: Is it possible to not have any key columns visible?

js/ag_reference.js
```diff
@@ -371,7 +484,30 @@ AttributeGroupReference.prototype = {
...
+        var findCol = function (list) {
+            for (i = 0; i < list.length; i++) {
+                if (list[i].name === name) {
+                    return list[i];
+                }
+            }
+            return false;
+        };
+
+        var c = findCol(this._allColumns);
```
jrchudy: if `_allColumns` is a `Columns` object like we have defined in `core.js`, this function is redundant because of `Columns.get()`.

js/ag_reference.js
```diff
@@ -784,6 +938,63 @@ AttributeGroupColumn.prototype = {
...
+     * @desc An array of objects that have `column` which is column name (this is different from other sortColumns),
+     * and `descending` which is true/fals. The `descending` boolean indicates whether we should change the direction of sort or not.
+     * The column name is going to be database column names (equivalent to this.term)
+     *
+     * - if sortable passed as false, it will be empty.
+     * - if baseColumn exists, it will return the column order of baseColumn
+     * - otherwise the current column
+     * @type {Object[]}
+     */
+    get _sortColumns() {
```
jrchudy: I thought these functions were already available on the column object? 

Unless the body of determine sortable is different here, which would make sense why this is redefined.

js/column.js
```diff
@@ -2212,6 +2212,66 @@ FacetColumn.prototype = {
...
+        return this._hideNumOccurrences;
+    },
+
+    /**
+     * Returns the sortColumns when we're sorting this facet in scalar mode
+     * - uses row_order if defined.
+     * - otherwise it will be descending of num_occurrences and column order of base column.
+     * @type {Array}
+     */
+    get sortColumns() {
```
jrchudy: Maybe the name of this function should be changed.. It makes me think it's for a "set of columns" rather than the "columns used to sort the current column".

js/reference.js
```diff
@@ -1226,10 +1226,7 @@
                     sortObject = this.display._rowOrder.map(function (ro) {
                         return {"column": ro.column.name, "descending": ro.descending};
                     });
-                    sortColNames = {};
-                    sortObject.forEach(function (so) {
-                        sortColNames[so.column.name] = true;
-                    });
+                    processSortObject(this);
```
jrchudy: Was this function defined before and it just wasn't being used here yet?

js/ag_reference.js
RFSH: It's not. It's an array of `AttributeGroupColumn` object which is different from `Column`. Besides I didn't follow the pattern of creating objects for each array container like we have in `core.js` (So the `_allColumns` is actually an array and not an object that contains an array). It might look better, but it's just redundant and reduces the flexibility.

js/column.js
RFSH: I can remove these APIs (sortColumns and hideNumOccurrences) now that I have added `scalarValuesReference`. I created these two at first, then I decided to add the third API. That's the only place that these two APIs are going to be used and it's internal.

But the name `sortColumns` is what I have used for other types of column, I can change that if you think it's not clear. What is a better name for it?

jrchudy:
@RFSH You can merge this PR. Make sure to merge this before merging the [chaise PR](https://github.com/informatics-isi-edu/chaise/pull/1626).

PR #707, opened 2018-09-07
Remove file size comparison and check chunks.complete

jrchudy:
Changed updateProgressBar function to check for the total number of chunks that have `complete == true` rather than relying on the uploadEventHandler progress function when it updates the current progress for each chunk.

The uploadEventHandler triggers when the browser has sent bytes to the server which will return to the client before the actual bytes have been written to the file. Locally this happens pretty quickly, but when there is a long delay over the network that may fluctuate, the client may report uploaded bytes much sooner than those bytes have been written.

We saw a case where 3 chunks finished uploading after the POST for the file. One can assume that the uploadEventHandler progress function was returning that it had finished uploading the chunk much sooner than when the callback to upload the chunk actually occurred. 

We wrongly based the completion of a file upload job based on the total size that had been transferred to the server rather than when every chunk's success callback function was invoked.

### DIFF
--- a/js/hatrac.js
+++ b/js/hatrac.js
@@ -834,14 +834,16 @@ var ERMrest = (function(module) {
      */
     upload.prototype._updateProgressBar = function(xhr) {
         var length = this.chunks.length;
-        var done = 0;
-        for (var i = 0; i < this.chunks.length; i++) {
-            done = done + this.chunks[i].progress;
+        var progressDone = 0;
+        var chunksComplete = 0;
+        for (var i = 0; i < length; i++) {
+            progressDone = progressDone + this.chunks[i].progress;
+            if (this.chunks[i].completed) chunksComplete++;
         }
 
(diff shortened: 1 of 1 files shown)


jrchudy:
@RFSH  I have updated the body of this PR and made sure that both `ermrestJS` and `chaise` tests are unaffected.

RFSH:
I might be repeating what you explained in the description, so sorry in advance. I just want to make sure that I understand it.

To summarize, you changed the `upload.completed` to be based on the number of `chunk.completed` instead of summation of `chunk.progress`, because the value of `chunk.progress` might be wrong. And the reason for that is client calling the `uploadEventHandler` before actually finishing the chunk upload request.

If what I said is correct, then why are we not getting rid of `uploadEventHandler`? It's changing the value of `chunk.progress` which is something that users will see. Shouldn't we just update the `chunk.progress` on the success callback of chunk upload?

PR #716, opened 2018-09-18
query timeout error class added

jrchudy:
This PR adds support for the handling of the special case of `400 Bad Request` that returns a 400 error with the body: `Query run time limit exceeded`. 

This error case was added in [this ermrest PR](https://github.com/informatics-isi-edu/ermrest/commit/11e17a97e3838b3726771d08b6fa289081909b80) and outlined by [this chaise issue](https://github.com/informatics-isi-edu/chaise/issues/1644).


jrchudy:
@RFSH  The changes for this PR are all set and ready to be reviewed.

js/ag_reference.js
```diff
@@ -171,6 +174,17 @@ AttributeGroupReference.prototype = {
...
      },
 
+     /**
+      * This will generate a new unfiltered reference each time.
+      * Returns a reference that points to all entities of current table
+      *
+      * @type {ERMrest.Reference}
+      */
+     get unfilteredReference() {
+         var newLocation = new AttributeGroupLocation(this.location.service, this.location.catalogId, [module._fixedEncodeURIComponent(this.table.schema.name),module._fixedEncodeURIComponent(this.table.name)].join(":"));
```
RFSH: Add a check to verify that `this.table` is defined. If it doesn't exist you can either return `null`, or throw an appropriate error (if you just call verify function that'll do it).

js/ag_reference.js
```diff
@@ -1028,6 +1042,8 @@ function AttributeGroupLocation(service, catalog, path, searchObject, sortObject
      */
     this.catalogId = catalog;
 
+    this.api = "attributegroup";
```
RFSH: Was there any reason behind adding this? Is chaise looking for this attribute?

test/specs/ag_reference/tests/01.ag_reference.js
```diff
@@ -59,6 +59,7 @@ exports.execute = function (options) {
 
         describe("AttributeGroupLocation, ", function () {
             beforeAll(function () {
+                console.log("before all tests");
```
RFSH: I think you missed this.

js/ag_reference.js
jrchudy: I updated all places in the code to include a table in the constructor for `AttributeGroupReference` so if one of these references is missing the property and `unfilteredReference` is called, I think it should throw an error. 

There are 2 places in the tests that I didn't add the table to the constructor where I could test the error case of this function.

jrchudy:
@RFSH  the changes for this PR have been complete. This PR should be merged before the corresponding [chaise PR](https://github.com/informatics-isi-edu/chaise/pull/1647).

PR #717, opened 2018-09-22
always hide the null choice

jrchudy:
@hongsudt  asked me to disable this feature for now so we can avoid some degraded (and possibly confusing) functionality.


RFSH:
The build failed on the `make install` step, but you need to fix the test cases too.

RFSH:
It has the same issue as the ermrsetjs PR. The test cases need to be removed too.

Issue #701, opened 2018-08-31
hueristics for export 


hongsudt:
If the export annotation doesn't exist, we should apply the following heuristics that generate a bag of files consisting of: 
- [x] csv file for the main entity
- [x] csv file for all the related tables (whether in-line or not)
  - single hop : the fk to the main entity should be there already
  - multi-hops: add fk to the main entity as an extra column
- [x] fetch output for all the columns that have asset annotation

RFSH:
In the following I will try to explain how export is working and how I am generating the default export template.



Based on current definition of the annotation, you need to pass `api`, `table` (in "schema:table" format), and an optional `path`. This is how export is using these:

1. If the given table is the same as the current table that reference is representing:

```
<api>/<current_reference_path>/<path>
```

2. If the given table is not the same as the current table:

```
<api>/<current_reference_path>/X:=<table>/<path>
```

So it's assuming that the table that you're defining (X) has a direct relationship with the reference table (M). But for the inbound related tables that have longer path this is not true, so I ended up defining them this way:

```
{
  "table": "schema:main_table",
   "path": "the foreignkey path to the related table in ()=() format"
}
```

Another requirement was adding the main table's key to the projection list. To do so, I had to use the alias that I'm using in the parser logic. I think we should add this to the export documentation, I am always using the `M` alias for the table that is being projected. 


## Example

Having the following ERD, with the given annotation for the table.

![ERD](https://dev.isrd.isi.edu/~ashafaei/wiki-images/pseudo_col_erd.png)

```javascript
"tag:isrd.isi.edu,2016:visible-foreign-keys": {
  "*": [
    ["schema", "fk2_cons"], 
    ["schema", "fk3_cons"],
    {
      "source": [
        {"inbound": ["schema", "fk3_cons"]},
        {"outbound": ["schema", "main_f3_cons"]},
        {"inbound": ["schema", "f4_cons"]},
        "f4_id"
       ]
     }
  ]
}

```

This would be the generated default export template:


```javascript
{
  "name": "default_template",
  "format_name": "BAG",
  "format_type": "BAG",
  "outputs": [
    {
      "destination": {
        "name": "main",
        "type": "csv"
      },
      "source": {
         "api": "entity",
        "table": "schema:main"
      }
    },
    {
      "destination": {
        "name": "f2",
        "type": "csv"
      },
      "source": {
        "api": "entity",
        "table": "schema:f2"
      }
    },
    {
      "destination": {
[...]

PR #718, opened 2018-09-27
Default export

RFSH:
This PR will add `defaultExportTemplate` and `exportTemplates ` to reference. We wanted an API to return the default export template to chaise when templates are not defined in the schema. To do so, I created the `exportTemplates` API in `Reference` which will return the default in the described situation (no template + detailed context).

I am creating the outputs as it is described in the issue, but while doing so I realized that we might want to re-evaluate the template definition or ermrestjs implementation of it. I added [a new comment](https://github.com/informatics-isi-edu/ermrestjs/issues/701#issuecomment-425168368) to the issue to explain this. We need to discuss about this.


RFSH:
I'm going to close this PR for now, since we decided to do more changes. I described the changes [in the issue](https://github.com/informatics-isi-edu/ermrestjs/issues/701#issuecomment-426851981).

PR #719, opened 2018-09-27
modify hideNullChoice logic

RFSH:
This PR will change the `hideNullChoice` to be based on what the issue describes. I only wasn't sure about the last note, so I didn't add that. We can add that later as an improvement to this.

P.S. Chaise test cases will break after merging this, I will update chaise test cases and create another PR.


RFSH:
Further changes are required for this PR. Detailed explanation of how we should present and treat `null` are described in #728 issue.

Issue #677, opened 2018-07-17
Expose Why CanCreate and CanUpdate Return False


RFSH:
Currently we're just returning false without giving any reason. This can be confusing on the client side.

RFSH:
@hongsudt Yes, we wanted a better error message. Currently in recordedit we're just checking for `canCreate` and if it returns `false` we're just throwing a forbidden error with the "You are not authorized to perform this action" message. 
The simple solution is printing the reason in console, but the better solution would be ermrestjs returning the reason to client and client displaying that in the message details.

jrchudy:
The changes for this issue have been merged into master. reassigning to @hongsudt  for verification.

PR #720, opened 2018-09-29
Can create and can update false reason

jrchudy:
This PR addresses issue #677 . The following text will be added before the error text (`You are not authorized to perform this action. Please report this problem to your system administrators.`) when the user navigates to `recordedit` for a table they can't create or update on:

Text reasons for `canCreate`:
```
  case view:
     "Table is a view."
  case generated:
     "Table is generated."
  case no insert permission for user:
     "No permissions to create."
  case all columns disabled:
     "All columns are disabled."
```

Text reasons for `canUpdate`:
```
  case view:
     "Table is a view."
  case generated:
     "Table is generated."
  case no insert permission for user:
     "No permissions to update."
  case all columns disabled:
     "All columns are disabled."
```


js/reference.js
```diff
@@ -877,18 +906,50 @@
...
+        /**
+         * Indicates the reason as to why a user cannot update for
+         * the referenced resource(s). In some cases, this won't be set
+         * because the user can update, so the value will be `undefined`.
+         * @type {(String|undefined)}
+         */
+        get canUpdateReason() {
+            if (this._canUpdateReason === undefined) {
+                // will set the _canUpdateReason property
+                this.canUpdate();
```
RFSH: canUpdate is not a function, so you cannot call it that way. You have to say `var res = this.canUpdate;`

js/reference.js
```diff
@@ -836,18 +836,47 @@
...
+        /**
+         * Indicates the reason as to why a user cannot create for
+         * the referenced resource(s). In some cases, this won't be set
+         * because the user can create, so the value will be `undefined`.
+         * @type {(String|undefined)}
+         */
+        get canCreateReason() {
+            if (this._canCreateReason === undefined) {
+                // will set the _canCreateReason property
+                this.canCreate();
```
RFSH: Same comment as `canUpdateReason`, `canCreate` is not a function.

js/reference.js
```diff
@@ -836,18 +836,47 @@
                 // 3) not all visible columns in the table are generated
                 var ref = (this._context === module._contexts.CREATE) ? this : this.contextualize.entryCreate;
 
-                this._canCreate = ref._table.kind !== module._tableKinds.VIEW && !ref._table._isGenerated && ref._checkPermissions("insert");
+                if (ref._table.kind === module._tableKinds.VIEW) {
+                    this._canCreate = false;
+                    this._canCreateReason = "Table is a view. ";
```
RFSH: Can you please add an object like the ones that we have [at the bottom of utilities.js](https://github.com/informatics-isi-edu/ermrestjs/blob/master/js/utilities.js#L3182) to have these messages in one place?

hongsudt:
Instead of putting those sentences at the beginning, it might be better to put them in the detail section.

Issue #715, opened 2018-09-18
Share: use resolver_implicit_catalog in generating a record shared link instead of chaise url


hongsudt:
chaise config name : `resolver_implicit_catalog`. The scope is only on record page:
* `undefined` or absent: use the chaise path (same as existing permalink) 
* `false`: use `/id/{{{catalog_number}}}/{{{RID}}}`
* `<implicit catalog #>`: two scenarios
  - if catalog number in the chaise URL is the same as `<implicit catalog #>` then use `/id/{{{RID}}}`
  - else, use `/id/{{{catalog_number}}}/{{{RID}}}` 

Note: We will be adding `@{{{version}}}` after `{{{RID}}}` when we have a way to navigate between snapshots through chaise.

hongsudt:
@carlkesselman if I also have a DOI, do you want to also listed DOI as one of the shared links?

hongsudt:
@jrchudy I added this config to rbk-dev to reflect implicit default catalog, which is 2.

jrchudy:
The changes for this issue have been merged into master, reassigning to @hongsudt  to verify the feature is working as expected.

jrchudy:
@karlcz  pointed out that if the catalog defined in the config option, `resolverImplicitConfig`, is not the same as the current catalog number, we should use the current catalog number like `/id/{{{current_catalog_number}}}/{{{RID}}}`.

This is currently implemented incorrectly.

PR #721, opened 2018-10-04
added catalog id and version to location object

jrchudy:
This PR has changes that chaise relies on to resolve issue #715. The location object has 2 new APIs, `catalogId` and `version` which can be used to get the individual parts of the the catalog identifier provided.

### DIFF
--- a/js/parser.js
+++ b/js/parser.js
@@ -106,7 +106,11 @@
         this._service = parts[1];
 
         // catalog id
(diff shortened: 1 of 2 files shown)


js/parser.js
```diff
@@ -494,6 +498,22 @@
             return this._catalog;
```
RFSH: As we discussed this should be changed to `this._catalogSnapshot`

PR #754, opened 2019-01-29
Py3

RFSH:
Recently all the back end components have been updated to use python3. This PR is going to change `travis.yml` to use the python3 version of our back end components.

I created this PR just as a reminder. I'm not sure when would be the appropriate time to merge this PR.


jrchudy:
From the discussion we had this morning, when we have a stable version of each of our deriva modules running with python 3, we can merge this PR after changing the branch checkouts back to "master" for each repository.

PR #756, opened 2019-02-04
Pseudo array options

RFSH:
This PR will add support of `array_options` to array aggregate pseudo-columns. It is going to have two attributes:
  - `order`: Has the same syntax as `column_order` to modify ordering of the returned values.
  - `max_length`: Will dictate the maximum length of returned values.

These two attributes are going to affect the returned result. The have no effect on the generated ermrest query.

Other changes:
- I realized that I am using the `row_name/<context>` for generating list of row-names in entity array aggregate. After a discussion, we decided to change this to always use the `row_name/compact` context regardless of its parent reference context.
- Add a `compare` function to Column object. Which is used for computing the added `order` attribute.


RFSH:
One thing that is worth mentioning is that currently there's an `array_display` attribute available on pseudo-columns (it will control how to join the array of values). I did not make any change to that attribute. In the near future, with the addition of `display` to pseudo-columns, I will move that attribute. For now, it remains the same.

PR #760, opened 2019-02-08
public renderHandlebarsTemplate function

jrchudy:
Added public function for `_renderHandlebarsTemplate()`. I have 2 questions about this:
 1. Right now the catalog object is included but only for the purpose of providing the catalog ID
  - should that continue to be the case?
  - or should we allow the public API to pass just an ID that gets added to a faux catalog object?

 2. What should the tests for this feature look like? 
   - If we keep the function as is, the tests would be redundant after making sure that the API does work for one case. 
   - If we change the function to just pass in the catalog ID, then testing that case would make sense.


jrchudy:
@RFSH  I made the changes that you suggested.

PR #761, opened 2019-02-12
Record display options

jrchudy:
This PR has changes for `ermrestJS` to help support changes to `chaise` for [this issue](https://github.com/informatics-isi-edu/chaise/issues/1710) in the chaise repo. Added more options to the table display annotation. I think that's the right place to add tests for this feature. Should I also test the case that it is not defined?


test/specs/annotation/tests/03.table_display.js
```diff
@@ -62,6 +63,9 @@ exports.execute = function (options) {
         var tableCatalogAnnotEntityUri = options.url + "/catalog/" + catalog_id + "/entity/" + schemaName + ":" +
             tableNameCatalogAnnot;
 
+        var tableCompactOptionsEntityUri = options.url + "/catalog/" + catalog_id + "/entity/" + schemaName + ":" +
```
RFSH: Change `tableNameCatalogAnnot` to`tableNameCompactOptions`.

RFSH: I didn't notice test cases are failing, so I approved it. But now I realized why your new test case is failing. You just need to change the table name to be the new table that you have created.

jrchudy: Ah, thank you. I couldn't run the tests locally last night so I opened the PR so the tests could run and I would debug them after. Thank you.

Issue #748, opened 2018-12-08
Support display in Source Definition


RFSH:
**You can ignore the comment on this issue. I updated this description to be based on what we want to implement.**

We wan to be able to modify the display value of pseudo-column or related references. For that, we propose adding `display` support to the following:

- `visible-foreign-keys`: all the related entities. This will be an extra step for getting the markdown display of related entities (the alternative display that we have for the tabular display). 
- `visible-columns` when we are displaying a list of entities. That is inline table in record page, or entity aggregate pseudo-columns.

We are going to add a `"display"` attribute to pseudo-columns with the following attributes:

```
{
  "source": ..,
  "display": {
    "markdown_pattern": "value",
    "template_engine": ""
  }
}
```
In the `markdown_pattern` you can access to the current pseudo-column data with `$self` namespace. The structure of the available data is going to be different based on pseudo-column type.

- inline table (visible-foregin-keys):
```
{
  "$self": [
    {
      "values": {
        "RID": "",
        "_RID": "",
       "$fkeys": {}
      },
      "rowName": "",
      "uri": {
        "detailed": ""
      }
    },
    ...
  ]
}
```

- entity array aggregate:
```
{
  "$self": [
    {
      "values": {
        "RID": "",
        "_RID": ""
      },
      "rowName": "",
      "uri": {
        "detailed": ""
      }
    },
    ...
  ]
}
```

- all-outbound entity:

```
{
  "$self": {
    "values": {
      "RID": "",
      "_RID": ""
    },
    "rowName": "",
    "uri": {
      "detailed": ""
    }
  }
}
```

- scalar array aggregate:
```
{
  "$_self":  ["v1", "v2"],
  "$self":  "v1, v2"
}
```

- scalar min/max/cnt/cnt_d aggregate or any scalar column:
```
{
  "$self":  "1,234",
  "$_self":  1234
}
```

RFSH:
We had a long discussion about this. These are a few take aways from the meeting:

1. We might want to add these display settings to other types of pseudo-columns too (scalar aggregate, scalar all-outbound, entity all-outbound) and therefore the `table-display` won't be applicable to them. Besides, having context inside the pseudo-column seems confusing. So we decided to instead have a `display` which will have the similar attributes to `table-display`. So the pseudo-column definition would look like this:

```
"tag:isrd.isi.edu,2016:visible-columns": {
  "detailed": [
      {
        "source": [{"inbound": ["s", "fk"], "RID"}],
[...]

PR #765, opened 2019-03-07
add display to source syntax

RFSH:
This PR will add display support to source syntax. To do so I had to change the following routes of code:

#### `Reference.display`

I wanted to make sure the reference has knowledge of this new attribute. This is only used for nested tables (inline or related) in the `Page.content` function.

#### `Tuple.templateVariables`

This might not have been necessary now. But I felt like I should avoid computing these variables over and over again. It will be used for computing any templates.

####  `Page.content`
 
This is what chaise is displaying for nested tables (inline or related). the template is coming from `Reference.display` and the key-value pairs are coming from the `Tuple.templateVariables`.

#### `PseudoColumn.display`

The same way that Reference now has the knowledge of this new template, columns should be aware of it. This is used for generating presentation in both `formatPresentation` and `getAggregatedValue`.
 
#### `formatPresentation`

I didn't want to change the `Column.formatPresentation` because `Column` is supposed to be the columns on the database. Therefore I had to make changes for each individual `formatPresentation` that are overriding this parent function. This inlcudes:

- `ReferenceColumn.formatPresentation`: This function is being called for normal columns and also as part of heuristics for different types of pseudo-columns. The assumption is that this will be called only for scalar columns.

- `PseudoColumn.formatPresentation`: All-outbound entity and scalar columns rely on this function.

- `ForeignKeyPseudoColumn.formatPresentation`: single outbound entity.

- `KeyPseudoColumn.formatPresentation`: self-link.

- `AssetPseudoColumn.formatPresentation`: asset column presentation.

#### `PseudoColumn.getAggregatedValue`

This is the API that chaise uses for getting the aggregated values.

Other changes are just adding helper functions and test cases.

Related Issue: #748


RFSH:
@jrchudy The PR is now ready to be reviewed. I added a documentation explaining the structure of data that you have access to in the `markdown_pattern`. I'm now working on the next part of that documentation that will include more examples.

jrchudy:
This looks good. I think adding `templateVariables()` was a good idea, especially because there are used in almost every context now for display purposes.

PR #766, opened 2019-03-08
added custom base32 decode function for version snapshots.

jrchudy:
Added the base32 custom decode function to ermrestJS for ease of use and for testing. This is to help resolve [issue #1673](https://github.com/informatics-isi-edu/chaise/issues/1673) and [issue #1733](https://github.com/informatics-isi-edu/chaise/issues/1733) in the chaise repo.

### DIFF
--- a/js/utilities.js
+++ b/js/utilities.js
@@ -2948,6 +2948,50 @@
         return unescape(module._fixedEncodeURIComponent(JSON.stringify(obj)));
     };
 
+    module._versionDecodeBase32 = function (version) {
+        // use 5-bit value as index to find symbol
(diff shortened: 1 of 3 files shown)


js/utilities.js
```diff
@@ -2948,6 +2948,50 @@
         return unescape(module._fixedEncodeURIComponent(JSON.stringify(obj)));
     };
 
+    module._versionDecodeBase32 = function (version) {
```
RFSH: Please just add the jsdoc to explain what this function is doing. You don't need to explain the whole algorithm just a simple line of saying what is the input and what the output is going to be.

PR #767, opened 2019-03-12
added snaptime to the catalog object during introspection

jrchudy:
@RFSH  should I add a test case for this?

I don't think we can _rely_ on the value being right, but we can check if the value is present or not. I didn't see any other tests for the server or catalog classes.


jrchudy:
@RFSH  I made changes to fix the errors that we were seeing. 

First, my current snaptime request was wrapping it's "error" in the `module.responseToError()` function which was then passed to the catch handler of `_introspect()`. That catch handler would then wrap it a second time. That's why removing `/schema` from the mock url didn't work.

Second, I learned that we have to pass a catalog id that we don't recognize so that an error http response is returned. From what I can understand, you can only repaint an error response as another error response in nock. You can't repaint a success response as an error response.

test/specs/errors/helpers/http_error.js
```diff
@@ -1,68 +1,69 @@
...
 exports.setup = function(options) {
-	server = options.server;
-	ermRest = options.ermRest;
-	ermrestUrl = options.url.replace('ermrest', '');
+    server = options.server;
+    ermRest = options.ermRest;
+    ermrestUrl = options.url.replace('ermrest', '');
 };
 
 exports.testForErrors = function(method, errorTypes, cb, message, mockUrl) {
```
RFSH: I feel like now that you've spend the time to figure out this function, it's worth adding few comments explaining it. So we don't have to go through the same process again.

PR #777, opened 2019-04-09
IE incompatibility

Divyata-Singh:
- Made a note of some of the javascript coding constructs that are incompatible with IE


RFSH:
@Divyata-Singh I'm a little bit confused. Are we using any of the incompatibilities that you've listed here in ermrestjs/chaise? Or is this just a list of things that we should be cautious about and avoid?

Divyata-Singh:
I was using most of these things in Boolean-Search but I have removed them now. I haven't checked if these are being used in Chaise/Ermrestjs. Also, if you think this doc should be moved somewhere else then let me know.

PR #790, opened 2019-05-21
Conflict link

jrchudy:
Made changes to when create or update error out by passing the reference to the error mapping function. When it's a duplicate error, a uri is generated that's attached to the error object so chase can resolve a reference from it to generate an appLink.

### DIFF
--- a/js/errors.js
+++ b/js/errors.js
@@ -32,14 +32,18 @@
      * @param  {string} redirectPath path that would be added to the host to create full redirect link in Chaise
      * @constructor
      */
-    function ERMrestError(code, status, message, subMessage, redirectPath) {
+    function ERMrestError(code, status, message, subMessage, redirectPath, reference) {
+        this.errorData = {};
         this.code = code;
         this.status = status;
         this.message = message;
         this.subMessage = subMessage;
-        if(redirectPath !== undefined && redirectPath !== null){
-           this.errorData = {};
+        if(redirectPath !== undefined && redirectPath !== null) {
            this.errorData.redirectPath = redirectPath;
+        }
+
+        if(reference !== undefined && reference !== null) {
+           this.errorData.reference = reference;
         }
     }
 
@@ -136,8 +140,8 @@
      * @param  {type} subMessage technical message returned by http request
      * @constructor
      */
-    function ConflictError(status, message, subMessage) {
-        ERMrestError.call(this, module._HTTPErrorCodes.CONFLICT, status, message, subMessage);
+    function ConflictError(status, message, subMessage, reference) {
+        ERMrestError.call(this, module._HTTPErrorCodes.CONFLICT, status, message, subMessage, null, reference);
     }
 
     ConflictError.prototype = Object.create(ERMrestError.prototype);
@@ -168,8 +172,8 @@
      * @param  {type} subMessage  technical message returned by http request
      * @constructor
      */
-    function DuplicateConflictError(status, message, subMessage) {
-        ConflictError.call(this, status, message, subMessage);
+    function DuplicateConflictError(status, message, subMessage, duplicateReference) {
+        ConflictError.call(this, status, message, subMessage, duplicateReference);
     }
 
     DuplicateConflictError.prototype = Object.create(ConflictError.prototype);
--- a/js/errors.js
+++ b/js/errors.js
@@ -32,18 +32,14 @@
      * @param  {string} redirectPath path that would be added to the host to create full redirect link in Chaise
      * @constructor
      */
-    function ERMrestError(code, status, message, subMessage, redirectPath, reference) {
+    function ERMrestError(code, status, message, subMessage, redirectPath) {
         this.errorData = {};
         this.code = code;
         this.status = status;
         this.message = message;
         this.subMessage = subMessage;
         if(redirectPath !== undefined && redirectPath !== null) {
-           this.errorData.redirectPath = redirectPath;
-        }
-
-        if(reference !== undefined && reference !== null) {
-           this.errorData.reference = reference;
+            this.errorData.redirectPath = redirectPath;
         }
     }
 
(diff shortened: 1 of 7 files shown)


js/reference.js
```diff
@@ -1124,10 +1124,9 @@
                       "failed": null
                     });
                 }, function error(response) {
-                    var error = module.responseToError(response);
-                    return defer.reject(error);
+                    return defer.reject(module.responseToError(response, self));
```
RFSH: Since you've changed this, I feel like there's no point in having this reject callback. Just the catch-block should be sufficient.

js/utils/helpers.js
```diff
@@ -1268,16 +1271,30 @@
...
+          var conflictKeyValues = [];
+          primaryColumns.forEach(function (colName, idx) {
+              // strip off " character, if present
+              if (colName.indexOf('"') != -1) colName = colName.split('"')[1];
+              conflictKeyValues.push(colName.trim() + "=" + conflictValues[idx].trim());
+          });
+
+          var uri = reference.unfilteredReference.uri + '/' + conflictKeyValues.join('&');
+          var error = new module.DuplicateConflictError(errorStatusText, mappedErrMessage, generatedErrMessage);
+          error.uri = uri;
```
RFSH: Instead of attaching `uri` in here, it should be a property in the `DuplicateConflictError` prototype. Also it's better if you just pass the reference instead of a `uri` that chaise has to resolve again.

 So ine here

```javascript
var existingRowReference = new Reference(module.parse(uri),  reference.table.schema.catalog);
return new module.DuplicateConflictError(errorStatusText, mappedErrMessage, generatedErrMessage, existingRowReference);
```
And in the constructor of `DuplicateConflictError`, you just attach the `existingRowReference` (you most probably going to need to use a better name). So chaise can call this `exception.existingRowReference.contextualize.detailed.appLink`.

test/specs/errors/tests/07.http_409_error.js
```diff
@@ -137,13 +146,26 @@ exports.execute = function (options) {
...
 
-        it("if it's a duplicate key error, we should generate a more readable message.", function (done) {
-            nock(url, ops)
-                .get("/ermrest/catalog/1235")
-                .reply(409, duplicateErrorServerResponse)
-                .persist();
+        it("on create, if it's a duplicate key error, we should generate a more readable message.", function (done) {
+            duplicateReference.contextualize.entryCreate.create([{duplicate_id: 1}]).then(null, function (err) {
+                expect(err.code).toBe(409, "invalid error code");
+                expect(err.message).toBe(duplicateErrorMappedMessage, "invalid error message");
```
RFSH: You should test the new attribute that you've added here.

RFSH: And since testing in ermrestjs is easier, it's better if you add a test case for the composite key in here too.

jrchudy:
@RFSH, changes have been made to create the reference in `ermrestJS`. I changed the constructor `ERMrestError ` as well. I figured we may want that functionality in the future for other error types, not just duplicate or other conflict errors.

js/utils/helpers.js
```diff
@@ -1252,32 +1252,61 @@
...
+
+          // parse out column names and values from generatedErrMessage
+          var conflictKeyValues = [];
+          primaryColumns.forEach(function (colName, idx) {
+              // strip off " character, if present
+              if (colName.indexOf('"') != -1) colName = colName.split('"')[1];
+              conflictKeyValues.push(colName.trim() + "=" + conflictValues[idx].trim());
+          });
+
+          var uri = reference.unfilteredReference.uri + '/' + conflictKeyValues.join('&');
```
RFSH: Make sure reference is defined and then use it here. This code should not fail if the reference is missing since it's not a required parameter.

js/errors.js
```diff
@@ -32,15 +32,19 @@
...
         this.message = message;
         this.subMessage = subMessage;
-        if(redirectPath !== undefined && redirectPath !== null){
-           this.errorData = {};
+        if(redirectPath !== undefined && redirectPath !== null) {
            this.errorData.redirectPath = redirectPath;
         }
+
+        if(reference !== undefined && reference !== null) {
+           this.errorData.reference = reference;
```
RFSH: I think just adding the reference to the `DuplicateConflictError` for now is enough. And name it something more specific. Like the `duplicateReference` name that you've chosen. Just `ERMrestError.reference` is confusing as what is this reference of.

I understand that you've added it for future usage. But I think this will just be more confusing and we should be more specific about these optional parameters.

test/specs/errors/tests/07.http_409_error.js
```diff
@@ -137,15 +152,42 @@ exports.execute = function (options) {
...
+                var data = response.tuples[0].data;
+
+                // updateData = { "duplicate_id": 2 }
+                data.duplicate_id = 2
+
+                return duplicateReferenceUpdate.update(response.tuples);
+            }).then(null, function (err) {
                 expect(err.code).toBe(409, "invalid error code");
                 expect(err.message).toBe(duplicateErrorMappedMessage, "invalid error message");
+                expect(err.errorData.reference).toBeDefined();
```
RFSH: Instead of just checking for existence of reference, check for its path (reference.location.compactPath) to make sure we're extracting the correct column name and values.

js/utils/helpers.js
```diff
@@ -1252,32 +1252,61 @@
...
-          return new module.DuplicateConflictError(errorStatusText, mappedErrMessage, generatedErrMessage);
-      }
-      else{
+
+          // parse out column names and values from generatedErrMessage
+          var conflictKeyValues = [];
+          primaryColumns.forEach(function (colName, idx) {
+              // strip off " character, if present
+              if (colName.indexOf('"') != -1) colName = colName.split('"')[1];
+              conflictKeyValues.push(colName.trim() + "=" + conflictValues[idx].trim());
```
RFSH: I think you need to encode the value and column names. To make sure, you can just change your existing test case to use `duplicate id` instead of `duplicate_id` and change its column type so you can test the encoding too. You don't need to add new test case for it and just changing the existing test should be sufficient.

jrchudy:
@RFSH I updated this PR with changes again.

PR #792, opened 2019-05-22
add support for rid markdown template

jrchudy:
This PR adds a new markdown template that allows users to supply an RID that will create a template with a resolvable link to the record associated with that RID.


jrchudy:
@RFSH  this PR is now ready to review after adding the documentation.

jrchudy:
@RFSH  This PR has been updated so that rid templating is done with a simpler syntax, `[[<RID>]]`.

PR #793, opened 2019-05-23
add new parameter to skip 401 handling

RFSH:
A while ago I changed the [chaise `$http` usage to `ERMrest.http`](https://github.com/informatics-isi-edu/chaise/pull/1772). This [caused issues related to 401 handling](https://github.com/informatics-isi-edu/chaise/commit/194e6f76f8c1baf9b766cb8d83317b153a593cb8).

To summarize the issue, In `http.js` if a request throws 401, we're going to set a flag and any request after that won't be fired until chaise tells `http.js` that the user has logged in and can continue. This caused issues since after 401, chaise will send a request to webauthn in order to get the actual login link. 

In this PR, I fixed this by adding a new flag. If this flag is present, we're going to skip the 401 handling. So any chaise authen request should send this flag. But I'm still not sure about the following two:

1. What's the best name for this flag? I currently used `allowUnauthorized` to be aligned with the name that we have chosen for the 401 handling callback (`setHttpUnauthorizedFn`). But maybe `skipUnauthorized` would be a better name? 

2. While writing the test cases I realized that `ERMrest.http` is not wrapping the http errors in appropriate ermrestjs error objects. It will return the plain http response that we're getting from the server. It makes sense since the whole purpose of `http.js` was to be the same as Angularjs's `$http`. But this means that any where that we're using the `ERMrest.http`, in the catch-block, we have to call the `ERMrest.responseToError` function to repaint these http errors. And that's how we've done it in ermerestjs. However since now chaise is going to use the `ERMrest.http` directly, I don't think it would be appropriate for chaise to always call this function. Maybe by default I should call the `responseToError`? Although this won't work for all the cases because `ERMrest.responseToError`  has evolved and we're passing different parameters based on the context. So maybe it's fine that chaise has to call the `ERMrest.responseToError`?

### DIFF
--- a/js/http.js
+++ b/js/http.js
@@ -62,17 +62,17 @@
 
     /**
      * function that is called when a HTTP 401 Error occurs
-     * @type {httpUnauthorizedFn}: Should return a promise
+     * @type {function}: Should return a promise
      * @private
      */
-    module._httpUnauthorizedFn = null;
+    module._http401Handler = null;
 
     /**
      * set callback function which will be called when a HTTP 401 Error occurs
      * @param {httpUnauthorizedFn} fn callback function
      */
-    module.setHttpUnauthorizedFn = function(fn) {
-        module._httpUnauthorizedFn = fn;
+    module.setHTTP401Handler = function(fn) {
+        module._http401Handler = fn;
     };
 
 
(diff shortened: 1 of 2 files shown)


RFSH:
@jrchudy I created this PR so I can move on to the next task. I didn't add you as a reviewer yet since I think we should discuss this next week with @hongsudt.

RFSH:
We discussed about the flag name and got to the discussion of authorization vs. authentication. And realized even though the 401 http error message is "unauthorized", unauthenticated makes more sense for this 401 error handling. But in the end I didn't want to introduce a new terminology and to avoid deviating from the name that we've given to the error object (`UnauthroizedError`), I decided to be very speficic about the naming and use the name `skipHTTP401Handling`. I also renamed the `setHttpUnauthorizedFn` function to `setHTTP401Handler`.

And about the second point that I made in the description, we decided to not change the `ERMrest.http` and any where else that is going to use this http client, should call the `ERMrest. responseToError`.

PR #797, opened 2019-06-12
Pseudo-Columns in Templating Environment

RFSH:
## Overview

This PR will add pseudo-column in templates that are specified in the source syntax of `visible-columns` and `visible-foreignkeys`. To achieve this, we had to introduce a table level annotation that you can use to define the variables that should be available in any pattern environment.  The annotation would look like this:

```
"tag:isrd.isi.edu,2019:source_definitions" : {
  "columns": ["col1"] // a list of columns. true means all the columns
  "fkeys": [["schema", "fk1"]] // a list of fkeys. true means all
  "sources": {
    "some-key-name": {"source": [{"inbound": ["schema", "fk"]}, "RID"], "entity": true}
  }
}
```

Ultimately you can define any foreign-key or column in the `sources` but we are going to include `fkeys` and `columns` just for backward compatibility. Setting them to `true` means that you wan all the outbound foreign-keys or columns of the table.  Then your visible-columns could be like the following:

```
"tag:isrd.isi.edu,2016:visible-columns": {
  "detailed": [
      {
        "source": [{"inbound": ["s", "cons"}, "column"],
        "display": {
          "markdown_pattern": "<prefix>{{{key1.0.rowName}}} <suffix> {{{key1.0.values.RID}}}",
          "template_engine": "handlebars",
          "waitfor": ["key1"]
        }
      }
   ]
}
```

Based on the pseudo-column, the available values are going to be different. For example an all-outbound pseudo-column would only give you a row of data, while an entity array aggregate will provide you with an array of rows. This implementation is based on this [wiki page](https://github.com/informatics-isi-edu/ermrestjs/wiki/Pseudo-Column-in-Templating-Enviornments).

### Table.sourceDefinitions

Since the source-definitions is not context based, I added this API directly to the table and not the Reference object. This will include the following attributes:

- columns: a list of columns (added for backwards compatibility).
- foreignKeys: a list of outbound foreignkeys (added for backwards compatibility).
- sources: a key value pair from the name that is defined to the pseudo-column definition.
- sourceMapping: a key value pair from the hash name that I'm using internally for columns to the names that are defined in the `sources` object.

### ReferenceColumn.Waitfor

The other issue that we had to solve is the facet that we might need to send multiple requests for getting the data of the defined sources. Therefore some pseudo-column's value might take a while to load. For this we have added `waitfor`. Which will list the pseudo-columns that the current pseudo-column relies on. For instance in the given example, we won't load the data for this column until we have the data for `key1`.

### Reference.ActiveList

To make the processing of pseudo-columns in visible-columns and their `waitfor`s easier for chaise, I have introduced the `activeList`. Active list, is the list of secondary sources of data for computing the list of visible-columns and visible-foreignkeys. It will go over the list of visible columns and related entities and create these lists:

- `allOutbounds`: Used to determine all of the foreignkeys that we need to attach to the main entity request.

- `aggregates`: All the aggregate columns that we need to send a request for. Each element will have,
  - `column`: the aggregate column
  - `objects` (_maybe I should use a better name_): the referred objects. Array of objects that rely on this aggregate column (through `waitfor` or just their source definition). To avoid passing multiple objects around, each element of this array instead will have the instruction of how to retrieve this object (it has `column: true` to indicate it's from visible columns, `related: true` to indicate it's from related entities, and `index: integer` to find the element from the array). 

### ReferenceColumn.SourceFormatPresentation

Active list will help chaise to construct the requests and the universal templating object. Now, we needed a new function that passing this universal templating object would return the value of the column. To make things easier, I decided to create a whole new function instead of reusing the `formatPresentation`. Although its current implementation is a bit messy and repetitive. I'm not entirely sure about this part yet. I might refactor it if I get the chance.

## Test Cases

Since this PR required a lot changes in different places, the number of test cases are more than an average PR too. The following are test cases that I've added:

- source definition test spec
- usage of `sourcekey` in visible-columns
- usage of `sourcekey` in faceting
- usage of `sourcekey` in visible-foreignkeys
-  active list test spec

pending features:
- Allowing collision with column names in the defined source names.


js/reference.js
```diff
@@ -4731,6 +4936,14 @@
          }
 
 
+    };
+
+    function Citation () {
```
jrchudy: Did you mean to leave this here?

PR #798, opened 2019-06-28
Sytem Columns Heuristics

jrchudy:
This PR includes changes to the heuristics for the columns list when no visible columns annotation is defined. If either of the following chaise config propertied are defined (`SystemColumnsDisplayCompact` or `SystemColumnsDisplayDetailed`), the system column heuristics will be applied. If `true`, the heuristics will be applied to all system columns. If the property is defined as an array, only the system columns in that array will have the heuristics applied to them, the rest will be hidden (unless part of a foreign key).

The new heuristic can be summarized as:
If config options are defined, and no visible columns list, system columns should be at end of the list when not a part of a foreign key relationship. RID is the exception, and will show first as a key column, with a self link.

Note:  This PR should be merged with the chaise PR (_pending..._). The chaise PR will also include a basic test for this feature.


js/reference.js
```diff
@@ -2716,18 +2724,60 @@
...
+                // if false, add none
+                // if array, add the ones defined
+
+                // order of system columns will always be the same
+                // RID will always be first in the visible columns list
+                // the rest will always be at the end ('RCB', 'RMB', 'RCT', 'RMT')
+
+                var systemColumnsMode = null;
+                // if compact or detailed, check for system column config option
+                if (module._isCompactContext(this._context)) {
```
RFSH: If you're sending the context to chaise, there's no reason to check the context here too. The same as `appLinkFn` send the context to that function and let that function deal with it.

RFSH: I know you've done this to simplify checking for sub-contexts of `compact`, but I feel like it's better to have all of these logic in one place. ermerstjs just needs to know that there's a callback that it can call.

js/reference.js
```diff
@@ -2716,18 +2724,60 @@
...
+                    systemColumnsMode = module._systemColumnsHeuristicsMode("compact");
+                } else if (this._context == module._contexts.DETAILED) {
+                    systemColumnsMode = module._systemColumnsHeuristicsMode("detailed");
+                }
+
+                // if array or true, add RID to the list of columns
+                if (systemColumnsMode) {
+                    if ((typeof systemColumnsMode == "object" && systemColumnsMode.indexOf("RID")) || systemColumnsMode == true) {
```
RFSH: What is this `(typeof systemColumnsMode == "object" && systemColumnsMode.indexOf("RID"))` condition checking for? This translates to "add RID display key if it's not the first element of the array" which I don't think that's what you want.

It's better to change the object check to actual array check (Array.isArray). Can RID be part of `systemColumnsMode`?

js/reference.js
```diff
@@ -2716,18 +2724,60 @@
...
+                        var ridKey = null;
+                        this._table.keys.all().forEach(function (key) {
```
RFSH: Change this to

```javascript
var ridKey = this._table.keys.all().find(function (key) {
  return key.simple && key.colset.columns[0].name === "RID";
});

if (ridKey) {
   this._referenceColumns.push(new KeyPseudoColumn(this, ridKey));
  // should only ever be 1 column for RID key colset
  consideredColumns[ridKey.colset.columns[0].name] = true;
}
```

js/reference.js
```diff
@@ -2797,6 +2850,35 @@
...
+                if (systemColumnsMode) {
+                    // array of column names to add
+                    var columnsToAdd = [];
+                    if (systemColumnsMode == true) {
+                        // add all, includes RID which will be skipped because of consideredColumns
+                        columnsToAdd = module._systemColumns;
+                    } else {
+                        // add the ones defined in array of config property
+                        // preserves order defined in `_systemColumns`
+                        columnsToAdd = module._systemColumns.filter(function (col) {
```
RFSH: This will ignore the ordering of `systemColumnsMode` and will order them based on `module._systemColumns`.

js/reference.js
```diff
@@ -2797,6 +2850,35 @@
...
+                        });
+                    }
+
+                    for (i=0; i < columnsToAdd.length; i++) {
+                        var colName = columnsToAdd[i];
+                        if (colName in consideredColumns) {
+                            continue;
+                        }
+
+                        if (this._table.columns.has(colName)) {
```
RFSH: The `has` function will scan the array (it's calling filter). So it's just unnecessary to call it when you want to use the column object. you can just call the `get` function and then wrap it in try-catch.

```javascript
try {
  var column = this._table.columns.get(colName);
  addColumn(column);
} catch() {}
```

You should also do the `consideredColumns[colName] = true;` after the `if (colName in consideredColumns) `.

test/specs/column/tests/01.columns_list.js
```diff
@@ -757,6 +760,88 @@ exports.execute = function (options) {
...
                 areSameColumnList(entryEditRef.generateColumnsList(), entryEditRef.columns);
             });
+
+            describe('with system columns heuristic config options defined,', function () {
+                function systemColumnsHeuristicsMode(context) {
+                    var mode = null;
+                    if (context == 'compact') {
+                        mode = true;
+                    } else if (context == 'detailed') {
+                        mode = ['RMB', 'RCT'];
```
RFSH: You should test an array that is rearranged. To make sure we're preserving the order.

test/specs/column/tests/01.columns_list.js
```diff
@@ -757,6 +760,88 @@ exports.execute = function (options) {
...
+                    compactBriefInlineSystemColumnsModeColumns, compactSelectSystemColumnsModeColumns;
+
+                var detailedSystemColumnsModeRef, detailedSystemColumnsModeColumns;
+                beforeAll(function (done) {
+                    // set this here so it doesn't affect above columns list tests
+                    options.ermRest.systemColumnsHeuristicsMode(systemColumnsHeuristicsMode);
+                    options.ermRest.resolve(singleEnitityUriSystemColumnsHeuristics, {
+                        cid: "test"
+                    }).then(function (response) {
+                        compactSystemColumnsModeRef = response.contextualize.compact;
```
RFSH: I don't think you need to test all sub-contexts of `compact` here. As I mentioned I think it's chaise's responsibility to consume the contexts correctly.

test/specs/column/tests/01.columns_list.js
RFSH: Also add RID, a non-system column, and an invalid column to this list. To make sure those branching logic that you have in the code are working as expected.

RFSH:
@jrchudy The heuristics that you described is not still specifying what happens if the RID is part of the arrays that the function is returning. What should we do in that case?

Also please update the `generateColumnsList` jsdoc to include this newly added heuristics.

js/reference.js
jrchudy: I wasn't sure of this when I was writing this. I saw that we had boolean checks similar to this in `ermrestJS` already, so mimic'ed that without comparing it to the applink function case.

js/reference.js
jrchudy: The intention of this check was to check if it's an array (array's type is "object") and that RID is present in the array (that check should be `indexOf() != -1`). If that check fails, it verifies if the value is set to `true` instead.

`Array.isArray` is a better way to check this. If `RID` is not in the array of system columns, then it won't be added.

js/reference.js
jrchudy: Hongsuda and I discussed the ordering, and decided ordering won't be honored. If the property is set to a list (array), it will just add the columns defined in the array in a preset order (`RCB, RMB, RCT, RMT`) no matter what. `RID` will always be first, the rest in the above order at the end.

I should have updated the issue and mentioned this in the PR.

jrchudy:
@RFSH  I updated the PR with changes based on your comments. I also updated the issue to include more details about the implementation. Notably that order in the property won't be preserved, and instead it will always be the same order that we have defined in the constants file (also noted the order in the code and jsdoc).

Issue #800, opened 2019-07-11
Change Asset Presentation


RFSH:
Currently if the `filename_column` is not define in the asset annotation and the asset is not from hatrac, only in `compact` context, we're using the last part of url for the caption of the download button. We should change it this way:

- Don't check the context.
- Don't check for `filename_column` being defined or not. As long as its value is null (annotation is not defined or its value in database is null), we should fall back to using the last part url for caption.
- To make sure we're not hiding any data from the users in `detailed` context, we're going to append the origin name beside the button.

RFSH:
The changes for this has been merged with master. Assigning to @hongsudt  for verification.

You can use this link for testing this: https://dev.gudmap.org/chaise/record/#2/Common:Publication/RID=17-EZ7M

PR #801, opened 2019-07-11
modify asset presentation heuristics

RFSH:
This PR will implement the set of changes that are described in #800.

The following is an example of how an asset that is not hosted in hatrac will look like: https://dev.gudmap.org/~ashafaei/chaise/record/#2/Common:Publication/RID=17-EZ7M

To make sure the added span looks good, I had to add [a new css rule to chaise](https://github.com/informatics-isi-edu/chaise/blob/master/common/styles/app.css#L775,L779).


RFSH:
@jrchudy Thanks for running the chaise test cases. I ran only the presentation test cases and forgot that this change also affects the entry modes. I updated the test cases that you mentioned in [this PR](https://github.com/informatics-isi-edu/chaise/pull/1821).

PR #802, opened 2019-07-12
use visible-columns to order export columns

RFSH:
In this PR I changed the default export heuristics to use the `export` context in `visible-columns` annotation instead of the table columns. If the context is not defined, it will fallback to the `detailed` annotation. 

I had to make changes to the test cases. I might have to refactor the export test cases at some point. When I wrote them, I wanted to make sure getting annotations from schema and table works the same and therefore created two identical schemas. That's why any change that I made to one schema I had to make to the other one too. I could refactor this to use a very simple table for testing schema vs. table instead of the main table that is used for testing the heuristics.


RFSH:
It seems like one of the dependencies that we have has been updated (most probably markdown-it or markdown-it-attrs). We’re using `vendor` folder when we’re including ermrestjsin chaise but in node environment we rely on the `package.json`. So I fixed the version of dependencies that we have in `package.json` to be aligned with the files that we’ve included in `vendor` folder.

If we decided to update any of the dependencies in the `vendor` folder, we should make sure to update the `package.json` versions too.

RFSH:
@jrchudy Previously we were not using the `visible-columns` annotation at all for generating the default export template. I was going based on the column definitions on the table. Now that I'm going based on `visible-columns`, I had to modify the code to take care of different types of pseudo-columns.

PR #805, opened 2019-07-23
Check asset origin

jrchudy:
The following changes were made:
 - change `deriva-url-validate` to `asset-permission`
 - conditionally add `asset-permission` to asset pseudo column default display heuristic if same origin
 - apply the same condition to adding `uinit` and `cid` query parameters
 - Added a new option that the http wrapper will check for to skip the retry logic for specific cases (when error code 0 or -1 is returned)

### DIFF
--- a/test/specs/column/tests/02.reference_column.js
+++ b/test/specs/column/tests/02.reference_column.js
@@ -749,11 +749,11 @@ exports.execute = function (options) {
 
                     it ("otherwise, if asset column has filenameColumn and it's value is not empty, should use it for caption.", function () {
                         val = assetRefCompactCols[10].formatPresentation({"col_asset_3": "https://example.com", "col_filename": "filename"}).value;
-                        expect(val).toEqual('<a href="https://example.com?uinit=1&amp;cid=test" download="" class="download asset-permission">filename</a>', "value missmatch.");
+                        expect(val).toEqual('<a href="https://example.com" download="" class="download">filename</a>', "value missmatch.");
 
                         val = assetRefCompactCols[10].formatPresentation({"col_asset_3": "https://example.com?query=1&v=1", "col_filename": "filename"}).value;
                         //NOTE this is the output but it will be displayed correctly.
-                        expect(val).toEqual('<a href="https://example.com?query=1&amp;v=1&amp;uinit=1&amp;cid=test" download="" class="download asset-permission">filename</a>', "couldn't handle having query params in the url.");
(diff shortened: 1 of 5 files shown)


test/specs/column/tests/02.reference_column.js
```diff
@@ -749,16 +749,16 @@ exports.execute = function (options) {
 
                     it ("otherwise, if asset column has filenameColumn and it's value is not empty, should use it for caption.", function () {
                         val = assetRefCompactCols[10].formatPresentation({"col_asset_3": "https://example.com", "col_filename": "filename"}).value;
-                        expect(val).toEqual('<a href="https://example.com?uinit=1&amp;cid=test" download="" class="download deriva-url-validate">filename</a>', "value missmatch.");
+                        expect(val).toEqual('<a href="https://example.com?uinit=1&amp;cid=test" download="" class="download asset-permission">filename</a>', "value missmatch.");
```
RFSH: `asset-permission` shouldn't be in the expected value, right? (it's not the same origin).

test/specs/column/tests/02.reference_column.js
```diff
@@ -749,16 +749,16 @@ exports.execute = function (options) {
...
 
                         val = assetRefCompactCols[10].formatPresentation({"col_asset_3": "https://example.com?query=1&v=1", "col_filename": "filename"}).value;
                         //NOTE this is the output but it will be displayed correctly.
-                        expect(val).toEqual('<a href="https://example.com?query=1&amp;v=1&amp;uinit=1&amp;cid=test" download="" class="download deriva-url-validate">filename</a>', "couldn't handle having query params in the url.");
+                        expect(val).toEqual('<a href="https://example.com?query=1&amp;v=1&amp;uinit=1&amp;cid=test" download="" class="download asset-permission">filename</a>', "couldn't handle having query params in the url.");
```
RFSH: same thing about here

test/specs/column/tests/02.reference_column.js
RFSH: I think the issue here is the input: `https://example.com`. It doesn't have any `/`. I don't think this is really a valid input so you should just change it to `https://example.com/file.png` or something that is valid.

Also you could change all these links to be relative, and then add a special `it` just to test the same-origin logic that you've added. When I wrote these test cases the input didn't matter that much, so you might want to modify them.

jrchudy: Ah yes, I didn't even realize that was happening here.

PR #820, opened 2019-11-07
Modify log headers

jrchudy:
Modified the log headers function to take an object as a parameter instead of a string representing the action. 

I left a NOTE in the code to mark where I was uncertain what should be done to handle missing parameters.

### DIFF
--- a/js/core.js
+++ b/js/core.js
@@ -145,7 +145,6 @@
          */
         this.http = module._wrap_http(module._http);
         this.http.contextHeaderParams = contextHeaderParams;
-        this.http.contextHeaderParams.cid = this.http.contextHeaderParams.cid;
 
         /**
          * context-id: shows the id of app that this server is being used for
--- a/js/core.js
+++ b/js/core.js
@@ -317,13 +317,23 @@
          * @return {Promise} a promise that returns json object or snaptime if resolved or
          *      {@link ERMrest.ERMrestError} if rejected
          */
-        currentSnaptime: function (action) {
+        currentSnaptime: function (header) {
             var self = this, defer = module._q.defer(), headers = {};
-            headers[module.contextHeaderName] = {
-                action: action || "model/snaptime",
-                catalog: self.id
+
+            if (header) {
+                headers[module.contextHeaderName] = header;
+            } else {
+                headers[module.contextHeaderName] = {
+                    action: "model/snaptime",
+                    catalog: self.id
+                };
+            }
+
+            var config = {
+                headers: headers
             };
-            this.server.http.get(this._uri, {headers: headers}).then(function (response) {
+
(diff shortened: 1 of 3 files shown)


js/core.js
```diff
@@ -308,13 +317,23 @@
          * @return {Promise} a promise that returns json object or snaptime if resolved or
          *      {@link ERMrest.ERMrestError} if rejected
          */
-        currentSnaptime: function (action) {
+        currentSnaptime: function (header) {
```
RFSH: Just for consistency with the rest of functions in ermrestjs let's rename the parameter to `contextHeaderParams `. Please update the jsdoc as well.

js/core.js
```diff
@@ -168,15 +168,24 @@
         /**
          * should be used to log information on the server to different log locations
          * @param {Object} headers - the headers to be logged, should include action
-         * @param {String} location - the path for logging (terminal_error || button_action)
+         * @param {String} location - the path for logging (terminal_error || client_action)
          **/
-        this.logHeaders = function (actionHeader, location) {
+        this.logHeaders = function (contextHeaderParams, location) {
```
RFSH: Do we need to pass the `location`?  Currently, the `terminal_error` is specified only in ermrestjs and chaise has no knowledge of that. Why chaise should send the location in this case? Are we going to use multiple locations for our logs?

js/core.js
```diff
@@ -168,15 +168,24 @@
...
             var defer = module._q.defer();
 
+            if (!contextHeaderParams || (contextHeaderParams === Object(contextHeaderParams) && !Array.isArray(contextHeaderParams))) {
```
RFSH: I think throwing error makes sense since the whole point of this function is to log the passed object.

jrchudy:
@RFSH  I addressed your comments and added the functionality to include `elapsed_s` as an additional attribute in the `dcctx` header.

js/core.js
```diff
@@ -166,20 +165,29 @@
...
-         * @param {String} location - the path for logging (terminal_error || button_action)
          **/
-        this.logHeaders = function (actionHeader, location) {
+        this.logClientAction = function (contextHeaderParams) {
             var defer = module._q.defer();
 
+            if (!contextHeaderParams || (contextHeaderParams === Object(contextHeaderParams) && !Array.isArray(contextHeaderParams))) {
+                var error = new module.InvalidInputError("Context header params were not passed");
+                // Errors for client action logging should not force a terminal error
+                defer.resolve(error);
```
RFSH: You need to return the promise here too. If not, it will run the rest of the function as well. Also, it's better if you reject the promise instead of resolve. You could write this in one line like the following:

```javascript
return defer.reject(error), defer.promise;
```

Issue #744, opened 2018-12-05
General search feature improvement


hongsudt:
People sometimes get confused on the expectation of the general search feature (e.g. the search box on top of the result set). For example, on the gene page, they assume that the search should only search on the gene Name and Synonyms. People get confused when there are many false positives. Here are a few things that I think can help alleviate some of these confusion: 
- [ ] Provide advanced search feature that first allow dba to configure the scope of the search within a table. As a starting point, we can focus on columns within the same table. 
- [ ] Change the hint text to be more specific whether it is a column search of table search
  - column search: search <column1>, <column2>
  - table search: search all columns in a table
- [ ] Provide GUI for users to set the scope themselves 
- [ ] Extend the search to go beyond one table (See #353).

hongsudt:
The first 2 bullets should be part of the UX mitigation plan to address the custom filter for GUDMAP gene sesarch.

RFSH:
### How should we define the list of search columns?
 For defining search columns, we can add define a new context called "search" under `visible-columns`.  DBAs should be able to define pseudo-columns here as well. We might want to define which types of pseudo-columns are allowed here. Does it have to be all-outbound paths?

### What should be the displayed as placeholder in search box?
We are going to show displayname of column names in the search input placeholder in the format of "Search col1, col2, and col3". We should show the parsed HTML values of displaynames in the placeholder (i.e. If you have bold text, the displayed placeholder will be bold). It will be the DBAs responsibility to ensure that the displayname value and representation makes sense as a placeholder.

What should we use as placeholder when the `search` context is missing? Should we use the table displayname ("Search within tablename")? If so, the table name can be markdown and DBAs cannot customize it in this case. Or should we use the general static "search all columns" placeholder instead?

RFSH:
We had a discussion about this feature, the following are important point of it:


### Current main search syntax in facets

Currently the main search is stored as part of facets. If you go to a recordset page with `test` in the main search, the following is what ermrestjs will store in the url as a facet:

```json
{
  "and": [
    {"source": "*", "search": ["test"]}
  ]
}
```

This has been done to make it easer for ermrestjs to parse the url. We're treating `*` as special case and we're only looking for the `search` filter in this syntax. It is also consistent with other types of facet sources. They all support having `search` filter in ermrestjs, so we didn't need to add any extra logic of parsing the `search` filter in here.

[...]

PR #822, opened 2019-11-21
Ability to define search columns

RFSH:
This PR will implement the changes discussed in [#744 comment](https://github.com/informatics-isi-edu/ermrestjs/issues/744#issuecomment-545063518), which mainly includes:

- Support `sourcekey` in the facet part of URLs.
- Add support for special `search-box` sourcekey, which will be used to manipulate the search functionality. 

These are the summary of changes that I had to make to support this:

- `Table.searchSourceDefinition` that is used in both parser helper function and `Reference`.
- `Reference.searchColumns` that chaise will use to construct the placeholder.
- Add a mechanism to attach the `catalog` object to `Location` object. Since we want to support `sourcekey`, we have to get some information from the table and therefore needed the catalog.
- Refactor the old `_JSONToErmrestFilter ` function:
  - I didn't like the name so I renamed it to `_renderFacet`. It's better but still not ideal. This function will turn a given facet object into its equivalent ermrest URL.
  - Moved all the helper functions outside and scoped it to `_renderFacetHelpers` so it's more apparent what they are used for. I think I should do the same with the rest of the functions that were the same as this function e.g. `Reference.facetColumns` and `PseudoColumn.getAggregatedValue`.
 - Add support for `sourcekey` to ` _renderFacet` function.
 - Test cases:
   - I had to change all the places that were using old search syntax, to the new one.
   - Parser test cases were not using any table, but for the `sourcekey` support test cases I had to add a schema to this spec. I also had to modify the other test cases based on this.

P.S. The only remaining task from the issue is the ability to go beyond the table (defining sources with foreign key paths in the `search-box`) which requires more discussion and should be implemented later.


js/parser.js
```diff
@@ -153,6 +159,14 @@
         this._catalog = catalogParts[0];
         this._version = catalogParts[1] || null;
 
+        if (catalogObject) {
```
jrchudy: Isn't it the case that the catalog is always used now? Didn't you make changes to each calling site?

js/utils/pseudocolumn_helpers.js
```diff
@@ -215,116 +196,200 @@
...
+     * @param       {string} catalogId the catalog id
+     * @param       {ERMrest.catalog} [catalogObject] the catalog object (could be undefined)
+     * @param       {Object[]} [consNames] the constraint names (could be undefined)
+     * @constructor
+     * @return      {object} An object that will have the following attributes:
+     * - successful: Boolean (true means successful, false means cannot be parsed).
+     * - parsed: String  (if the given string was parsable).
+     * - message: String (if the given string was not parsable)
+     */
+    _renderFacet = function(json, alias, schemaName, tableName, catalogId, catalogObject, consNames) {
```
jrchudy: The diff for this file is a bit goofy. It looks like this is where you changed `_JSONToErmrestFilter` to `_renderFact`. It looks like you also moved functions that were scoped within that function to be outside of it for reuse. I don't see many changes to the actual logic though besides explicitly checking for the `SEARCH_BOX` source key.

js/parser.js
RFSH: Internally in ermrestjs, this function is used without passing the `catalog` object. For example, the `resolve` function is using this to figure out which catalog it should use, so it doesn't have access to the `catalogObject` yet and cannot pass it.

js/utils/pseudocolumn_helpers.js
RFSH: That's right. The only actual change to this function is supporting `sourcekey` which includes the `SEARCH_BOX` support.

PR #826, opened 2019-12-14
batch association reference

jrchudy:
In this PR I added a function to `Ermrest.Reference` to create an association reference for multiple entities given a set of tuples. This is being done to make batch delete easier by generating a robust reference that has a path to the association table linked directly back to the main table rather than using a conjunction of filters to identify the FK relationships to both tables.


test/specs/reference/tests/02.related_reference.js
```diff
@@ -470,6 +470,61 @@ exports.execute = function(options) {
...
+                                return inputArray.reduce(function (res, current, index, array) {
+                                    return res.concat([current, current]);
+                                }, []);
+                            }
+
+                            var tupleArray = doubleArray(doubleArray(batchPage.tuples));
+                            var batchRefs = batchPage.tuples[0].reference.getBatchAssociationRef(tupleArray);
+                            expect(batchRefs.length).toBe(2, "incorrect number of references generated after reducing the url limit");
+                            expect(batchRefs[0].uri.length).toBeLessThanOrEqual(options.ermRest.URL_PATH_LENGTH_LIMIT);
+                            expect(batchRefs[1].uri.length).toBeLessThanOrEqual(options.ermRest.URL_PATH_LENGTH_LIMIT);
```
RFSH: I think it's better if you check for the `.location.compactPath` or maybe call `read` on these two functions to make sure your logic returned the correct url instead of just checking its length.

js/reference.js
```diff
@@ -1857,6 +1857,80 @@
...
+
+                        if (data === undefined || data === null) return null; // tuple data for keyCol doesn't exist
+                        if (j != 0) filter += '&';
+                        filter += module._fixedEncodeURIComponent(keyCol.name) + '=' + module._fixedEncodeURIComponent(data);
+                    }
+
+                    filter += ')';
+
+                    // check url length limit if not first one
+                    if (i != 0 && (currentUrl + filter).length > module.URL_PATH_LENGTH_LIMIT) {
```
RFSH: The limit is just on path, but this is checking the total length of the url. Also, you're using `currentUrl = reference.uri` which could include page and sort modifiers.

 You should change your currentUrl to use the `.location.compactPath` instead, and then change the line that is creating reference objects to always prepend the rest of url.

test/specs/reference/tests/02.related_reference.js
```diff
@@ -470,6 +470,61 @@ exports.execute = function(options) {
...
+                describe('Tuple.getBatchAssociationRef, ', function () {
+                    var batchPage, selectedTuples;
+                    it('should return an array of references including the filtered assocation reference.', function() {
+                        var url = options.url + "/catalog/" + catalog_id + "/entity/";
+                        // there are 3 rows in the set for "pageWithToName"
+                        related[2].sort([{"column":"id", "descending":false}]).read(3).then(function(response) {
+                            batchPage = response;
+                            expect(batchPage.tuples.length).toBe(3, "pageWithToName set is not of size 3");
+
+                            selectedTuples = [batchPage.tuples[0], batchPage.tuples[1]];
```
RFSH: Doesn't `Page.tuples` already return an array for you? Why do you need to do it this way?

test/specs/reference/tests/02.related_reference.js
```diff
@@ -470,6 +470,61 @@ exports.execute = function(options) {
...
+                        });
+                    });
+
+                    it('should return 2 references if the number of tuples to create an association for exceeds the url length limit.', function () {
+                        related[2].sort([{"column":"id", "descending":false}]).read(3).then(function(response) {
+                            var batchPage = response;
+
+                            // modify length limit to make testing this easier
+                            var oldLimit = options.ermRest.URL_PATH_LENGTH_LIMIT;
+                            options.ermRest.URL_PATH_LENGTH_LIMIT = 500;
```
RFSH: I'm not sure how I feel about this. I understand you didn't want to add data to those two tables (the association and leaf table), but this doesn't feel right at all. You don't need to add too many rows. Just make sure the values that you're using are long. The name of the columns are already long too, so that should help. As I mentioned [I did the same](https://github.com/informatics-isi-edu/ermrestjs/blob/master/test/specs/column/conf/pseudo_column_schema/data/inbound%204%20long%20table%20name.json) for pseudo-column aggregate test cases.

test/specs/reference/tests/02.related_reference.js
jrchudy: In this case, batchPage is length 3 but I'm only "selecting" 2 of them for the following test.

test/specs/reference/tests/02.related_reference.js
jrchudy: Won't take long to add lengthy data, was trying to cut a corner.

jrchudy:
Made changes to use the read path to check for `URL_PATH_LENGTH_LIMIT` because it can include foreign key info which will make it longer. Modified data for test cases to test this more easily without doubling the array to avoid defining more data. 

New data broke paging tests, so I commented more about those test cases and how they are meant to work behave. I believe the changes I made to those test cases should be fine to still test the paging stuff properly.

PR #831, opened 2020-02-11
Src template classes

jrchudy:
Added a new class to some of the markdown helpers that we wrote (`image`, `iframe`, `video`). Modified test cases for this as well.

This PR can be reviewed and merged. It will have no effect on chaise as it is only modifying templates to add a new class that isn't used anywhere yet.


jrchudy:
Made the suggested changes and merged the logging changes into my branch.

jrchudy:
Merging this PR since it only adds  class and won't have any negative effects.

Issue #617, opened 2018-03-27
Allow pseudo columns in templating environments


karlcz:
You don't need to follow the conversation below and this comment will summarize the list of tasks that we agreed on. You can refer to [this wiki page](https://github.com/informatics-isi-edu/ermrestjs/wiki/Pseudo-Column-in-Templating-Enviornments) that is explaining this change in depth.

- [x] Allow data-modelers to define a set of source keys for each table.
- [x] Allow `sourcekey` in `visible-columns` and `visible-foreignkeys`.
- [x] Add a new `wait_for` attribute in the `display` settings of pseudo-columns.
  - [x] in visible-columns for `compact` context.
  - [x] in visible-columns for `detailed`context.
  - [x] in visible-foreign-keys annotation.
  - [x] in citation annotation.
- [x] Allow pseudo-columns without any `source` or `sourcekey` (markdown_name and markdown_pattern are required in this case).

<details>
<summary>Old description</summary>
Along with introducing pseudo columns in the visible column list, it would be very helpful to have some new annotation mechanism where a pseudo column definition could be injected into the mustache (or handlebars) template environment (without adding it to the rendered visible columns).

We should think this over a bit before diving in.  One question is whether such a mechanism should be additive to the current environment or instead override the default.  An override behavior could be better for future performance, so we don't have to prepare all kinds of raw and rendered data "just in case" the DBA wants to reference it in a template. Instead, they would have a way to declare their data dependencies and we only need to prepare content actually needed by their templates.

The pseudo column mechanism is also more general than the previous foreign key based content, so that feature would seem largely redundant once this is available...

@hongsudt @robes @ljpearlman @bugacov  @mikedarcy 
</details>

karlcz:
@RFSH @hongsudt @robes 

There is an open PR to add handlebars and support a field like `"template_engine": "handlebars"` as a sibling to our many template pattern fields to switch the engine from mustache to handlebars.  I propose that we introduce a similar field `"template_environment": {...}` with new syntax for the DBA to structure the environment explicitly.

[...]

PR #835, opened 2020-03-03
Pseudo template changes

RFSH:
This PR inlcudes the changes that are required for the [chaise#1913 PR](https://github.com/informatics-isi-edu/chaise/pull/1913). To summarize, this PR:

- adds a new type of `ReferenceColumn` called `VirtualColumn`. The name is generated by the visible-columns logic (`generateColumnsList`) to ensure the uniqueness. It will follow the `$<source.markdown_name>` pattern and if such column name already exists, we will append `-<integer>` to it.

- adds `wait_for` support to `visible-foreign-keys` and `citation`.

- modifies the `Reference.activeList` to support the new logic of ordering the requests. Instead of returning different lists for aggregates and entitysets, this function will return a `requests` array that has all the secondary requests that page needs. This includes aggregates, entity sets, inline entities, and related entities. [This comment](https://github.com/informatics-isi-edu/ermrestjs/issues/617#issuecomment-591602232) explains the new order of sending requests.

- fixes a referential bug in `display` API of columns that would cause one column to return the attribute of another column.
 
- removes the `formattedValues` that was passed as an option to `formatPresentation` functions, and takes the `templateVariables` from the options argument, to be its own argument.

- adds new test cases.

related issue: #617 

_The PR is ready to be reviewed and the only remaining task for this PR is documentation_


js/reference.js
```diff
@@ -3301,52 +3361,66 @@
...
-            //fkeys
-            sds.fkeys.forEach(function (fk) {
-                if (fk.name in consideredOutbounds) return;
-                consideredOutbounds[fk.name] = true;
-                allOutBounds.push(new ForeignKeyPseudoColumn(self, fk));
+            // columns with aggregate
+            columns.forEach(function (col,i ) {
+                if (hasAggregate(col)) {
+                    addInlineColumn(col, i);
+                }
```
jrchudy: Looking at these 2 loops, I'm not sure what's different except the condition checking the opposite in each one.

js/reference.js
```diff
@@ -5165,6 +5239,10 @@
                 // TODO should we move these to the _getFormattedKeyValues?
                 // the multi-fk all-outbounds
                 var sm = this._pageRef.table.sourceDefinitions.sourceMapping;
+
+                // TODO the problem here is that activeList might have been called before, so we don't have the allOutBounds
+                // Currently the wait_for is under the tuple and therefore needs the tuple...
+                // TODO HEEEEEREEEEE!!!!!
```
jrchudy: I think you left a TODO in here

js/column.js
```diff
@@ -330,20 +330,23 @@ ReferenceColumn.prototype = {
...
-                    res.sourceMarkdownPattern = displ.markdown_pattern;
-                    res.sourceTemplateEngine = displ.template_engine;
+                    sourceDisplay.sourceMarkdownPattern = displ.markdown_pattern;
+                    sourceDisplay.sourceTemplateEngine = displ.template_engine;
                 }
             }
 
+            // I'm using assign to avoid changing the original colDisplay
+            // it was causing issues and some columns were using other column's display properties
+            Object.assign(res, colDisplay, sourceDisplay);
```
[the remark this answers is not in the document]
RFSH: Correct. Instead of changing the original `colDisplay` and `sourceDisplay` objects, it will copy all the attributes of these two objects into the res (by value and not by reference).

js/reference.js
RFSH: And this is resolved too by moving the `Citation` object (and therefore it's `waitFor` definition) to the `Reference` API.

js/reference.js
RFSH: Yup. that's the only difference. The first one will process all the columns that are not aggregate or don't have aggregate in their `wait_for` definition. The other function will take care of the rest of the columns.

PR #836, opened 2020-03-05
Version info for vendor assets

jrchudy:
Modified build command to create another file with a version variable that will be appended to the `ermrest.js` build file that is generated.

Removed some dependencies from `make test` and `make testsingle` commands so that developers are required to call `make all` or `make install` first.

I also re-ordered `travis.yml` to be more consistent with the layout of the one in chaise.


jrchudy:
@RFSH  this is ready for review. I have looked at it so many times that I'm unsure if I missed something and would another pair of eyes to make sure everything we discussed has been implemented.

Makefile
```diff
@@ -69,16 +71,22 @@ JSDOC=jsdoc
 LINT=.make-lint
 TEST=.make-test.js
 
+.PHONY: pre-generate-files-for-build
+pre-generate-files-for-build:
+	mkdir -p $(BUILD)
+	# create the version variable and use the current date + time for versioning
+	echo 'var version=$(shell date +%Y%m%d%H%M%S);' > $(BUILD)/makefile_variables.js
```
RFSH: Since this variable will be available in all the ermrestjs files, just to make sure it's not going to collide with any other variables, let's use a more specific name for it instead of just `version`. You could call it `buildTime` (`ermrestjsBuilTime`?) or `buildVersion` (`ermrestjsBuildVersion`?) or something more specific even.

PR #850, opened 2020-07-28
Contextualized comment and comment display

jrchudy:
In this PR, I made changes to allow for data modelers to add `comment` and `commentDisplay` to a few different annotations. Along with these additions, heuristics were introduced for picking the comment from source syntax, foreign key annotation, table display annotation, or the model itself. Please see the [attached issue](https://github.com/informatics-isi-edu/chaise/issues/1842) to see the discussion about this.

#### columns.js
 - modified `comment` and added `commentDisplay` to `PseudoColumn`
 - added `_comment` and `_commentDisplay` to `PseudoColumn.reference`

#### core.js
 - added `getDisplay()` to `Table`
   - tests added to `01.comment.js`
 - modified `getDisplay()` for `ForeignKeyRef`
   - should tests be added for this?

#### reference.js
 - added `comment` and `commentDisplay` to `Reference`
 - added `commentDisplay` to `Reference`
 - modified `_generateRelatedReference` to attach `_comment` and `_commentDisplay` to the new reference

#### tests
 - spec `18` tests `comment` and `commentDisplay` for related tables defined in visible-columns
 - spec `19` tests `comment` and `commentDisplay` for related tables defined in visible-foreign-keys

### Comments:
I duplicated the schema for these specs and only changed the name between the 2 schema documents so the relationships should be the same. In spec 18, I am testing `column.reference.comment` and in spec 19 I'm testing `related.comment`. Should that be the case in spec 18 or should I be testing `column.comment`?
   - NOTE: after doing the tests in chaise, I noticed a good number of tests for record presentation rely on `column.comment`, I think we need to discuss this further

I did some more messing around in my tests and found that calling `column.comment` would trigger the logic to get comment from `InboundForeignKeyPseudoColumn`. I didn't make change to that `class` since that didn't seem right to me. But I'm not sure so should `InboundForeignKeyPseudoColumn` be modified to also include `comment` and `commentDisplay`?


jrchudy:
I added tests to test the changes I made to the foreign key annotation. 

Also modified the schema to remove the annotations from foreign key definitions and table definitions so if a heuristic is wrong, the value will be wrong rather then using another heuristic.

js/reference.js
```diff
@@ -3569,6 +3611,23 @@
                     newRef._displayname = otherFK.colset.columns[0].table.displayname;
                 }
 
+                // NOTE: this is for contextualized toComment and toCommentDisplay, to be implemented later
+                // display = otherFK.getDisplay(this._context);
+
+                // comment
+                if (otherFK.to_comment) {
```
RFSH, line 3618: In this case, `to_comment` is already processed, so we don't need to do `_isValidModelComment`. But you should still do string type check.  The same comment is applicable to `from_comment` case below (an empty string is a valid value which is not handled here.)

Also in all the codes below that you're setting the value of `_commentDisplay`, you don't need any extra check and you should just use the value. With the current implementation (with the extra check), the `comment` might be coming from the source object, while the `commentDisplay` is derived from the `to_comment_display `.

js/core.js
```diff
@@ -3777,6 +3835,18 @@
                 if(jsonAnnotation.to_name){
                     this.to_name = jsonAnnotation.to_name;
                 }
+
+                if(jsonAnnotation.to_comment){
+                    // check for null, false, empty string when digesting comment for first time
+                    this.to_comment = _processModelComment(jsonAnnotation.to_comment);
+                    if (jsonAnnotation.to_comment_display) this.to_comment_display = jsonAnnotation.to_comment_display;
```
RFSH, line 3842: I think you should set a default value for `to_comment_display` and `from_comment_display`.

js/column.js
```diff
@@ -270,6 +270,18 @@ ReferenceColumn.prototype = {
         return this._comment;
     },
 
+    get commentDisplay() {
+        if (this._comment === undefined) {
+            if (this.sourceObject && this.sourceObject.comment && this.sourceObject.comment_display) {
+                // only change commentDisplay if comment and comment_display are both defined
+                this._commentDisplay = this.sourceObject.comment_display;
```
RFSH, line 277: This comment is applicable to all the different places that you're populating the `commentDisplay` API value from the annotation. If you're not doing any extra check for "valid" string values (`tooltip`, `inline`) in ermrestjs, then in your chaise code you have to make sure you're properly handling this.  I think an invalid value, should be treated as if the annotation is missing and therefore should be treated the same as `tooltip`.

test/specs/reference/tests/18.comment_display_inline.js
```diff
@@ -0,0 +1,189 @@
+exports.execute = function (options) {
```
RFSH, line 1: Can you please add a comment before this line, to explain what are the different cases that you're testing? (for both of the specs that you've added)

You can summarize the tables that you have and what are you testing for each of the columns. I tried to do on the specs that are testing a lot of different scenarios, for example [this one](https://github.com/informatics-isi-edu/ermrestjs/blob/master/test/specs/column/tests/03.pseudo_column.js#L1). You don't need to explain everything, just some high-level details should be enough.

PR #851, opened 2020-07-30
Print iframe fix

DamCousik:
This PR includes the following fixes

- [x] Fixed the print for iframes with Treeview document to show the tree hierarchy during print.
- [x] Persisted scrolling during print
- [x] Parsing the URL to check if there is an embedded YouTube video. If yes, then add a class to apply the above discussed custom styles.
- [x] The same has been processed for the <video> markdown pattern as well.

- Applying the thumbnail to the YouTube video on print was complicated as the styles are applied via YouTube and we have limited access to it. Hence, the alternative was to display a note informing the user that the video is hidden during print and possibly provide the URL of the hidden video.


js/utils/helpers.js
```diff
@@ -1850,11 +1850,22 @@
...
 
                             // Add all attributes to the iframe
                             openingLink.attrs.forEach(function(attr) {
                                 if (attr[0] == "href") {
+                                    isYTlink  = (attr[1].indexOf("www.youtube.com") != -1 ? true : false);
+                                    
+                                    // Add a class to hide the iframe only if the markdown is a YouTube video
+                                    if(isYTlink)
+                                    {
+                                      iframeHTML += 'class="hide-in-print"' + " ";
```
RFSH, line 1863: As I mentioned this will cause double `class` attributes on the element. So you need to modify it. 

If you define an iframe with classes this way:
```
::: iframe [CAPTION](https://example.com){link=https://example.com .test .another-class } \n:::
```

One of the attributes that you get in this forEach loop is going to have `attr[0] === "class"`. So you should merge the class that you're adding here and that class together.  

You can see how we're doing this in the video tag. In there we're always attaching a class, so we had to do something similar (Take a look at the `videoClass` attribute in there.)

RFSH: You should change your code to cover more patterns of youtube links. It's also cleaner if in here you just set the `isTYlink` and capture the link value. And move the part that you're defining your "videoText" to be in the same `if (isYTLink)` block that you have below.

So

```javascript
var isYTlink = false, videoURL = "";
openingLink.attrs.forEach(function(attr) {
  if (attr[0] == "href") {
    isYTlink  = (attr[1].match("^(http(s)?:\/\/)?((w){3}.)?youtu(be|.be)?(\.com)?\/.+") != null);
    videURL = attr[1];
  }
  ...
}

...

if (isYTlink) {
  html =  '<span class="video-info-in-print" style="visibility:hidden">';
  htm += 'Note: YouTube video ( ' + attr[1] + ' ) is hidden in print';
  htm += '</span>' + html;
}



```

DamCousik: - Made all the changes that were requested. All test cases are passing except for this one: 

`Testing for tuple 0 with row values {id=4000, some_markdown= **date is :**, name=Hank, url= https://www.google.com, some_gene_sequence= GATCGATCGCGTATT, video_col= http://techslides.com/demos/sample-videos/small.mp4, catalog_snapshot_uri=schema:table, catalog_id_uri=schema:table}`

 ```
✗ should check value and isHTML for 'video_col' column
 - Expected '<span class="video-info-in-print" style="visibility:hidden">Note: Video (http://techslides.com/demos/sample-videos/small.mp4) is hidden in print </span><video controls class="-chaise-post-load hide-in-print " height=500 width=600 loop ><source src="http://techslides.com/demos/sample-videos/small.mp4" type="video/mp4"></video>' to be '<video controls class="-chaise-post-load" height=500 width=600 loop ><source src="http://techslides.com/demos/sample-videos/small.mp4" type="video/mp4"></video>'.
```

I checked in print_utils and I have changed all the expected behavior of the video test cases. @RFSH / @jrchudy could you please take a look at this for me? Once this test case passes, we are good to go. Everything else has passed.

File: js/utils/helpers.js [hunk 2]
```diff
@@ -2180,11 +2197,11 @@
                             }
 
                             if(captionHTML.trim().length && flag && posTop){
-                                html +=  "<figure><figcaption>"+captionHTML+ "</figcaption>" + videoHTML + videoClass +">"+ srcHTML +"</video></figure>" ;
+                                html +=  "<figure><figcaption>"+captionHTML+ "</figcaption>" + infoHTML + videoHTML + videoClass +">"+ srcHTML +"</video></figure>" ;
```
RFSH, line 2200: In the iframe implementation, you're always attaching this at the beginning of the computed HTML (regardless of position attribute). Why are you not doing the same for video?

js/utils/helpers.js
```diff
@@ -1867,12 +1870,22 @@
...
+                                    newClass += " " + attr[1];
                                 } else {
                                     iframeHTML +=  attr[0] + '="' + attr[1] + '"';
                                 }
                                 iframeHTML += " ";
                             });
-                            html += iframeHTML + "></iframe>";
+
+                            //During print we need to display that the iframe with YouTube video is replaced with a note
+                            if(isYTlink)
```
RFSH, line 1882: As I mentioned in one of my comments, you can define extra classes for iframe, so computed iframe can have extra class. Therefore the `newClass` attribute that you added is something that you always have to add to your `iframeHTML`, so

```javascript
html += iframeHTML + newClass + "></iframe>";
if(isYTlink) {
  html = '<span class="video-info-in-print" style="visibility:hidden">' + videoText + "</span>" + html;
}
```

You should modify your existing youtube test case to include classes as well. So something like this

```
::: iframe [SOME LINK CAPTION](https://www.youtube.com/embed/op1-Cw_l1Ow){width=640 height=480 link=https://www.youtube.com/embed/op1-Cw_l1Ow .some-class .another-class} \n:::
```

Notice the two classes that I've added at the end.

js/utils/helpers.js
```diff
@@ -1850,11 +1850,14 @@
                         if (attrs[0].children[0].type == "link_open") {
                             var iframeHTML = "<iframe ", openingLink = attrs[0].children[0];
                             var enlargeLink, posTop = true, captionClass = "", captionStyle = "", iframeClass = "", iframeStyle = "";
+                            var isYTlink = false, newClass='class="' + "hide-in-print" + '"', videoURL = "";
```
RFSH, line 1853: The last time that I looked at your code I forgot to mention this.

 In ermrestjs, we define all the constants in `js/utils/constant.js` file.  If you search for `module._classNames` we actually have a list of classes as well. These two classes that you're adding should be defined there. 

Notice how in the `video` tag implementation we're using the `module._classNames.postLoad`.

js/utils/helpers.js
RFSH: Also, try to follow the code styles that you see on the file. So the curly braces that you're adding after `if` should be on the same line (just like the rest of the code). 

And avoid defining `if` or `else` blocks without curly braces as well. Even if it's one line, you should add curly braces.

js/utils/helpers.js
```diff
@@ -2134,8 +2147,10 @@
                     if (attrs && attrs.length == 1 && attrs[0].children) {
                         // Check If the markdown is a link
                         if (attrs[0].children[0].type == "link_open") {
-                            var videoHTML="<video controls ", openingLink = attrs[0].children[0];
-                            var srcHTML="", videoClass='class="' + module._classNames.postLoad, videoAttrs="", flag = true, posTop = true;
+                            var videoHTML="<video controls ", videoClass='class="' + module._classNames.postLoad + " " + "hide-in-print" + " ", openingLink = attrs[0].children[0];
```
RFSH, line 2150: You don't need the extra space after `hide-in-print`. You should also change the expected values that you changed (and remove the space).

test/specs/print_utils/tests/01.print_utils.js
```diff
@@ -162,23 +167,28 @@ exports.execute = function (options) {
...
+                expect(printMarkdown(videoMarkDown)).toBe(videoHTML);
 
                 //Check for proper rendering of video tag with some invalid attributes
                 var videoMarkDown = '::: video [caption](http://techslides.com/demos/sample-videos/small.mp4){loop=5 width=800} \n:::';
-                var videoHTML = '<figure><figcaption>caption</figcaption><video controls class="-chaise-post-load" width=800 ><source src="http://techslides.com/demos/sample-videos/small.mp4" type="video/mp4"></video></figure>';
-                expect(printMarkdown(videoMarkDown)).toBe(videoHTML, "The video tag is not rendered properly with invalid attributes ");
+                var videoHTML = '<figure><figcaption>caption</figcaption><span class="video-info-in-print" style="visibility:hidden">Note: Video (http://techslides.com/demos/sample-videos/small.mp4) is hidden in print </span><video controls class="-chaise-post-load hide-in-print " width=800 ><source src="http://techslides.com/demos/sample-videos/small.mp4" type="video/mp4"></video></figure>';
+                expect(printMarkdown(videoMarkDown)).toBe(videoHTML);
+
+                //Hide video markdown during print
```
RFSH, line 188: You didn't need to add any extra test cases specific to your change (as other tests are essentially the same.) But since you added a test case and it seems like the class test is missing from this, you should modify this to have some classes. So your `videoMarkDown` should be 

```javascript
"::: video [caption](http://techslides.com/demos/sample-videos/small.mp4){width=400 .some-class .another-class} \n:::"
```

This will ensure your logic of adding class doesn't interfere with the classes that might have already been defined on the video.

js/utils/helpers.js
DamCousik: That is because, `<iframe>` implementation has html written this way : 
// Encapsulate the iframe inside a figure tag
`html = '<figure class="embed-block ' + module._classNames.postLoad + (iframeClass.length ? (" "  + iframeClass): "") + '" style="' + (iframeStyle.length ? (" "  + iframeStyle ) : "") + '">' + html + "</figure>";`

So irrespective of the position, the `<span>` element is always inside the `<figure>`. However for the <video> this is not the case. Adding the `<span>` irrespective of position might cause the `<span>` to be defined outside the <figure>. That is why I had defined it that way.

js/utils/helpers.js
DamCousik: The default behavior of the iframe is that, they shouldn't be hidden during print. Only when the user wants to, they can define the class - "hide-in-print" and make a decision to hide it. However, iframes containing YouTube videos must be hidden by default. Since I am using the same class to hide these, I am choosing to define them only if i`sYTLink` is true. 

Earlier, this code for adding the new classes was not present. So I guess, with my new code it will take care of that. I have changed this to suit the condition mentioned.

js/utils/helpers.js
```diff
@@ -1871,21 +1871,21 @@
...
-                              html = '<span class="video-info-in-print" style="visibility:hidden">' + videoText + "</span>" + html;
+                            if(isYTlink){
+                              html = '<span class="' + module._classNames.showInPrintMode + '" style="visibility:hidden">' + videoText + "</span>";
+                              newClass += module._classNames.hideInPrintMode;
                             }
-                            else
-                              html += iframeHTML + "></iframe>";
+                            // add closing quote
+                            newClass += '"';
+                            html += iframeHTML + newClass + "></iframe>";
```
DamCousik, line 1888: I have added the class attribute to the iframe as requested. I would like to point out that the class attribute was not added before to the iframe element and hence this change had to be incorporated for all iframes in the test case as well.

DamCousik:
@RFSH / @jrchudy , it took me quite sometime to figure out but I finally understood why the test cases were failing. So from what I observed, add a .className to the attribute list in the mark down results in created a double class name in the result of the markdown. 

E.g. ` '::: video [caption](http://techslides.com/demos/sample-videos/small.mp4){width=800 height=200 loop muted .someClass} \n:::';`

Resulted in : <figure><figcaption>caption</figcaption><span class="video-info-in-print" style="visibility:hidden">Note: Video (http://techslides.com/demos/sample-videos/small.mp4) is hidden in print </span><video controls class="-chaise-post-load **_hide-in-print hide-in-print_**" width=800 height=200 loop muted ><source src="http://techslides.com/demos/sample-videos/small.mp4" type="video/mp4"></video></figure>'

This is because in the code where I have defined how the video markdown should be interpreted and rendered, I have added the "hide-in-print" class by default to it. 

helper.js
`videoClass='class="' + module._classNames.postLoad + " " + module._classNames.hideInPrintMode`. 
          `expect(<CALL THE API THAT YOU WANT TO TEST>).toBe(<EXPECTED VALUE>, <ERROR MESSAGE>);`
Thus for the above line, the <CALL THE API THAT YOU WANT TO TEST> is interpreted and rendered according to the code in helper.js file. It then checks if that output matches with the <EXPECTED VALUE> in the test case. 

I also noticed that in case there is no class passed to the iframe, there still happens to be an extra space for it which is failing in the test cases. I tried to re-write the code but there were way too many changes so I decided to just go ahead and add the spaces. Now all test cases for the video and iframe markdowns have passed.

```
Executed 53 of 53 specs SUCCESS in 6 secs.
Test suite passed
```

However, on running make test, one test case failed. This refers to the iframe markdown defined in some other file which I was unable to find out. If you could tell me how I can go about finding it, that would be great. Please review my changes and let me know if there are is an issue.

DamCousik:
@RFSH / @jrchudy : I also tried to check if I could bypass the browser settings to always display the background images but I was unable to. I tried the following options as suggested in this [post](https://stackoverflow.com/questions/6670151/how-can-i-force-browsers-to-print-background-images-in-css) but had no luck!

1) Added the changes to `<body>`

```
body {
    content:url(32px.png) !important;
    -webkit-print-color-adjust: exact !important;   /* Chrome, Safari */
    color-adjust: exact !important;                 /*Firefox*/
  }
```

2) Added it to the universal selector *

```
/*To force the browser to print background images and colors*/
  * {
    -webkit-print-color-adjust: exact !important;   /* Chrome, Safari */
    color-adjust: exact !important;                 /*Firefox*/
  }
```

Since nothing worked, I haven't committed the changes. I also ready a couple of articles and in most of them, the discussion say we have very little control over the browser's printing capabilities. However, I found this [post](https://rewordify.com/BackgroundColors.htm) which explains how we can turn on the settings in all major browsers to have the background images printed.

RFSH:
@DamCousik You should be able to merge all of your PRs. After you're done with the merge, make sure it's working properly on dev.facebase and dev.rbk (you need to wait for hourly update). When you verified that it your changes were working properly, announce the change in the issues that are related to the changes.

DamCousik:
Hi Aref, regarding your comment. I did understand why the old ones were failing. It is because of the newClass I added. I did so because you said it should be applied to all. The reason I did not correct you is because we have scenarios where we are expected to add class to the iframes in those cases it will not add since while rendering the iframe, the class element is missing. That’s why I went ahead and made those changes. 

So just to point it out again, there is no "class" element attached to the iframe markdown now apart from the ones having a youtube video. So if a class needs to be defined , we may have to remake those changes.

PR #859, opened 2020-09-12
Add regexp String return

jrchudy:
Added a new handlebars helper for regular expression matching that returns the substrings that match the given regular expression.


RFSH:
@jrchudy  I think we should come up with a better name for this. `regexSearch` or `regexExtract` might be better options for naming this.

And we should discuss whether this should match the first one (and return it), or return all the matches as it currently does. In most cases, I feel like you just want the first match. If we force an array output, then they have to use handlebars and use the non-trivial syntax of arrays.

Let's just discuss this in one of our meetings.

jrchudy:
@RFSH I cleaned up the code to properly test `regexFindAll` which returns an array of matches. I added another helper named `regexFindFirst` which will return the first matching result. I think this addresses the concern from your comment above.

RFSH:
You should add these two new functions to [our handlebars documentation](https://github.com/informatics-isi-edu/ermrestjs/blob/master/docs/user-docs/handlebars.md). And if you can find a proper regular expression that would extract the filename and add it in the documentation that would be great. I think that was the first needed use case for this.

RFSH (approved):

PR #862, opened 2020-10-27
COMPACT_ENTRY context

jrchudy:
New context added to the reference API in ermrestJS denoted as compact/entry. This context has a special case for removing columns that generate additional requests for more information.

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -2844,7 +2844,7 @@
 
             var context = this._context;
             isEntry = module._isEntryContext(context);
-            isCompactEntry = context.startsWith(module._contexts.COMPACT_ENTRY);
+            isCompactEntry = (typeof context === "string" && context.startsWith(module._contexts.COMPACT_ENTRY));
 
             // check if we should hide some columns or not.
             // NOTE: if the reference is actually an inbound related reference, we should hide the foreign key that created this link.
--- a/js/reference.js
+++ b/js/reference.js
@@ -1247,8 +1247,9 @@
                     }
 
                     var ref = new Reference(module.parse(uri), self._table.schema.catalog);
+                    ref = (response.data.length > 1 ? ref.contextualize.compactEntry : ref.contextualize.compact);
                     //  make a page of tuples of the results (unless error)
-                    var page = new Page(ref.contextualize.compactEntry, etag, response.data, false, false);
+                    var page = new Page(ref, etag, response.data, false, false);
 
                     //  resolve the promise, passing back the page
                     return defer.resolve({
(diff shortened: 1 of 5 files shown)


js/reference.js
```diff
@@ -3062,7 +3062,13 @@
 
                             // in entry mode, pseudo-column, inbound fk, and key are not allowed
                             if (!(isEntry && (refCol.isPathColumn || refCol.isInboundForeignKey || refCol.isKey) )) {
-                                this._referenceColumns.push(refCol);
+                                // NOTE: should I check refCol.hasAggregate ?
+                                // check we are in entry/compact mode before excluding more columns
+                                if (context.startsWith(module._contexts.COMPACT_ENTRY) && (refCol.hasWaitFor || !refCol.isUnique || (refCol.hasPath && refCol.isUnique && refCol.foreignKeys.length > 1)) ) {
```
RFSH, line 3067: To make it a little bit "cleaner", you could do something like

```javascript
ignore = !(isEntry && (refCol.isPathColumn || refCol.isInboundForeignKey || refCol.isKey) ) ||
         NEW_CONDITION;
if (!ignore) {
  this._referenceColumns.push(refCol);
}
```

And as for the new condition, you can make it simpler. What I wrote originally was a literal translation of what we want but I understand that it's not easy to read and it's just too long.

- Technically since it's non-detailed, you could just check for `hasAggregate`.
- We don't need to check all that three conditions and just `refCol.isPathColumn && refCol.foreignKeys.length > 1` should be enough.

Also, this function works with "uncontextualized" references and let's just make sure it still continues to do so

So 
```javascript
// this should be added at the beginning of this function where we're initializing all these type of variabels
var isCompactEntry = typeof context === "string" && context.startsWith(module._contexts.COMPACT_ENTRY);

...


ignore = (isEntry && (refCol.isPathColumn || refCol.isInboundForeignKey || refCol.isKey) ) ||
         (isCompactEntry && (refCol.hasWaitFor || refCol.hasAggregate || (refCol.isPathColumn && refCol.foreignKeys.length > 1));
```

RFSH:
And as for testing, you don't need to change any schema definition. You can add a new `it` after this one:

https://github.com/informatics-isi-edu/ermrestjs/blob/master/test/specs/column/tests/03.pseudo_column.js#L249

And test the columns list based on the existing definition.

RFSH (approved):

js/reference.js
```diff
@@ -2838,11 +2838,13 @@
                 pseudoNameObj, pseudoName, isHash,
                 hasInbound, isEntity, hasPath, isEntityMode,
                 isEntry,
+                isCompactEntry,
                 colFks,
                 ignore, cols, col, fk, i, j;
 
             var context = this._context;
             isEntry = module._isEntryContext(context);
+            isCompactEntry = context.startsWith(module._contexts.COMPACT_ENTRY);
```
RFSH, line 2847: You should add the `typeof context === "string"` before this. 

```javascript
var isCompactEntry = typeof context === "string" && context.startsWith(module._contexts.COMPACT_ENTRY);
```

As I mentioned, we support contextualized References and that's why some test cases are failing.

PR #865, opened 2021-01-28
Responsive iframe documentation

jrchudy:
This PR mostly changes the documentation for iframes. Changes to the render function itself were minimal and mostly refactoring.


RFSH (commented):
These are a few things that came to my mind about the documentation:

1. You should add `chaise-auto-fill` to the list of special classes that are defined [here](https://github.com/informatics-isi-edu/ermrestjs/blob/master/docs/user-docs/markdown-formatting.md#special-classes).
2. I think the "a.5" example should be changed so it only has width, hight, style and class (remove the link)
3. You should separate the Responsiveness and "other cases".  Based on what I'm seeing, the other cases are just "Styling the figure", right? In that case I think they could just be part of "Styling the iframe".
4. I feel like there's a disconnect between the explanations that you've added and examples. The existing doc was trying to explain everything by example. So if someone's reading this doc, most probably they will miss the explanations that you added and will just directly go to the examples. I'm not sure what's the best way to solve this or if it even matters. Even if someone reads the explanations, there's no linkage to the examples that you have after it, so there's a disconnect here. Maybe move the paragraph explaining styles to the "a.Styling the iframe" and the other paragraph to the "c. Responsiveness and Other cases"?  As I said I'm not entirely sure what works.

PR #869, opened 2021-02-11
New tab for iframe fullscreen button

jrchudy:
Changed the functionality of the fullscreen button to open the iframe content in the same tab when clicked. Added the ability to have this open in a new tab by setting `fullscreen-target=_blank`.

This same default is going to be used for the caption going forward as well when a link is defined for it. The caption can be set to open in a new tab by using `caption-target=_blank`.

Other changes include cleaning up the templating further to remove unnecessary spaces and only add the style property when styles exist for that element.


RFSH (approved):
Changes look good. While announcing this change we should also make sure to mention the following:

- Usage of `link` is deprecated and they should change it to `caption-link`.
- Default behavior of caption link and fullscreen links have been changed and if they want to use tab, they have to follow the new instructions.

PR #873, opened 2021-03-18
Fix for max-width style with fullscreen button

jrchudy:
This change parses out the min/max width styles to apply to the caption and fullscreen button wrapper (container).

### DIFF
--- a/js/utils/helpers.js
+++ b/js/utils/helpers.js
@@ -1942,10 +1942,10 @@
                                         // min/max width needs to be applied to the wrapper of the caption and fullscreen button for consistent button placement
                                         // check for min-width style
                                         var minWidthIdx = attr[1].indexOf("min-width");
-                                        if (minWidthIdx >= 0) {
+                                        if (minWidthIdx != -1) {
                                             endStyleIdx = attr[1].indexOf(";", minWidthIdx);
                                             // get the min-width `key: value` pair
-                                            if (endStyleIdx) {
+                                            if (endStyleIdx != -1) {
                                                 subStrStyle = attr[1].substring(minWidthIdx, endStyleIdx);
                                             } else {
                                                 // if no `;` after min-width, assume end of string
@@ -1956,11 +1956,10 @@
 
                                         // check for max-width style
                                         var maxWidthIdx = attr[1].indexOf("max-width");
-
-                                        if (maxWidthIdx >= 0) {
+                                        if (maxWidthIdx != -1) {
                                             endStyleIdx = attr[1].indexOf(";", maxWidthIdx);
                                             // get the max-width `key: value` pair
-                                            if (endStyleIdx) {
+                                            if (endStyleIdx != -1) {
                                                 subStrStyle = attr[1].substring(maxWidthIdx, endStyleIdx);
                                             } else {
                                                 // if no `;` after max-width, assume end of string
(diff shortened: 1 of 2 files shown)


js/utils/helpers.js
```diff
@@ -1938,6 +1938,19 @@
                                         }
 
                                         // handles `style="some: style;"` case from template
+                                        // min/max width needs to be applied to the wrapper of the caption and fullscreen button for consistent button placement
+                                        // check for min-width style
+                                        var minWidthIdx = attr[1].indexOf("min-width");
+                                        if (minWidthIdx >= 0) {
+                                            widthStyles += attr[1].substring(minWidthIdx, attr[1].indexOf(";", minWidthIdx)+1);
```
RFSH, line 1945: If `min-width` is the last style defined and therefore `;` is missing, this will not work properly (`attr[1].indexOf(";", minWidthIdx)` returns -1 in this case). You would have to check that. 

Also I think it might be better to avoid one liners for stuff like this. It's hard to follow what you're doing.

js/utils/helpers.js
```diff
@@ -1989,7 +2002,8 @@
                             // Checks for a width being defined. If it's defined and not a number, assume it has `px` or `%` appended already and use as is.
                             // If width is defined and is a number, assume it's in pixels and append `px`.
                             // If no width, use "100%"
-                            var contentsWidth = 'style="width: ' + (frameWidth ? (frameWidth + (isNaN(parseInt(frameWidth)) ? "" : "px")) : "100%")+ ';"';
+                            // if min/max width are defined for the iframe, apply here as well
+                            var contentsWidth = 'style="width: ' + (frameWidth ? (frameWidth + (isNaN(parseInt(frameWidth)) ? "" : "px")) : "100%") + ';' + (widthStyles ? " " + widthStyles : "" ) + '"';
```
RFSH, line 2006: I think kit's better if you break this one into multiple lines as well

js/utils/helpers.js
```diff
@@ -1938,6 +1938,19 @@
...
+                                        // check for min-width style
+                                        var minWidthIdx = attr[1].indexOf("min-width");
+                                        if (minWidthIdx >= 0) {
+                                            widthStyles += attr[1].substring(minWidthIdx, attr[1].indexOf(";", minWidthIdx)+1);
+                                        }
+
+                                        // check for max-width style
+                                        var maxWidthIdx = attr[1].indexOf("max-width");
+                                        if (maxWidthIdx >= 0) {
+                                            widthStyles += (minWidthIdx >= 0 ? " " : "") + attr[1].substring(maxWidthIdx, attr[1].indexOf(";", maxWidthIdx)+1);
```
RFSH, line 1951: Instead of `(minWidthIdx >= 0 ? " " : "")` I think you should do `(widthStyles.length > 0 ? " " : "")`.

Although I think it might be better if `widthStyles` is an array, then in the end you would check for its length, and also would do `widthStyles.join(" ")` if you really want space between your rules.

P.S. I don't think space between rules are actually necessary. So I would just get rid of the check altogether and add the rules after each other without any space in between.

js/utils/helpers.js
```diff
@@ -1937,7 +1937,38 @@
...
 
+                                        var endStyleIdx, subStrStyle;
                                         // handles `style="some: style;"` case from template
+                                        // min/max width needs to be applied to the wrapper of the caption and fullscreen button for consistent button placement
+                                        // check for min-width style
+                                        var minWidthIdx = attr[1].indexOf("min-width");
+                                        if (minWidthIdx >= 0) {
+                                            endStyleIdx = attr[1].indexOf(";", minWidthIdx);
+                                            // get the min-width `key: value` pair
+                                            if (endStyleIdx) {
```
RFSH, line 1948: This should be `endStyleIdx != -1`

RFSH:
I think you should also add some test cases to `01.print_utils.js` related to this. To make sure the width/min-width/etc are properly defined on the button as well.

RFSH (approved):

PR #882, opened 2021-07-09
1801 chaise google dataset metadata

kpherwan:
Issue - https://github.com/informatics-isi-edu/chaise/issues/1801

Related to PR - https://github.com/informatics-isi-edu/chaise/pull/2097

- Appending JSON-LD to the head tag of records with the new annotation of tag:isrd.isi.edu,2021:google-dataset

- Validation of generated JSON-LD is done using the types and allowed attributes in jsonldSchema.js. This file was created by parsing the TTL file provided by schema.org and follows schema.org https://schema.org/Dataset strictly. We have included only particular classes as of now for the sake of simplicity. 

- Added a logging utility to enable logging as per logging levels of "trace","debug","info","warn","error". Eventually chaise-config can supply the logging level to be used and logs would be enabled accordingly.
Future logging done by calling module._log.warn / module._log.info etc will be managed by the utility.


js/googleMetadata.js
```diff
@@ -0,0 +1,253 @@
+
```
RFSH, line 1: Rename this file to `json-ld-validator.js` since it's closer to what it actually is doing.

js/googleMetadata.js
```diff
@@ -0,0 +1,253 @@
...
+var typeKeyword = "@type";
+var datasetType = "Dataset";
+
+/**
+ * 
+ * @param  jsonLdOrig - the generated jsonLd by templating that needs to be validated
+ * @returns isValid - boolean flag that indicates whether the jsonLd is valid and should be appended to the DOM by Chaise
+ *  modifiedJsonLd - the input object is modified wherever necessary to ensure that the correct properties of the metadata still go through
+ */
+module.performJsonLdValidation = function (jsonLdOrig) {
```
RFSH, line 13: Let's rename this to `valdiateJSONLD`

test/specs/annotation/spec.js
```diff
@@ -6,14 +6,16 @@ require('./../../utils/starter.spec.js').runTests({
         "/annotation/tests/04.visible_foreign_keys.js",
```
RFSH, line 6: Instead of adding this one to the annotation spec, let's create a new spec and call it "json-ld". And then you should have two separate specs:

1. The test cases that are based on data and testing the integration of validator with Reference API and annotation.
2. Test cases that are directly calling the validator.

And if you want to store the json-lds in a separate file like you're already doing, they should not be in `data` folder (data folder is only for the table data). So the whole folder structure would be

```
test/
test/
|-- specs
|   |-- json-ld
|   |   |-- conf
|   |   |   |-- google-dataset
|   |   |   |   |-- data                # Data
|   |   |   |   |   `-- TABLE_NAME.json
|   |   |   |   `-- schema.json         # Schema Definition
|   |   |   `-- google-dataset.conf.json
|   |   |-- files
|   |   |   |-- json-ld-w-whatever.json
|   |   |   |-- other-json-ld-file.json
|   |   |-- tests                       
|   |   |   |-- 01.reference-google-dataset.js
|   |   |   `-- 02.json-ld-validator.js
|   |   `-- spec.js
```

Notice how all the filenames in ermrestjs are kebab-case (lower case with dash) and you should follow the same convention. 

And I think you can add more test cases for the `02.json-ld-validator.js` spec. You should for example test your logic of checking types. It might be easier if you add a comment on top of each file and list the scenarios that you're testing and we can discuss if we should add more or not.

js/reference.js
```diff
@@ -5825,3 +5843,75 @@
...
+
+            if (!templateVariables) {
+                templateVariables = tuple.templateVariables.values;
+            }
+
+            var metadata = {};
+            setMetadataFromTemplate(metadata, metadataAnnotation.dataset, metadataAnnotation.template_engine, templateVariables, table);
+
+            // remove null attributes so they don't get included in the json
+            metadata = removeEmptyOrNull(metadata);
```
RFSH, line 5876: Why do you have to do this before calling the validator? Why shouldn't this be part of the validator given that it's already making some other modifications to the input?

RFSH (changes requested):
Overall the changes look good. The main issue is comment. I didn't go over your validator code and I will do that after you've added better documentation for it.

Also I made some suggestion for rearranging the test cases so the new spec would be separate from the rest.

test/specs/json_ld/conf/google_metadata/data/google_metadata_w_handlebars.json
```diff
@@ -0,0 +1,31 @@
...
+  "thumbnail": "3,001",
+  "cover": "3,001",
+  "opened_on": "1976-06-15 00:00:00",
+  "luxurious": "true",
+  "RID": "1Y2",
+  "RCT": "2021-06-29 15:16:24",
+  "RMT": "2021-06-29 15:16:24"
+},
+{
+  "id": "2003",
```
RFSH, line 18: Change this to a different value

test/specs/json_ld/conf/google_metadata/schema.json
```diff
@@ -0,0 +1,104 @@
+{
+    "schema_name": "google_metadata_schema",
+    "tables": {
+        "google_metadata_w_handlebars": {
+            "kind": "table",
+            "table_name": "google_metadata_w_handlebars",
+            "schema_name": "google_metadata_schema",
```
RFSH, line 7: Add a key definition for `id`, so 

```
"keys": [{
  "unique_columns": ["id"]
}]
```

test/specs/json_ld/conf/google_metadata/schema.json
```diff
@@ -0,0 +1,104 @@
...
+                    }
+                },
+                {
+                    "name": "summary",
+                    "type": {
+                        "typename": "text"
+                    }
+                },
+                {
+                    "name": "RMT",
```
RFSH, line 42: Remove `RMT` and `RCT` definitions from here. These two are added automatically by test framework.

test/specs/json_ld/tests/01.reference_google_dataset.js
```diff
@@ -0,0 +1,84 @@
...
+                    console.dir(err);
+                    done.fail();
+                });
+            });
+
+            it("The google_metadata with all required attributes is successfully generated from the annotation", function (done) {
+                tuple = tuples[0];
+                var metadata = reference.googleDatasetMetadata.compute(tuple);
+                expect(metadata).toBeObject();
+                expect(JSON.stringify(metadata)).toBeJsonString(JSON.stringify(require("../resources/generated_json_ld.json")));
```
RFSH, line 71: I should have mentioned that the value of RID is determined by ermrest and we have to look at database to get it. To get around this, when we're creating the test data, we're creating a `options.entitites` which can be used to get the RID value of a row. You can see an example of it [here](https://github.com/informatics-isi-edu/ermrestjs/blob/master/test/specs/annotation/tests/03.table_display.js#L69).

Ideally this should be a utility function that you can use everywhere in test cases, but since we don't have that capability now, I suggest adding the same helper function and then using it the same way. 

Assuming you added the same helper function, you could do something like this instead:

```js
var expectedVal = require("../resources/generated_json_ld.json");

// since RID value is generated by database, we have to attach it here
var RIDVal = findRID(tableName1, "2003");
expectedVal.url = "https://dev.isrd.isi.edu/chaise/record/google_metadata_schema:google_metadata_w_handlebars/RID=" + RIDVal;

// TODO compare the objects
```

And about the actual expect statement, does `toBeJsonString` even check the object? I think it only makes sure the given string is json parsable. If you're stringifying both objects, then you could just compare their string values.  Although that would make the debugging very difficult. I think you can use the `toHaveSameItems` custom matcher that is defined in [here](https://github.com/informatics-isi-edu/ermrestjs/blob/master/test/utils/jasmine-matchers.js#L9).

test/specs/json_ld/tests/02.json_ld_validator.js
```diff
@@ -0,0 +1,107 @@
+
+exports.execute = function (options) {
+    var module = options.includes.ermRest;
+    describe("validation of structured data", function () {
+        var jsonInputOrig = require("../resources/input_json_ld.json");
+
+        it("The google_metadata with incorrect attribute 'verson' returned w/o it", function (done) {
```
RFSH, line 7: I don't like how you're using the same object and deleting attributes or adding attributes. This spec is supposed to be self contained. So Just create a bare-minimum example of json-ld for each one and test based on that. You don't need a huge test file with long values and a lot of attributes.

Also write the description of `it` in a general tone. For example this one is `should ignore attributes that are not part of the vocabulary.`

test/specs/json_ld/tests/01.reference_google_dataset.js
kpherwan: Thanks for pointing this out!!

js/json_ld_validator.js
```diff
@@ -0,0 +1,261 @@
...
+                case "DateTime":
+                    if (Date.parse(element)) {
+                        result = true;
+                        return false;
+                    }
+                    break;
+
+                case "Number":
+                case "Integer":
+                    if (typeof element == "number") {
```
RFSH, line 115: Let's change this to also accepts strings that parse properly to number.

js/json_ld_validator.js
```diff
@@ -0,0 +1,261 @@
...
+                        result = true;
+                        return false;
+                    }
+                    break;
+            }
+
+            return true;
+        });
+
+        if (element[typeKeyword]) {
```
RFSH, line 133: Please add an inline comment explaining why you're doing this check and what this is doing

js/json_ld_validator.js
```diff
@@ -0,0 +1,261 @@
...
+            module._log.error("Validation property data json not defined!! \n");
+        }
+    }
+    catch (err) {
+        module._log.error(err);
+    }
+    return {isValid: false, modifiedJsonLd: jsonLd};
+};
+
+function validateSchemaOrgProps(obj) {
```
RFSH, line 35: Please add jsdoc to this

RFSH (approved):
Let's add a little more documentation to your validator and with that I think this PR is ready to be merged.

js/json_ld_validator.js
kpherwan: Good catch

PR #889, opened 2021-07-30
1801 chaise google dataset

kpherwan:
For the annotation `tag:isrd.isi.edu,2021:google-dataset` , if we add null to the array values or if templating results in a null or empty value then currently it gets appended as it is. This PR will remove all null and empty values from the array.


RFSH (changes requested):
Since you squashed the changes in this branch to merge it with master, now github is showing all the commits for this small change.

Given that this is a small change, let's just close this PR, remove the branch, and then create another PR based on a new branch created from master..

You should also add test cases for this.

PR #893, opened 2021-08-10
Saved query display annotation

jrchudy:
Changes for adding `show_saved_query` to the display annotation that is available for catalogs, schemas, and tables. More details found in issue [#2080](https://github.com/informatics-isi-edu/chaise/issues/2080) in chaise in the description under configuration.

Updated to include test cases. I added the catalog test case to the existing spec that tests catalog annotation values and added the other tests to their own spec in "reference".


js/reference.js
```diff
@@ -3626,6 +3626,41 @@
...
+         *    a. if true or false, we are done and use that value
+         *    b. if undefined, display annotation defined but show_saved_query property was undefined
+         *    c. if null, display annotation was not defined (initial value)
+         *  3. check the catalog to see if the value is a boolean
+         *    a. if true or false, we are done and use that value
+         *    b. if undefined, display annotation defined but show_saved_query property was undefined
+         *    c. if null, display annotation was not defined (initial value)
+         *  4. default to false if not defined on any of the above
+         */
+        get showSavedQuery() {
```
RFSH, line 3645: Let's move this to be inside the `Reference.display` API as all the other "display" settings are there. So then chaise would have to do `reference.display.showSavedQuery`

RFSH (approved):

PR #895, opened 2021-08-12
More handlebars helpers

jrchudy:
Adding more handlebars helpers to ermrestJS. 

`jsonStringify` is added for support for facet blobs being returned as objects in the case of plot app for rbk. `jsonParse` is not added because handlebars is meant to return strings to be used in HTML and `jsonParse` was meant to be used to return an "object". From the [handlebars documentation](https://handlebarsjs.com/guide/):
> It uses a template and an input object to generate HTML or other text formats.

`toTileCase` is added for a case in CFDE where we wanted to allow data modelers to capitalize the first letter of each word in a string. `replace` is added for a case in CFDE where we wanted to allow data modelers to replace underscore (`_`) with whitespace (` `). `regexReplace` is not added since `replace` will handle both cases (similar to [String.replace](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace)).

_pending documentation..._


RFSH:
The last time that we talked about plot app, we talked about adding a separate `encodeFacet` that works with objects. Are you not going to do that? So in plot app you're going to use a combination of `jsonStringify ` and `encodeFacet`? I think it's useful if you mention how you're going to use it here (and also have it as part of your test).

jrchudy:
We talked about the change I made to encodeFacet to do some checking to see if the passed in value was an object instead of a string which we agreed was worse than writing `encodeFacetObject` to pair with `encodeFacet` (takes a string input). Then we discussed adding `jsonParse` and `jsonStringify` instead since this gives more flexibility for other potential use cases (that may not actually exist).

In plot app we are going to use `encodeFacet` after calling `jsonStringify`. I'll add a test case for this.

RFSH (approved):

PR #898, opened 2021-08-19
User favorites table path added to 2021:table-config annotation

jrchudy:
Changes to support `user_favorites.storage_table` property of the table-config annotation. `reference.favoritesPath` will return the ermrest path to the table that favorites are stored in for the current table. If the `storage_table` property of `user_favorites` is not defined, `reference.favoritesPath` will return null.

Test cases added to test the default value (`null`), a properly defined annotation, and an annotation with no storage table defined. I could add another test case for having table-config annotation defined but no `user_favorites` but I figured testing the case when `storage_table` is missing would imply the same thing.

_pending annotation.md changes_

### DIFF
--- a/js/core.js
+++ b/js/core.js
@@ -1059,13 +1059,17 @@
          * @desc The path to the table where the favorite terms are stored
          * @type {string}
          */
-        this._favoritesPath = null;
+        this.favoritesPath = null;
         if (this.annotations.contains(module._annotations.TABLE_CONFIG)) {
             var userFavorites = this.annotations.get(module._annotations.TABLE_CONFIG).content.user_favorites;
-            console.log(userFavorites);
-            if (userFavorites && userFavorites.storage_table) {
+            // make sure user_favorites is defined
+            // make sure storage table is an object
+            if (userFavorites && typeof userFavorites.storage_table == "object") {
                 var favoritesTable = userFavorites.storage_table;
-                this._favoritesPath = "/ermrest/catalog/" + favoritesTable.catalog + "/entity/" + favoritesTable.schema + ":" + favoritesTable.table;
+                // make sure each key is present and the value is a non empty string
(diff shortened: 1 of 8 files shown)


js/reference.js
```diff
@@ -2015,6 +2015,24 @@
...
+        /**
+         * The path to the table where the favorite terms are stored. Checks the
+         * 2021:table-config annotation for `user_favorites.storage_table`. If
+         * present, returns a string in the form of:
+         *   "/ermrest/catalog/<catalog-id>/entity/<schema>:<table>".
+         *
+         * If the annotation was not properly defined or not defined at all, this
+         * will return null.
+         */
+        get favoritesPath () {
```
RFSH, line 2027: Let's remove this one and change the `_favoritesPath` that you added in table to be `favoritesPath`. There's no reason to add an API for this in Reference when you can just directly get it from the table

js/core.js
```diff
@@ -1055,6 +1055,20 @@
...
 
+        /**
+         * @desc The path to the table where the favorite terms are stored
+         * @type {string}
+         */
+        this._favoritesPath = null;
+        if (this.annotations.contains(module._annotations.TABLE_CONFIG)) {
+            var userFavorites = this.annotations.get(module._annotations.TABLE_CONFIG).content.user_favorites;
+            console.log(userFavorites);
+            if (userFavorites && userFavorites.storage_table) {
```
RFSH, line 1066: You should make sure `userFavorites.storage_table` is an Object, and all the `.catalog`, `.schema`, `.table` are non empty strings.

RFSH: Also don't forget to remove the console.log

jrchudy: For the favoritesPath, in the case of create, I’m using the the path as is to create a reference. But for delete I’m appending 2 filters to the path (`user_id` and `table.name`) before creating the reference.

RFSH (approved):

PR #911, opened 2021-09-29
$session and $location added to templating

jrchudy:
This PR resolves the issue we opened to add the `$session` object to the templating environment. This PR also includes changes for the `$location` object too. For session, the changes were pretty straightforward after talking with Aref about how we should register the session object with ermrestjs.

For `$session` tests, I tested both mustache and handlebars, with the tests in handlebars being more thorough. I wasn't able to test `$location` since ermrestJS tests operate in a non-browser environment so there was no "location" to begin with. This is something I might have to test in chaise.


js/utils/helpers.js
```diff
@@ -2742,6 +2742,16 @@
         // date object
         obj.$moment = module._currDate;
 
+        // if there is a window object, we are in the browser
+        if (window && window.location) {
+            obj.$location = {
+                origin: window.location.origin,
+                host: window.location.host,
+                hostname: window.location.hostname,
+                chaise_path: window.location.pathname.substring(0, window.location.pathname.indexOf("chaise/") + 7) // +7 to account for "chaise/"
```
RFSH, line 2751: You should not do this. We went through all the trouble of defining chaise and ermrestjs path as part of build so we don't have to hardcode them.

Using `ermrestjsBuildVariables` you can access the variables that we're populating while building ermrestjs and `chaiseBuildVariables` is the chaise equivalent of that.

I would say use `chaiseBuildVariables.chaiseBasePath` if defined, otherwise just default to `/chaise/`.

js/utils/helpers.js
```diff
@@ -2760,6 +2770,56 @@
...
+                    tempAttr.type = "identity";
+                }
+
+                var matchIdx = null;
+                // determine if 'tempAttr' exists in $session.attributes
+                matchIdx = obj.$session.attributes.findIndex(function (targetAttr) {
+                    return targetAttr.id === tempAttr.id;
+                });
+
+                if (matchIdx !== -1) {
```
RFSH, line 2811: I'm not entirely sure what you're doing here. Are you imposing a rule that the "id" attribute must be unique in the `session.attributes`?  Wasn't the goal just to add some extra attributes (type, webpage, etc) to each one?

jrchudy: We discussed merging identities that were the same. I documented that in the issue when we had the discussion about it.

js/utils/helpers.js
```diff
@@ -2760,6 +2770,56 @@
...
+                    Object.assign(obj.$session.attributes[matchIdx], tempAttr);
+                } else {
+                    obj.$session.attributes.push(tempAttr);
+                }
+            });
+
+            // sort the newly created atrtibutes array by display_name
+            obj.$session.attributes.sort(function(a, b) {
+                if (a.display_name && b.display_name) return a.display_name.localeCompare(b.display_name);
```
RFSH, line 2820: Just to make sure this is more robust and would produce deterministic results, I think you need to add more cases here,

- What if only `a` or `b` has `display_name`?
- What if none of them have? I think in this case you would need to sort based on `id`
- Is it safe to assume all the attributes will have `id`?




```

js/utils/helpers.js
RFSH: I forgot that's what we agreed on. In that case I would say add some small comments just explaining what this code is supposed to do

RFSH (approved):

PR #926, opened 2021-12-07
Add a function to add facets to a reference

jrchudy:
Add function to ermrestJS which will addFacets to an existing reference that returns a new reference which is a clone of the original one with the new facets. Added test cases for this but wasn't sure what else I should test.


RFSH (approved):
Changes look good. I think for sanity check you should add another test case for addFacets where the Reference has already facets. Just to make sure addFacets will include both the added facet and the existing one.

PR #929, opened 2022-01-19
Batch unlink association

jrchudy:
Changes in this PR include a delete function for reference for removing multiple association rows at the same time. The intended use of this function is to call it on a reference from one of the rows for the leaf table of the association. 

Some things to note, the function will reject the promise with an error if it fails before any delete requests are made. If delete requests start being made (data validation succeeds), an error response object will be resolved which might not include any errors if all rows were removed successfully. 

Added tests to the delete test spec to verify delete works without deleting all associated rows to main (only the ones passed to the function) and that the validation cases are working.

Edit:
I added more tests to try and delete rows with dynamic acls set and more tests for url length limitation issues.


js/reference.js
```diff
@@ -2117,8 +2136,13 @@
...
                         var keyCol = keyColumns[j],
                         data = tupleData[mapping.get(keyCol).name];
-
-                        if (data === undefined || data === null) return null; // tuple data for keyCol doesn't exist
+                        if (data === undefined || data === null) {
+                            // this returns an error, if a value was shown in the UI with a null value for the right side of the
+                            //         uniqueness constraint, something about the model is broken and further UX actions should not continue
+                            // NOTE: include more context in the error message to communicate information about the parent, assocation, and leaf tables
+                            var err = new module.InvalidInputError("One or more " + associationRef.displayname.value + " tuples have a null value for " + mapping.get(keyCol).name);
+                            defer.reject(err);
```
RFSH, line 2144: Return the promise right after rejecting. Based on the behavior that I saw in the past, just rejecting will not terminate the code path. You could also write them in the same line:

```
return defer.reject(err), defer.promise;
```

js/reference.js
```diff
@@ -2127,25 +2151,103 @@
 
                     // check url length limit if not first one;
                     if (i != 0 && new Reference(module.parse(baseUri + currentPath + (i != 0 ? ';' : '') + filter), self.table.schema.catalog).contextualize.compact._getReadPath().value.length > module.URL_PATH_LENGTH_LIMIT) {
```
RFSH, line 2153: You don't need to call the Reference API at all. Just check the `baseUri + currentPath + (i != 0 ? ';' : '') + filter` length directly.

js/reference.js
```diff
@@ -2076,39 +2076,58 @@
...
+                // var compactPath = associationRef.location.compactPath + '/'; // TODO: remove after reviewing code
+
+                var referencePathObjs = [];
+
+                // keyColumns should be a set of columns that are unique and not-null
                 var keyColumns = associationRef.associationToRelatedFKR.colset.columns; // columns tells us what the key column names are in the fkr "_to" relationship
                 var mapping = associationRef.associationToRelatedFKR.mapping; // mapping tells us what the column name is on the leaf tuple, so we know what data to fetch from each tuple for identifying
 
                 var currentPath = compactPath;
+                var currentCount = 0; // count used for notifying chaise how many each delete request/response refers to
```
RFSH, line 2130: Do you need this? You have the list of requests, so you know the count.

js/reference.js
```diff
@@ -2076,39 +2076,58 @@
...
+         *
+         * @param {Array} mainTuple - an ERMrest.Tuple from Table1 (from example above)
+         * @param {Array} tuples - an array of ERMrest.Tuple objects from Table2 (same as self) (from example above)
          *
-         * @returns {Array} an array of ERMrest.Reference for the AssociationTable that maps the supplied tuples to a row from Table1
+         * @returns {Object} an ERMrest.BatchUnlinkResponse "error" object
          **/
-        getBatchAssociationRef: function (tuples) {
-            var self = this;
+        deleteBatchAssociationRef: function (parentTuple, tuples) {
```
RFSH, line 2103: Just like the `delete` function, this should receive `contextHeaderParams` that will be passed with all the delete requests that it will generate.

js/reference.js
```diff
@@ -2127,25 +2151,103 @@
 
                     // check url length limit if not first one;
                     if (i != 0 && new Reference(module.parse(baseUri + currentPath + (i != 0 ? ';' : '') + filter), self.table.schema.catalog).contextualize.compact._getReadPath().value.length > module.URL_PATH_LENGTH_LIMIT) {
-                        referencePaths.push(currentPath);
+                        // any more filters will go over the url length limit so save the current path and count
+                        // then clear both to start creating a new path
+                        referencePathObjs.push({
```
RFSH, line 2156: If our goal is to keep the rows that didn't get deleted in the list of "selectedRows" in chaise, you would need to also keep the array of original tuples and not just the count (or maybe an array of their indexes and then create the tuples based on that)

js/reference.js
```diff
@@ -2127,25 +2151,103 @@
...
+                // TODO: expand logging information
+                var headers = {};
+                headers[module.contextHeaderName] = {"action": "delete"};
+                var config = {
+                    headers: headers
+                };
+
+                // NOTE: - if we can't trust the url, make a get to loc.ermrestCompactUri to verify the data set is the same (iterate over keys)
+                //       - a further verification is to take etag of read and use as the if-match header as part of delete to make sure the fetched set is the same as the delete set
+                module._http.delete(pathObj.path, config).then(function (res) {
```
RFSH, line 2201: You should use the http object that is under `_server` in the Reference (take a look at Reference.delete or any other places that we're calling http module).

This will ensure the retry logic and all the other http module logic are used.

js/reference.js
```diff
@@ -2127,25 +2151,103 @@
...
-                    var reference = new Reference(module.parse(baseUri + path), self.table.schema.catalog);
-                    references.push(reference);
+                // After last iteration of loop, push the current path
+                referencePathObjs.push({
+                    path: baseUri + currentPath,
+                    count: currentCount
                 });
 
-                return references;
+                var k = 0;
```
RFSH, line 2178: This is not very important, but if you define all the variables in the beginning it makes the code a lot easier to read. I was struggling to find where you've instantiated `k` that is used in the `finally` clause in the `recursiveDelete` function.

js/reference.js
jrchudy: This was taken from the function how you wrote it before. I see why it's unnecessary.

js/reference.js
jrchudy: I decided to keep an array of objects with the column name as one value and the key value as the other. This lets me look up the row based on key column information

RFSH (approved):

PR #930, opened 2022-01-20
Upgrade markdown-it related dependencies

RFSH:
In this PR,
- Upgraded markdown-it to latest version because of the [GHSA-6vfc-qv3f-vr6c vulnerability](https://github.com/advisories/GHSA-6vfc-qv3f-vr6c) (Special patterns with length > 50K chars can slow down parser significantly.)
- Upgraded `markdown-it-attrs` and `markdown-it-container`
- Removed markdown-it related dependencies from package.json to ensure browser and node versions are using the same dependencies from the vendor folder.
- Removed unnecessary files from vendor folder

The only changed behavior because of this update was related to how attributes are working in combination with markdown table. Everything else should continue to work as before. 

I mainly created this PR to run test cases in CI and also start a conversation about dependencies in ermrestjs. I'm still not sure what's the best way to handle dependencies in ermrestjs. Currently, ermrestjs test cases are running in nodejs environment and we're using npm to install the dependencies. This also helps us with finding vulnerabilities through Github (we're listing them in `package.json` and Github is using this list to match with the list of known issues). But, to make sure chaise can also include the dependencies we're keeping the browser-compatible version of dependencies in the `vendor` folder.  As you can see, this is not really ideal. Ideally when we upgrade chaise and ermrestjs, ermrestjs will be using proper build tools to create a bundle that chaise can use. With that, we can ensure the dependencies that we're using while testing and in chaise are the same.

Given this, before merging this PR, I should make sure [this branch of ermrestjs works properly with chaise as well](https://github.com/informatics-isi-edu/chaise/actions?query=branch%3Aupgrade-markdown-it++).


jrchudy:
The only change to syntax is needing an extra `\n` character before the attribute block now?

RFSH:
@jrchudy yes, that's the only thing that I noticed.

RFSH:
Given that we decided to change how we're managing the dependencies in ermrestjs, I made a lot of other changes that are not related just to markdown-it. So for the sake of history I'm going to close this PR and create a new one that includes all the changes.

PR #931, opened 2022-01-24
use scalar projection for all-outbounds if we can

RFSH:
When an all-outbound path's value is needed for a page (either by being visible, or being used in the `wait_for` list of another column) we're modifying the main entity request to include the data for this path.

The current implementation doesn't differentiate between scalar and entity modes and will ask for the whole row instead of just the column that is being projected. For example, let's assume we're looking at table `main` where it has a foreign key relationship to table `o1` through the following column mapping:

```js
{
  "names": [["schema", "main_fk1"]],
  "foreign_key_columns": [{
    "column_name": "fk",
    "table_name": "main",
    "schema_name": "schema"
  }],
  "referenced_columns": [{
    "column_name": "RID",
    "table_name": "o1",
    "schema_name": "schema"
  }]
}
```

If we have a visible-column like the following
```js
{ "source": [{"outbound": ["schema", "main_fk1"]}, "RID"], "entity": true}
```
It will result in the following request:
```
M:=s:main/F1:=left(fk)=(o1:RID)/RID;M:=array_d(M:*),F1:=array(F1:*)@sort(RID)
```
And with this when we're populating data for that visible-column we have the whole row data which allows us to,
- Properly populate the "row name" of this foreign key relationship.
- Access all the column data in the templating environment.
- Honor the `row_order` for sorting.

But in case of scalar all-outbound, we might not need to fetch all the row values and we could potentially just fetch the column's value. To ensure that is the case, I added a `canUseScalarProjection` API to `PseudoColumn` which returns true if,
- It's a scalar all-outbound path
- The leaf column doesn't have  a customized  `markdown_pattern` in `column-display` annotation.
- The leaf column doesn't have a `column_order` that is based on other columns.

With this we can ensure that just fetching the scalar column is enough.


RFSH:
The build in Github Actions started failing for reasons that are not directly related to this PR. The issue was related to [jsdoc](https://github.com/jsdoc/jsdoc) dependency. It seems like the latest 3.6.9 was causing these issues and fixing the version to 3.6.7 solved the issue.

PR #942, opened 2022-04-05
Fix issues with sorting all-outbound paths

RFSH:
This PR will ensure we're using proper aliases while sorting all-outbound paths.

When I modified the `Reference.readPath` to not ask for the whole row of data in scalar all-outbounds, I [broke the sort functionality](https://github.com/informatics-isi-edu/chaise/issues/2167) and didn't properly test the changes. This PR will fix the sort functionality and also adds more test cases.

Also, the code assumed that the all-outbound path table instances are using aliases in the `F#` format. While this is generally the case, the alias name might be different when the path with prefix syntax is deployed. With these changes, the code is now using the actual alias instead of assuming the format.

### DIFF
--- a/docs/dev-docs/api.md
+++ b/docs/dev-docs/api.md
@@ -8021,8 +8021,8 @@ This function is used to configure the module
 
 | Param | Type | Description |
 | --- | --- | --- |
-| http | <code>Object</code> | Angular $http service object |
-| q | <code>Object</code> | Angular $q service object |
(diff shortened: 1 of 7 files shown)


RFSH:
While writing test cases, I realized that the shared prefix code is sometimes not correctly sharing prefixes when there are no facets in the URL. Since I wanted to test similar scenarios with sort, I fixed it as part of this PR.

PR #943, opened 2022-04-11
Compact select association sub context

jrchudy:
This PR adds new sub contexts to `compact` to allow for further control over display settings in each. The new contexts include:
 - `compact/select/association`
 - `compact/select/association/link`
 - `compact/select/association/unlink`
 - `compact/select/foreign_key`
 - `compact/select/saved_queries`
 - `compact/select/show_more`

This is being done to be consistent with changes being made in chaise to support configuration of individual search popup displays.


RFSH:
@jrchudy I don't remember what we agreed on last time, but what about other variations of `compact/select` (for foreign key, or other popups)? And didn't we have different contexts for `link` and `unlink`?

P.S. You should update [annotation.md](https://github.com/informatics-isi-edu/ermrestjs/blob/master/docs/user-docs/annotation.md#context-names) and also do `make all` to make sure the new API is included in the `api.md`.

jrchudy:
I added more contexts to align with the attached issue and added those to `annotation.md`. We didn't finalize on the names for each other the other search popups contexts, so I used what I wrote in the issue that Hongsuda saw before.

I noticed there was a `compact/entry` context in ermrestJS which I think is probably being used for foreign key popups, so I'm not sure what changes we should make regarding that. It seems like `contextualize.compactEntry` is only being called in ermrestJS code.

js/reference.js
```diff
@@ -53,6 +59,12 @@
             }
         }
 
+        // properly initialize facetPanelDisplay
+        var ccFacetDisplay = res.facetPanelDisplay;
```
RFSH, line 63: I don't think this makes sense. You shouldn't modify the properties inside the object. If the object is valid, we should use it without any modification.

If you define something like which is valid
```
{
  open: ["compact/select/association"]
}
```
Then this code is adding the `closed`, which doesn't make sense. 

If your goal is validation, you should validate each property and if all of them are invalid, then use the default.

test/specs/annotation/tests/03.table_display.js
```diff
@@ -636,13 +639,21 @@ exports.execute = function (options) {
             it ('should be able to access options in annotation.', function (done) {
                 options.ermRest.resolve(tableCompactOptionsEntityUri, {cid: "test"}).then(function (ref) {
                     var ref = ref.contextualize.compact
-                    expect(ref.display.collapseToc).toBeTruthy("Collapse ToC compact missmatch");
-                    expect(ref.display.hideColumnHeaders).toBeTruthy("Hide Column Headers compact missmatch");
+                    expect(ref.display.collapseToc).toBeTruthy("Collapse ToC compact mismatch");
+                    expect(ref.display.hideColumnHeaders).toBeTruthy("Hide Column Headers compact mismatch");
+                    expect(ref.display.facetPanelOpen).toBeTruthy("Facet Panel Open compact mismatch"); // should be true based on default value
```
RFSH, line 644: Shouldn't this be `false`? You're defining the config to only open on association, so it shouldn't be based on the default. right?

js/reference.js
```diff
@@ -36,7 +36,13 @@
 
         var defaultConfig = {
             internalHosts: {"type": "array", "value": []},
-            disableExternalLinkModal: {"type": "boolean", "value": false}
+            disableExternalLinkModal: {"type": "boolean", "value": false},
+            // NOTE: will be expanded later to include `closed` and `hidden`
+            facetPanelDisplay: {"type": "object", "value": {
```
RFSH, line 41: I don't think you should define the default here. It should be in chaise instead.

js/reference.js
```diff
@@ -2410,6 +2422,36 @@
                     this._display.sourceHasWaitFor = false;
                 }
 
+                // current default chaise uses is open for compact and close for all other contexts
+                var fpo = this._context === module._contexts.COMPACT ? true : false;
```
RFSH, line 2426: It's true that in the current implementation (and most probably always) the facet will only be displayed in the `compact` contexts, but you don't need to hard code it here. The less assumption we make and the more general/simple code is, it's easier to maintain.

Also the first line is completely ignoring the config property which would make understanding this feature unnecessary harder. I don't think you should hardcode the chaise's default behavior here. The code here should just look at the context and config without any other assumptions.

RFSH (approved):

PR #946, opened 2022-04-12
Always use shortestkey for self_link (key pseudo-column)

RFSH:
This PR will change self_link presentation logic to always use the shortest key. Most of the changes was updating the test cases.  We should be able to merge this PR after making sure chaise test cases are also updated and passing. To do so, I created [this branch](https://github.com/informatics-isi-edu/chaise/tree/test-self-link-changes) in chaise.


RFSH:
We decided that we don't want to make this change, so I'm going to close the PR.

PR #948, opened 2022-04-15
Update $session

jrchudy:
Change the template variables function for handlebars templating to create the `$session` object as a copy of the `session` passed from chaise. Previously `session.client` had its keys copied to the top level of `$session` and other properties of `session` were being removed (like `expires` and `since`).


RFSH (approved):
Changed look good. Please make sure the `type` property documentation is also updated based on your changes in chaise.

PR #954, opened 2022-06-09
Phsids map and lookup helper

jrchudy:
This PR creates an extra object in the `$session` templating variable to iterate over the `$session.client.extensions.ras_dbgap_permissions` array and use each `phs_id` as a key in the created object with `true` as the value. This is being done for a CFDE use case to prevent looping over a potentially large array for every single generated template.

This PR also adds the `lookup` helper to our templating environment.

### DIFF
--- a/js/utils/helpers.js
+++ b/js/utils/helpers.js
@@ -2776,10 +2776,10 @@
             // If extensions is present, put all dbgap permissions into a map
             // NOTE: not sure if we want to check for `has_ras_permissions` too or not since if that is true, it means ras_dbgap_permissions is also defined
             //       if it's false, the array won't be defined
-            if (session.client.extensions && session.client.extensions.ras_dbgap_permissions) {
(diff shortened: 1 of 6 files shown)


js/utils/helpers.js
```diff
@@ -2772,6 +2772,18 @@
             Object.keys(session).forEach(function (key) {
                 obj.$session[key] = session[key];
             });
+
+            // If extensions is present, put all dbgap permissions into a map
+            // NOTE: not sure if we want to check for `has_ras_permissions` too or not since if that is true, it means ras_dbgap_permissions is also defined
+            //       if it's false, the array won't be defined
+            if (session.client.extensions && session.client.extensions.ras_dbgap_permissions) {
```
RFSH, line 2779: Do we need to add more checks here? I don't know how much of these we can assume, but if we want to be very strict we would have to ensure:

- `session.client.extensions.ras_dbgap_permissions` is an Array
- Each element in the array is an object and has the `phs_id` key.

Without these checks, if the format changes in the webauthn, ermrestjs will throw an error while it's trying to compute the templates.

Issue #903, opened 2021-08-30
Allow more complicated formatting with pre_format (useful for byte count) 


RFSH:
## Summary

This issue was originally created to allow us to format byte count columns. And as a result of that, we introduced [the `humanizeBytes` handlebars function](https://github.com/informatics-isi-edu/ermrestjs/pull/978).

But we also talked about adding similar functionality to `pre_format`. Using `pre_format` is more preferred since it is more flexible. You just have to define it once for the table, and then you can access it in the templating environment. But with handlebars, you have to call this helper function specifically. Also, with `column-defaults,` we can do this on the table, schema, or catalog.

## Description

We talked about the following two ways of supporting this:

1. Allowing a new `ht` type field, with extra parameters called `exponential_based` and `unit_suffix`.
    ```js
    // for SI:
    "pre_format": {
      "format": "%h",
      "exponential_base": 1000,
      "unit_suffix": "B"
    }
    
    // for binary (not sure about this since `iB` doesn't work when it's just bytes)
    "pre_format": {
      "format": "%h",
      "exponential_base": 1024,
      "unit_suffix": "iB"
    }
    ```

2. Adding an extra `scaling_divisor` attribute that works with the existing format types. And allowing the usage of a new `scale` named type:
    ```js
    // for SI:
    "pre_format": {
      "format": "%d %(scale)B",
      "scaling_divisor":  {
        "mode": "si"
      }
    },
    
    // for binary
    "pre_format": {
      "format": "%d %(scale)iB",
      "scaling_divisor":  {
        "mode": "binary"
      }
    }

    // custom
    "pre_format": {
      "format": "%d %(scale)",
      "scaling_divisor":  {
        "mode": "fixed",
        "diviser": 900
      }
    }
    ```

P.S. Related to the SI vs binary, we should make sure that we're using proper conversion in chaise. Currently to humanize the file size as well as transfer rate while uploading, we're using binary. While we should use binary for file size and SI for transfer rate.

**This requires more discussion**

RFSH:
Moved to wishlist.

PR #977, opened 2023-06-01
Display human readable value for asset byte count

RFSH:
In #903, we talked about ways to allow data modelers to change raw integer values into a more readable values.

While we still want to pursue this, in this PR we defined a similar mechanism in asset annotation. This PR adds a new `byte_count` property to `display` settings of asset annotation to provide the "default" presentation setting for `byte_count_column`.

```json
"tag:isrd.isi.edu,2017:asset": {
  "url_pattern": "..",
  "byte_count_column": "bytes",
  "display": {
    "*": {
      "byte_count": "raw|si|binary",
    }
   }
}
```
Where,
- `raw`: Just return the value
-  `si`: 1000 divisor
-  `binary`: 1024 divisor

If the `byte_count` display setting is missing, we will use the "binary" method by default.

As part of this PR I also added the `humanizeBytes` handlebar function that calls the same function. The following is how it can be used:

```
{{humanizeBytes value format number_of_digits}}
{{humanizeBytes 1230000 "si" 3 }}
```

**What's left**:
- We need to finalize the property name and also the handlebar name.
- I only added test cases for the handlebar function. So test cases for the asset integration is missing.
- Depending on how chaise is using assets, this might break some test cases since I've changed the default presentation of byte count.
- Some test cases are failing locally. All of the are related to the error messages that we expect. This is most probably due to the postgres (or ermrest) update.

### DIFF
--- a/docs/user-docs/annotation.md
+++ b/docs/user-docs/annotation.md
@@ -701,6 +701,10 @@ Supported JSON payload patterns:
 Supported display _displayoption_ JSON payload patterns:
 
 - `{`... `"image_preview": true` ... `}`: Display a preview of the selected image below the default presentation of the asset. Be mindful that the client will not do any extra checks whether the selected file is an image, and you should guard against it by using `filename_ext_filter`. Current implementation of Chaise only supports this property in `entry` contexts and defining this for other contexts will not have any effect on Chaise.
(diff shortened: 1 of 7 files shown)


RFSH:
We had a discussion about this change. I'm going to close this PR and create another one with the new changes.

PR #980, opened 2023-06-24
Allow export template to skip adding the root path

RFSH:
This PR will add a new `skip_root_path` property to `source`s that are defined under `outputs` of export templates. This will allow data modelers to include other paths unrelated to the current table/row. 

For example, you can dump all the vocabularies in your deployment alongside the data of the current table. Or you could even store the schema document alongside the data.

P.S. We always add a `?limit=none` to the end of query paths. We need to discuss whether we should continue to do so or skip doing that when this flag is set. We could add another flag for this.


RFSH:
We decided to improve the code related to the `limit=none` I mentioned in the description.

- Instead of blindly adding `?limit=none`, the code will check for `?` in the path and use `&limit=none` instead. 
- We will add the `limit=none` query parameter only if all the following are met:
  - The used API is entity, attribute, attributegroup, or aggregate (by checking the start of the computed `query_path`. so it handles defining `api` explicitly, or having the api as part of `path`).
  - `limit` query param is not already part of the path (by doing a reg exp test).
  - `skip_limit` is not set to true (a new property that we've added to `source`).

PR #986, opened 2023-08-21
Foreign key ux mode annotation support

jrchudy:
This PR adds the `inputDisplayMode` property to foreign key pseudo columns. This is added for 3 annotations currently as the property `selector_ux_mode`. By default, this property is set to `popup` to use the more robust recordset modal popup that includes sorting, filtering, and updating page size.

For defining `inputDisplayMode`, it will look for `selector_ux_mode` in the following order:
 1. defined on visible-columns annotation
 2. defined on foreign-key annotation
 3. defined on table-display annotation when the table is the leaf table in a single outbound foreign key
 4. default value of `popup`
 
These changes go alongside the [changes in chaise](https://github.com/informatics-isi-edu/chaise/pull/2348) to add a another input field for the `dropdown` case.


js/core.js
```diff
@@ -4332,7 +4332,7 @@
 
         getDisplay: function(context) {
             if (!(context in this._display)) {
-                var self = this, annotation = -1, columnOrder = [], showFKLink = true;
+                var self = this, annotation = -1, columnOrder = [], showFKLink = true, inputDisplayMode = 'popup', toTableAnnotation = -1;
```
RFSH, line 4335: Let's use the constant that you defined here (`module._foreignKeyInputModes[0]`) so you don't have to hardcode any values.

docs/user-docs/annotation.md
```diff
@@ -1290,6 +1292,7 @@ The following attributes can be used to manipulate the presentation settings of
       - `ulist` for unordered bullet list.
       - `csv` for comma-seperated values.
       - `raw` for space-seperated values.
+  - `"selector_ux_mode"`: The display mode for the recordedit input field when this column directive is a foreign key relationship, either from the foreign key name or a source path with a single outb ound table. Supported values are `"popup"` and `"dropdown"`, with `"popup"` being the default. Currently only supported in `entry` contexts.
```
RFSH, line 1295: I don't think you need to say all of the "is a foreign key relationship, either from the foreign key name or a source path with a single out bound table". It makes it harder to read. 

I think you should use the exact same explanation as the one you wrote for `foreign-key` annotation. I think that would imply that this only applies only if the column directive is a single-hop outbound foreign key path without needing to explicitly mention it.

PR #987, opened 2023-08-23
Update Github Actions workflow to install missing deriva dependencies

RFSH:
The `flask` branch of webauthn, hatrac, and ermrest is now merged with master. So I had to update some dependencies to make sure our test cases are passing.

P.S. After adding the missing dependencies, Actions was throwing this error during installation:

```
AttributeError: module 'lib' has no attribute 'X509_V_FLAG_CB_ISSUER_CHECK'
``` 

So I had to add the following line (borrowed from [here](https://stackoverflow.com/questions/74981558/error-updating-python3-pip-attributeerror-module-lib-has-no-attribute-openss)):
```
sudo pip3 install -U pyopenssl cryptography
```

Not sure if this the best way to install dependencies or not, but it seems to be working properly.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -22,7 +22,7 @@ jobs:
           node --version
           sudo timedatectl set-timezone America/Los_Angeles
           export PATH="$(systemd-path user-binaries):$PATH"
-          sudo apt-get update
+          sudo apt-get -y update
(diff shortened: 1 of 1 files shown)


RFSH:
I'm going to merge this PR, if we decided to clean this up (or use a better method) we can do it in a separate commit/PR.

PR #988, opened 2023-08-24
Support iframe inputs

RFSH:
This PR will implement the ermrestjs side of changes for the new `input_iframe` property that we discussed [here](https://github.com/informatics-isi-edu/chaise/issues/2308#issuecomment-1670121576).

Summary of changes:

- Changed `Reference.generateColumnsList` to call the new `SourceObjectWrapper.processInputIframe` which will take care of processing `input_iframe` property of column directives.
  - If a column used in the mapping of another column's `input_iframe`, I'm removing it from the visible columns list.
- Changed `Reference.update` to properly add the extra mapped column values to the projection list.
- Added the following props to `ReferenceColumn`:
  - `isInputIframe`: will signal that the column has the `input_iframe` prop.
  - `inputIframeProps`: the computed props that ermrestjs and chaise can use to find the column mapping, iframe url, etc.
  - `renderInputIframeUrl`: Similar to `AssetPseudoColumn.filteredRef` will allow users to generate a url based on the selected data in the form.

Pending:
- documentation
- testing?


js/reference.js
```diff
@@ -1742,60 +1850,13 @@
...
-                                    keyColumns =  column.foreignKey.colset.columns;
-                                }
-
-                                for (n = 0; n < keyColumns.length; n++) {
-                                    keyColumnName = keyColumns[n].name;
-
-                                    addProjection(keyColumnName);
-                                }
+                        if (column.isInputIframe) {
+                            addProjectionForColumnObject(column);
```
jrchudy, line 1854: Looks like this case and the next case are where you are making changes to iterate over the other columns that are part of the input iframe annotation. I like how this is done so that as more properties are added/removed the code here doesn't need to change.

Could you add a comment here that states why this is done differently from the way asset annotation is done?

Should this be done similarly for asset columns that are part of an asset annotation? Currently each column is hardcoded in the above factored out functions.

js/reference.js
```diff
@@ -3402,6 +3426,7 @@
                         // 3. invalid self_link (must be not-null and part of a simple key)
                         // 4. invalid aggregate function
```
jrchudy, line 3427: Update this comment with the newly added case when the column is used as part of the iframe input annotation

test/specs/column/tests/06.iframe_input.js
```diff
@@ -0,0 +1,154 @@
...
+
+        // .optionalFieldNames
+        expect(col.inputIframeProps.optionalFieldNames).toEqual(expectedCol.inputIframeProps.optionalFieldNames, `optionalFieldNames missmatch for i=${i}, name=${col.name}`);
+
+      });
+    });
+
+
+    // TODO
+    describe('renderInputIframeUrl', () => {
```
jrchudy, line 139: A few other TODOs I wasn't sure if you meant to finalize.

jrchudy (commented):
A few TODOs leftover in the code from iframe-template-html and in the iframe_input test spec.

There was also a comment I left about the default value for `forceInputDisabled` since it seems like it is being defined for all `ReferenceColumn` types but should only be `true` for a limited set.

js/column.js
```diff
@@ -308,17 +324,26 @@ ReferenceColumn.prototype = {
         return this._hideColumnHeader;
     },
 
+    /**
+     * in some cases we want to forcefully disable the column even if annotation
+     * or acls allow it. The following are the cases:
+     * - in entry, when the column is used in an existing iframe mapping
+     */
+    forceInputDisabled: function () {
+        this._forceInputDisabled = true;
```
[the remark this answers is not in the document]
RFSH: This was a leftover code that I'll remove.

test/specs/column/tests/06.iframe_input.js
RFSH: Agh.. I completely forgot about these test cases. Thanks for catching it, I'll add them in my next commit.

Issue #909, opened 2021-09-23
Allow modification of source attributes while using sourcekey


RFSH:
Currently if `sourcekey` is used to refer to any of the defined `source-definitions`, the rest of "source attributes" will be ignored. By "source attributes" I'm referring to the attributes that change the definition of a pseudo-column (basically change the generated hash). These attributes are: `source`, `entity`, `aggregate`, and `self_link`. 

- [x] `source` should still be ignored if defined alongside `sourcekey`, but `entity`, `aggregate`, and `self_link` should be allowed. This would allow data-modelers to define a path once and then use `aggregate` in compact while showing the inline table in detailed (or showing it in facet without `aggregate`):

```javascript
"tag:isrd.isi.edu,2019:source-definitions": {
  "sources": {
    "S_taxonomy": {
      "markdown_name": "Taxonomy",
      "source": [
        {"sourcekey": "S_core_fact"},
        {"inbound": ["CFDE", "core_fact_ncbi_taxonomy_core_fact_fkey"]},
        {"outbound": ["CFDE", "core_fact_ncbi_taxonomy_ncbi_taxon_fkey"]},
        "nid"
      ]
    }
  }
},
"tag:isrd.isi.edu,2016:visible-columns": {
  "compact": [
       {"sourcekey": "S_taxonomy", "aggregate": "array_d"}
  ],
  "detailed": [
    {"sourcekey": "S_taxonomy"}
  ],
 "filter": {
   "and": [
     {"sourcekey": "S_taxonomy"}
   ]
 }
}
```

- [ ] If we allow this, to make the definition simpler, we might also want to allow sources to refer to each other using `sourcekey`:

```javascript
"tag:isrd.isi.edu,2019:source-definitions": {
  "sources": {
    "S_taxonomy": {
      "markdown_name": "Taxonomy",
      "source": [
        {"sourcekey": "S_core_fact"},
        {"inbound": ["CFDE", "core_fact_ncbi_taxonomy_core_fact_fkey"]},
        {"outbound": ["CFDE", "core_fact_ncbi_taxonomy_ncbi_taxon_fkey"]},
        "nid"
      ]
    },
    "S_taxonomy_array": {
      "sourcekey": "S_taxonomy",
      "aggregate": "array_d"
    }
  }
}
```

RFSH:
[...]

Issue #858, opened 2020-09-02
Tweak visible columns heuristics


RFSH:
The following are improvements that we should potentially make to the existing heuristics for visible columns:

- [x] If a column is part of multiple foreign keys, if any of them are composite, in addition to creating foreign key pseudo-columns, we're adding the column itself to the list of visible columns too. If any of these foreign keys are simple, this can be confusing to the users since they could have the same name and value (except one has a link.) So we should modify the heuristics and skip adding the column if any of the foreign keys that are based on this column is simple.

- [ ] As part of heuristics for `compact` context, we're adding a key pseudo-column to the list of visible columns. We will use a key which:
  - its columns are not part of any foreign key relationship.
  - none of its columns is `nullok`

  If multiple keys fit this description, we will sort them based on the following attributes and pick the first one:
     - shorter
     - has more `text` if they are the same size
     - has a lower column position in the list of `column_definitions` if they are the same size and have the same number of `text` columns.

  This means that if there's a key based on an asset column, it will be picked. In this case the value will display the URL but the link will direct users to the record page which is confusing. So we should add more rules to avoid this. We should at least skip asset columns (and potentially skip other metadata columns as well.)

RFSH:
We had a discussion about the asset issue and decided to change the priority this way:

- If there's a simple key that is made of specific columns (the same that are picked for rowname), use that key.
- Otherwise sort based on the same properties that are defined above while deprioritizing single keys that are made of asset columns (so even a composite key would be chosen over it.)

RFSH:
We had multiple discussions about this and decided to make more changes. For more information refer to the #955 PR.

PR #992, opened 2023-09-28
Changes related to sourcekey and visible columns

RFSH:
In this PR,

- Modified the `sourcekey` support to allow changing `entity`, `aggregate`, and `self_link` at the same time (#909).
- Improved the default visible columns heuristics regarding the display key (more info [here](https://github.com/informatics-isi-edu/ermrestjs/issues/858#issuecomment-807848587)).

In #909 we talked about allowing source definitions to refer to other definitions like this

```json
"tag:isrd.isi.edu,2019:source-definitions": {
  "sources": {
    "S_core_fact": {
      "source": [{"inbound": ["CFDE", "core_fact_id_namespace_fkey"]}, "nid"]
    },
    "S_taxonomy": {
      "markdown_name": "Taxonomy",
      "source": [
        {"sourcekey": "S_core_fact"},
        {"inbound": ["CFDE", "core_fact_ncbi_taxonomy_core_fact_fkey"]},
        {"outbound": ["CFDE", "core_fact_ncbi_taxonomy_ncbi_taxon_fkey"]},
        "nid"
      ]
    },
    "S_taxonomy_array": {
      "sourcekey": "S_taxonomy",
      "aggregate": "array_d"
    }
  }
}
```

What I pushed is just a simple syntactic sugar to avoid copy/paste, and so when this source is used in a visible-column, there is not signal that it's using a different sourcekey. So It cannot properly do the path sharing. But if we want this to also properly handle path sharing, we would need to think more about it and decide how it should be done.

### DIFF
--- a/docs/user-docs/column-directive.md
+++ b/docs/user-docs/column-directive.md
@@ -86,6 +86,9 @@ In this category, the [`sourcekey`](#sourcekey) proprety is used to refer to one
 ```
 {
   "sourcekey" : <source key>,
+  "entity": <true or false>,
+  "aggregate": <aggregate function>,
+  "self_link": <boolean>,
   "markdown_name": <display name>,
   "comment": <tooltip message>,
   "comment_display": <inline|tooltip>,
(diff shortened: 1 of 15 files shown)


RFSH:
After discussing these changes, we decided to revert the extra changes I included in this PR.

While the change related to column heuristics is something we should do, we have to make sure we're doing it consistently for row-name as well. If possible, we would like to keep these two as close as possible. 

And the change to allow `sourcekey` inside `source-definitions` requires more discussion.

Issue #463, opened 2017-07-24
Reference.update doesn't check all values for changes properly


jrchudy:
When the column is of type `date`, `timestamp`, or `timestamptz`, the check to make sure the values have changed before adding the column name to the list of projections may not work properly.

When I saw this issue, I had a `timestamp` column with:
 - old value = `2016-01-18T13:00:00`
 - new value = `2016-01-18T13:00:00.000-08:00`

The `timestamptz` column had these values:
 - old value = `2016-01-18T00:00:00-08:00`
 - new value = `2016-01-18T00:00:00.000-08:00`

One way to fix this is to convert values for these column types using moment and then compare the values.

hongsudt:
@jrchudy Is this still an issue?

jrchudy:
This is still the case and should be addressed to prevent losing precision for timestamp values.

RFSH:
Since the check in update function is a simple `old !== new`, it will also not properly handle cases where the user's timezone differs from the server's. This is not causing any data loss but will trigger an unnecessary update.

For example, assume the server is in Log Angeles, and the user is in  New York. In this case, the `entity` API would return the `timestamptz` values like `2023-02-02T11:00:00-8:00`, but since we're converting this time to be based on the user's timezone, the submitted data without any change would be `2023-02-02T7:00:00-8:00`.

PR #1006, opened 2024-02-27
Modify update to compare dates/timestamps better

jrchudy:
This PR addresses an issue in chaise where a timestamptz value can be submitted to ermrest for update but the value was unchanged. This can occur if the old/new value has fractional seconds and the other value doesn't. This can also occur when the timezone of the user is different from the timezone the data is returned in.

For the case of date and timestamp, it's almost unlikely that this precision matters but it's still worth converting these values to similar formats anyways to ensure a consistent comparison.

#### using `isSame()` from moment
For comparing dates and timestamps, instead of defining the format and outputting the moment object using `.format()`, I’m letting moment parse the value without telling it the format. To compare the values, using `isSame()` works how we expect when the values can be properly parsed to moment objects.

On thing to note, f the values are both null, moment will create an object with “invalid date” as the value for both. When you use `isSame()` for 2 invalid dates that both were created from null, they are considered different.

Related ermrestJS issue #463

### DIFF
--- a/js/reference.js
+++ b/js/reference.js
@@ -1673,29 +1673,26 @@
                     // don't add a column name in if it's already there
                     // this can be the case for multi-edit
                     // and if the data is unchanged, no need to add the column name to the projections list
-                    // NOTE: This doesn't properly verify if date/timestamp/timestamptz values were changed
-                    if (columnProjections.indexOf(colName) === -1) {
-                        var typename = colType.rootName;
-                        var compareWithMoment = typename === 'date' || typename === 'timestamp' || typename === 'timestamptz';
-                        if (compareWithMoment) {
-                            var moment = module._moment;
-                            var formats = module._dataFormats;
+                    if (columnProjections.indexOf(colName) !== -1) return;
 
-                            // default to DATE format, if timestamp or timestamptz, change the format used
-                            var formatToUse = formats.DATE;
-                            if (typename === 'timestamp') {
-                                formatToUse = formats.TIMESTAMP;
-                            } else if (typename === 'timestamptz') {
-                                formatToUse = formats.DATETIME.return;
-                            }
+                    var oldVal = oldData[colName]
+                    var newVal = newData[colName]
 
-                            var oldVal = moment(oldData[colName], formatToUse, true).format(formatToUse);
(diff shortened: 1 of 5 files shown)


js/reference.js
```diff
@@ -1669,13 +1669,33 @@
...
+                        var typename = colType.rootName;
+                        var compareWithMoment = typename === 'date' || typename === 'timestamp' || typename === 'timestamptz';
+                        if (compareWithMoment) {
+                            var moment = module._moment;
+                            var formats = module._dataFormats;
+
+                            // default to DATE format, if timestamp or timestamptz, change the format used
+                            var formatToUse = formats.DATE;
+                            if (typename === 'timestamp') {
+                                formatToUse = formats.TIMESTAMP;
```
RFSH, line 1687: I don't think this new format that you've added is needed.

You can use the same `YYYY-MM-DDTHH:mm:ssZ` format for both timestamp and timestamptz as long as you don't pass `true` for the last prop. The `true` prop is for "strict mode" which means that it will also validate the value. Which I don't think we need to do here.

js/reference.js
```diff
@@ -1669,13 +1669,33 @@
...
                 // add column name into list of column projections if not in column projections set and data has changed
-                var addProjection = function(colName) {
+                var addProjection = function(colName, colType) {
                     // don't add a column name in if it's already there
                     // this can be the case for multi-edit
                     // and if the data is unchanged, no need to add the column name to the projections list
                     // NOTE: This doesn't properly verify if date/timestamp/timestamptz values were changed
-                    if ( (columnProjections.indexOf(colName) === -1) && (oldData[colName] != newData[colName]) ) {
-                        columnProjections.push(colName);
+                    if (columnProjections.indexOf(colName) === -1) {
```
RFSH, line 1677: This is not that important but I feel like for these cases it's better if you use the reverse and just return here. That way you don't need to indent the code.

So

```js
if (columnProjections.indexOf(colName) !== -1) {
  return;
}

var typename = colType.rootName;
...

```


You should also update the comment above this line as you're now properly verifying date/timestamp/timetsamptz values.

RFSH (changes requested):
Apart from the minor code style changes that I mentioned, I think you should add test cases for this.

RFSH (approved):

PR #1009, opened 2024-03-20
File already in Hatrac, update with new filename

jrchudy:
This PR intends to modify the `fileExists` check in `hatrac.js` for updating the `content-disposition` for a file that already exists (same md5 and content), except the new file has a different name.

One note about this PR is that a new logging action is being added, `"upload/metadata/update"`. This follows a similar pattern with the existing `createUploadJob` that sets `"upload/create"` as the action. I figured `"upload/update"` wasn't specific enough (since we are only updating metadata, not the whole file) and in this case we are "updating" existing content rather than creating something new.

For documentation, where should this "action" be added? I didn't see our other action for upload in the [Action List](https://docs.google.com/spreadsheets/d/1cHhPR0AacvuH2o3QavWlJCIIyPajfn93fJzFnc7VA5M/edit#gid=1180557009) spreadsheet.

### DIFF
--- a/js/hatrac.js
+++ b/js/hatrac.js
@@ -193,6 +193,7 @@ var ERMrest = (function(module) {
 })(ERMrest || {});
 
 var ERMrest = (function(module) {
+    var FILENAME_REGEXP = /[^a-zA-Z0-9_.-]/ig;
 
     var allowedHttpErrors = [500, 503, 408, 401];
 
@@ -310,6 +311,7 @@ var ERMrest = (function(module) {
 
         this.http = this.reference._server.http;
 
+        this.updateDispositionOnly = false;
         this.isPaused = false;
         this.otherInfo = otherInfo;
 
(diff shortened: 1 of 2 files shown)


js/hatrac.js
```diff
@@ -430,13 +432,26 @@ var ERMrest = (function(module) {
...
                 deferred.resolve(self.url);
                 return;
             }
 
+            // hatrac only supports `filename*=UTF-8''` in the content disposition so check for this string to get the filename
+            var contentDispositionPrefix = "filename*=UTF-8''";
+            var filenameIndex = contentDisposition.indexOf(contentDispositionPrefix) + contentDispositionPrefix.length;
+
+            // check if filename in content disposition is different from filename being uploaded
+            // if it is, instead set a different flag and don't mark it as done
```
RFSH, line 448: Can't we send the put request to update the metadata here? That way chaise doesn't need to know anything about this filename/metadata change. From chaise perspective, the file exists and so it can just skip creating the upload job.

jrchudy:
@RFSH I've updated the code to make the change you suggested. It makes sense to update the `content-disposition` in the `fileExists` function in `hatrac.js` since chaise won't act on this any differently. We aren't planning on communicating to the user that a file has metadata updated instead of a file being uploaded in the UI, so there's no need to separate these steps.

One note I should have mentioned before is about the logging. I updated the main body above to comment about the added log action and if we should go with what I proposed or if there is a better phrase to use.

js/hatrac.js
```diff
@@ -430,13 +431,52 @@ var ERMrest = (function(module) {
...
+
+            // check if filename in content disposition is different from filename being uploaded
+            // if it is, create an update metadata request for updating the content-disposition
+            if (contentDisposition.substring(filenameIndex, contentDisposition.length) != self.file.name.replace(FILENAME_REGEXP, '_')) {
+                // Prepend the url with server uri if it is relative
+                var url =  self._getAbsoluteUrl(self.url + ";metadata/content-disposition");
+
+                var data = "filename*=UTF-8''" + self.file.name.replace(FILENAME_REGEXP, '_');
+
+                if (!contextHeaderParams || !_isObject(contextHeaderParams)) {
```
RFSH, line 454: If `contextHeaderParams` wasn't passed to this function, line 417-423 would create it. So this `if` statement is pointless and this code will always use the same action for both requests. Instead you should assume `contextHeaderParams` is available here and do

```
contextHeaderParams.action = 'upload/metadata/update';
```

So 

```ts
...

var data = "filename*=UTF-8''" + self.file.name.replace(FILENAME_REGEXP, '_');

contextHeaderParams.action = 'upload/metadata/update';
var config = {
    headers: _generateContextHeader(contextHeaderParams)
};
config.headers['content-type'] = 'text/plain';

...
```

RFSH (approved):
I manually tested uploading on chaise and ran the unit test cases. Everything works as epcted.

The changes overall look good and I just asked for a small change related to how the logging is done.

PR #1010, opened 2024-03-25
Custom filename for content-disposition

jrchudy:
This PR adds support for an new annotation property on the asset annotation, `stored_filename_pattern`. This property will allow for data modelers to customize the `content-disposition` value that is stored in hatrac for the file and the value stored in the `filename_column` (if defined on the annotation). This pattern uses the same template variables as `url_pattern` with a new one added called `filename_stem`, the name of the file without the extension.

More details in the comment on [issue 1949](https://github.com/informatics-isi-edu/chaise/issues/1949) in chaise.


RFSH (approved):
I think you should also update line 449 and 452 (where you recently changed so we would update the filename of already uploaded files) to use `.storedFilename` instead of `.file.name`.

Apart from that everything looks good to me and test cases passed locally.

js/hatrac.js
```diff
@@ -828,13 +829,19 @@ var ERMrest = (function(module) {
...
         row[this.column.name].sha256 = this.hash.sha256;
-        row[this.column.name].filename_ext = _getFilenameExtension(this.file.name, this.column.filenameExtFilter, this.column.filenameExtRegexp);
+        row[this.column.name].filename = this.file.name;
+        var filename_ext = _getFilenameExtension(this.file.name, this.column.filenameExtFilter, this.column.filenameExtRegexp);
+        row[this.column.name].filename_ext = filename_ext
+        // filename_basename is everything from the file name except the last ext
+        // For example if we have a file named "file.tar.zip"
+        //    => "file.tar" is the basename
+        //    => ".zip" is the extension
+        row[this.column.name].filename_basename = this.file.name.substring(0, this.file.name.indexOf(filename_ext));
```
RFSH, line 844: Since we know the exact length of the string that we want to remove, you should just remove that many characters from the end of the string (instead of using `indexOf`). Otherwise this might remove other parts of the filename.

Also `filename_ext` might be `null`.  So you should check for it and return the whole filename in this case.

RFSH (changes requested):

js/hatrac.js
```diff
@@ -298,6 +298,7 @@ var ERMrest = (function(module) {
         this.CHUNK_QUEUE_SIZE = otherInfo.chunkQueueSize || 4;
 
         this.file = file;
+        this.storedFilename = file.name; // the name that will be used for content-disposition and filename column
```
RFSH, line 301: It seems like chaise might call this function without providing any `file` object. So you should check that file is an object before using `file.name`.

I found this while working on migrating test cases. In this specific tests, we're not providing any files for an asset input and just submitting. Chaise throws an error on this line.

RFSH (approved):

PR #1013, opened 2024-05-17
Column RID exposed for use in chaise

jrchudy:
Added functions to ermrestJS to allow for chaise to use column RID values instead of column names for unique information cases where special characters shouldn't be used. We don't have control over the column names that are used in the database but we do know the format of RIDs.

Changes include:
 - added `RID` to `ReferenceColumn` and `ForeignKeyPseudoColumn`
 - added `linkedDataRIDs` to `Tuple` for getting the RID of the column linked data is used for

### DIFF
--- a/js/column.js
+++ b/js/column.js
@@ -230,10 +230,13 @@ ReferenceColumn.prototype = {
     },
 
 
+    /**
+     * @type {string}
+     * @desc ermrest generated RID for this column
+     */
     get RID() {
         if (this._RID === undefined) {
-            // what do we use if not simple?
-            this._RID = this._simple ? this._baseCols[0].RID : null;
+            this._RID = this._baseCols[0].RID;
         }
         return this._RID;
     },
@@ -2028,6 +2031,11 @@ Object.defineProperty(ForeignKeyPseudoColumn.prototype, "name", {
     }
 });
 
+/**
+ * returns the ermrest generated RID for the foreign key relationship this pseudo clumn represents
+ * @member {string} RID
+ * @memberof ERMrest.ForeignKeyPseudoColumn#
+ */
 Object.defineProperty(ForeignKeyPseudoColumn.prototype, "RID", {
     get: function () {
         if (this._RID === undefined) {
--- a/js/column.js
+++ b/js/column.js
@@ -2028,6 +2028,15 @@ Object.defineProperty(ForeignKeyPseudoColumn.prototype, "name", {
     }
 });
 
+Object.defineProperty(ForeignKeyPseudoColumn.prototype, "RID", {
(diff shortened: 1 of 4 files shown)


js/column.js
```diff
@@ -229,6 +229,14 @@ ReferenceColumn.prototype = {
         return this._displayname;
     },
 
+
+    get RID() {
+        if (this._RID === undefined) {
+            // what do we use if not simple?
+            this._RID = this._simple ? this._baseCols[0].RID : null;
```
RFSH, line 236: Let's just use `this._baseCols[0].RID` regardless of `this._simple`. That way this will always return something.

js/column.js
```diff
@@ -229,6 +229,14 @@ ReferenceColumn.prototype = {
         return this._displayname;
     },
 
+
+    get RID() {
```
RFSH, line 233: Please add a comment for this.

js/column.js
```diff
@@ -2020,6 +2028,15 @@ Object.defineProperty(ForeignKeyPseudoColumn.prototype, "name", {
     }
 });
 
+Object.defineProperty(ForeignKeyPseudoColumn.prototype, "RID", {
```
RFSH, line 2031: Please add a comment for this.

js/reference.js
```diff
@@ -5294,18 +5297,22 @@
...
                 // add the main table data to linkedData
                 this._linkedData = [];
+                this._linkedDataRIDs = [];
                 for (i = 0; i < data.length; i++) {
                     tempData = {};
+                    linkedDataMap = {};
                     for (j = 0; j < fks.length; j++) {
                         fkName = fks[j].name;
                         tempData[fkName] = {};
+                        linkedDataMap[fks[j].name] = fks[j].RID
```
RFSH, line 5307: Please use `fkName` like the line above it. I know these are the same thing, but it might be confusing why we didn't use the same thing later. Also you forgot the `;`.

js/reference.js
```diff
@@ -5757,6 +5765,19 @@
...
+          /**
+          * Foreign key data RID names.
+          * During the read we get extra information about the foreign keys,
+          * client could use these extra information for different purposes.
+          * One of these usecases is domain_filter_pattern which they can
+          * include foreignkey data in the pattern language.
+          *
+          * @type {Object}
+          */
+          get linkedDataRIDs() {
```
RFSH, line 5777: At first I didn't like this new API but now that I saw where you're using this in chaise I think this makes sense. And your comment explains it very well.

RFSH (approved):
Everything looks good to me. I asked for some minor changes and comments.

PR #1015, opened 2024-08-27
Association display

jrchudy:
This PR adds a `BulkCreateForeignKeyObject` object to ermrestJS that is attached to reference once the `compute` function is called to initialize the object. 

I left a PR comment about `prefillObject` maybe being here instead of `chaise` and another one about composite keys.

Test cases added in this PR:
 - table has 2 FKs + 3 columns that are NOT system columns
   - both FKs together are part of the same key
 - table has 2 FKs that are NOT system columns
   - the FKs are no part of key information
 - table has 3 FKs that are NOT system columns
 - table has a composite foreign key from the association table to the main table
   - ensures this feature isn't triggered for composite FKs
 - pure and binary association table


js/reference.js
```diff
@@ -4014,6 +4014,29 @@
...
+                // TODO: error with message "Call computePrefill first"
+            }
+            return this._prefill;
+        },
+
+        computePrefill: function (prefillObject) {
+            if (this._prefill === undefined) {
+                var annotation = null;
+                // var prefillAnno = table.annotations.get(module._annotations.TABLE_DISPLAY).content;
+                // if (prefillAnno) annotation = prefillAnno;
```
jrchudy, line 4034: I should probably remove this until ew make the annotation changes.

js/reference.js
```diff
@@ -4014,6 +4014,29 @@
...
 
+        /**
+         * If prefill object is defined and has the required attributes, will return
+         * a Prefill object with the necessary objects used for a association modal picker
+         *
+         * @type {ERMrest.Prefill}
+         */
+        get prefill() {
+            if (this._prefill === undefined) {
+                // TODO: error with message "Call computePrefill first"
```
jrchudy, line 4025: I wasn't sure what was the best way to handle this. My TODO was intended to remind me to discuss this with you.

js/reference.js
```diff
@@ -6595,3 +6618,126 @@
...
+     * @param {ERMrest.Reference} reference reference for the association table
+     * @param {Object} prefillObject computed prefill object after extracting the query param and fetching the data from cookie storage
+     * @param {Object} prefillAnnotation
+     */
+    function Prefill (reference, prefillObject, prefillAnnotation) {
+        verify(prefillObject, "'prefillObject' must be defined");
+
+        var self = this;
+        this._reference = reference;
+        this._prefillObject = prefillObject;
```
jrchudy, line 6633: One change I mentioned in Chaise code (I think) was that maybe I should move `prefillObject` entirely from chaise and have it as part of the reference. If I do that, then there should be a `get prefillObject()` function in the prototype of `Prefill`

js/reference.js
```diff
@@ -6595,3 +6618,126 @@
...
+            // column should be a foreignkey pseudo column
+            if (!column.isForeignKey) return;
+
+            nonSystemColumnFks.forEach(function(fk) {
+                // column and foreign key `.name` property is a hash value
+                if (column.name === fk.name) {
+                    if (prefillObject.fkColumnNames.indexOf(column.name) !== -1) {
+                        self.mainColumn = column;
+                    } else {
+                        // TODO: what if there are multiple "other" fk columns?
```
jrchudy, line 6664: I briefly discussed this with Hongsuda yesterday in standup. She stated at that point it's no longer an association since we don't have an annotation defined yet to "assist" in figuring this out.

js/reference.js
```diff
@@ -6595,3 +6618,126 @@
...
+
+        // to calculate isUnique
+        //   - One of the keys should contain the main and leaf foreign key columns
+        var tempKeys = reference.table.keys.all().filter(function(key) {
+            var keyCols = key.colset.columns;
+            return !(keyCols.length == 1 && (module._serialTypes.indexOf(keyCols[0].type.name) != -1 ||  module._systemColumns.indexOf(keyCols[0].name) != -1) && !(keyCols[0] in fkCols));
+        });
+
+        // if we determine there is a key with 2 foreign key columns in it, should we assume that's the main/leaf columns for the association
+        //   even if there are more than 2 foreign key columns?
```
jrchudy, line 6688: This comment goes along with the comment I made above with `OR` at the start of it. If we don't want to be "smart" and base `isAssociation` only on having 2 FKs, then this comment and the one above don't matter

js/reference.js
```diff
@@ -4014,6 +4014,29 @@
             return this._cascadingDeletedItems;
         },
 
+        /**
+         * If prefill object is defined and has the required attributes, will return
+         * a Prefill object with the necessary objects used for a association modal picker
+         *
+         * @type {ERMrest.Prefill}
+         */
+        get prefill() {
```
RFSH, line 4023: We should come up with a better and more specific name.

js/reference.js
```diff
@@ -6595,3 +6618,126 @@
...
         });
     }
+
+    /**
+     *
+     * @param {ERMrest.Reference} reference reference for the association table
+     * @param {Object} prefillObject computed prefill object after extracting the query param and fetching the data from cookie storage
+     * @param {Object} prefillAnnotation
+     */
+    function Prefill (reference, prefillObject, prefillAnnotation) {
```
RFSH, line 6628: You need to clean this up. It's really hard to understand what attributes this prototype has. Please follow what we've done in other places in ermrestjs. Populate the private props inside the constructor and then add getter function for any one that you want to expose.

```js
function Prefill(reference, prefillObject) {
   // populate all the _NAME props here, for example:
  this._mainColumn = ...
}

Prefill.prototype = {
  get mainColumn() {
    return this._mainColumn
 }
  
}
```

And do could do the following for props that have their own logic that doesn't involve other props:

```
get andFiltersForLeaf () {
  if (this._andFiltersForLeaf === undefined) {
    ...
    this._andFiltersForLeaf = ...
  }
  return this._andFiltersForLeaf;
}

```

js/reference.js
```diff
@@ -6595,3 +6618,126 @@
...
+
+            nonSystemColumnFks.forEach(function(fk) {
+                // column and foreign key `.name` property is a hash value
+                if (column.name === fk.name) {
+                    if (prefillObject.fkColumnNames.indexOf(column.name) !== -1) {
+                        self.mainColumn = column;
+                    } else {
+                        // TODO: what if there are multiple "other" fk columns?
+                        if (self.leafColumn) {
+                            // throw error saying "more than 1 foreign key column available"
```
RFSH, line 6666: You should handle any corner case that you think can cause issues here. Apart from the case that you mentioned, I'm assuming that the composite keys are also not supported by this, right?

If that's the case you should check for them and make sure we're not engaging the chaise UI in these cases.

I don't think you need to throw any errors. Just treat this the cases that we don't want to support now. So for example if `reference.prefill` returns a `null`, it means something was wrong and we should just continue as normal in chaise without using this new UI.

RFSH (changes requested):

js/reference.js
```diff
@@ -4014,6 +4014,74 @@
...
+            return this._bulkCreateForeignKeyObject;
+        },
+
+        computeBulkCreateForeignKeyObject: function (prefillObject) {
+            if (this._bulkCreateForeignKeyObject === undefined) {
+                if (!prefillObject) {
+                    this._bulkCreateForeignKeyObject = null;
+                } else {
+                    // ignore the fks that are simple and their constituent column is system col
+                    // TODO: composite FKs
```
jrchudy, line 4036: I'm not sure what to do for composite keys.

js/reference.js
```diff
@@ -4014,6 +4014,83 @@
...
+         */
+        computeBulkCreateForeignKeyObject: function (prefillObject) {
+            if (this._bulkCreateForeignKeyObject === undefined) {
+                if (!prefillObject) {
+                    this._bulkCreateForeignKeyObject = null;
+                } else {
+                    // ignore the fks that are simple and their constituent column is system col
+                    // TODO: composite FKs
+                    var nonSystemColumnFks = this.table.foreignKeys.all().filter(function(fk) {
+                        return !(fk.simple && module._systemColumns.indexOf(fk.colset.columns[0]) !== -1);
```
jrchudy, line 4047: I think I should modify this check to the following to check eliminate composite foreign keys:
```
return fk.simple &&  module._systemColumns.indexOf(fk.colset.columns[0]) === -1
```
The foreign key has to be simple and NOT a system column.

jrchudy:
@RFSH This is ready for review now with the changes to chaise that are ready. We should discuss about the 2 comments I left in this PR about:
 - how do we make sure the foreign key is not part of a composite foreign key
 - should the `prefillObject` be a property of `reference.bulkCreateForeignKey` or stay as a state variable in `chaise`

js/reference.js
```diff
@@ -4014,6 +4014,87 @@
...
 
+        /**
+         * If prefill object is defined and has the required attributes, will return
+         * a BulkCreateForeignKeyObject object with the necessary objects used for a association modal picker
+         *
+         * @type {ERMrest.BulkCreateForeignKeyObject}
+         */
+        get bulkCreateForeignKeyObject() {
+            if (this._bulkCreateForeignKeyObject === undefined) {
+                verify(false, 'Call "computeBulkCreateForeignKeyObject" with the prefill object first');
```
RFSH, line 4025: I feel like instead of throwing error we should just do

```
this._bulkCreateForeignKeyObject = null;
```

So if the compute is called, it will return the proper value. And if it's not called, it will just return `null`. We then don't need to necessarily call compute in all the different places in chaise just to make sure this doesn't throw any errors.

js/reference.js
```diff
@@ -6595,3 +6676,113 @@
...
+                }
+            });
+
+            if (leafMatch && mainMatch) self._isUnique = true;
+        });
+    }
+
+    BulkCreateForeignKeyObject.prototype = {
+        /**
+         * @returns ERMrest.ForeignKeyPseudoColumn
```
RFSH, line 6730: Please also explain what do we mean by "leaf column".

js/reference.js
```diff
@@ -6595,3 +6676,113 @@
...
+         */
+        get leafColumn () {
+            return this._leafColumn;
+        },
+
+        /**
+         * @returns boolean
```
RFSH, line 6737: Add description

js/reference.js
```diff
@@ -6595,3 +6676,113 @@
...
+         */
+        get isUnique () {
+            return this._isUnique;
+        },
+
+        /**
+         * @returns {Object[]} filters array for getting the rows that should be disabled
```
RFSH, line 6744: Add description

js/reference.js
```diff
@@ -4742,6 +4823,14 @@
             return this._contextualize(module._contexts.COMPACT_SELECT_FOREIGN_KEY);
         },
 
+        /**
+         * The _compact/select/foreign_key/bulk_ context of this reference.
+         * @type {ERMrest.Reference}
+         */
+        get compactSelectBulkForeignKey() {
+            return this._contextualize(module._contexts.COMPACT_SELECT_BULK_FOREIGN_KEY);
```
jrchudy, line 4831: I'm unsure how this name should be written. I considered `compactMultiSelectForeignKey` but that was inconsistent with the rest of our "select" contexts which don't specify single or multi.

I think `compactSelectBulkForeignKey` or `compactSelectForeignKeyBulk` both make sense, it comes down to where we want to put the term `bulk`. I could also change this to "multi" instead of bulk but figured bulk makes more sense since this is for "bulk create" compared to how we use "bulk edit" on recordset page.

js/utils/constants.js
```diff
@@ -90,6 +90,7 @@
         COMPACT_SELECT_ASSOCIATION_LINK: 'compact/select/association/link',
         COMPACT_SELECT_ASSOCIATION_UNLINK: 'compact/select/association/unlink',
         COMPACT_SELECT_FOREIGN_KEY: 'compact/select/foreign_key',
+        COMPACT_SELECT_BULK_FOREIGN_KEY: 'compact/select/foreign_key/bulk',
```
jrchudy, line 93: I figured having `bulk` at the end would be consistent with our other `contexts` and allows for proper inheritance of contexts. 

To reiterate from the other comment, using `multiselect` would be inconsistent with the other contexts we have with `select` in them. This also breaks inheritance. That's also why `bulk` (or `multi`) should be at the end of the string.

js/reference.js
RFSH: I think what you have is good.

js/column.js
```diff
@@ -1663,6 +1663,15 @@ function ForeignKeyPseudoColumn (reference, fk, sourceObjectWrapper, name) {
 // extend the prototype
 module._extends(ForeignKeyPseudoColumn, ReferenceColumn);
 
+/**
+ * Given the available tuple data, generate the uniqueId for the selected row from the table this pseudo column points to
+ *
+ * @param {*} linkedData
```
RFSH, line 1669: Can you please add a small description to this param that clarifies what this is. It's the key-value pair for the column values of the table that this pseudo column points to. I'm saying this because `linkedData` (Tuple.linkedData and what we're passing to `.filteredRef`) is not just a key-value pair of one table. It has the data for all the outbound fks.

RFSH (approved):
I think everything looks good and works as expected. I like the name that you used for the new context. Can you please add the new context to the ["Context names" section of annotation doc](https://github.com/informatics-isi-edu/ermrestjs/blob/master/docs/user-docs/annotation.md#context-names).

You should also do `make all`, so it generates the new `api.md` and then push it.

PR #1018, opened 2024-10-22
Bulk create foreign key property added to annotations

jrchudy:
This PR adds the functionality to check multiple annotations for the `bulk_create_foreign_key` property to use for the bulk create foreign key functionality in recordedit app. [This previous PR](https://github.com/informatics-isi-edu/ermrestjs/pull/1015) added the functionality when certain heuristics were met (which included limiting this to tables with only 2 foreign keys). This annotation allows for a table with any number of foreign keys to use this feature provided the model is annotated according to the spec.

Tests that are included in this PR are as follows:
 - for tables with 3 foreign keys, annotations are defined that use 1 of those foreign keys when coming from "main" as the prefill
   1. visible columns annotation with display property set on `main <- association` fk column source definition that points to the "third" table instead of the "leaf" table
   2. foreign key annotation with display property property set on `main <- association` fk column source definition that points to the "third" table instead of the "leaf" table
   3. table display annotation with `bulk_create_foreign_key_candidates` set on association table that points to "third" table instead of "leaf" table 
      - has ['leaf constraint name', 'third constraint name'] but leaf is not in visible columns so "third" is used instead
 - for tables with 2 foreign keys, annotations are defined to turn off the feature for both foreign keys or just 1 foreign key
   1. display annotation on table with false set to turn off feature for main prefill AND leaf prefill
   2. fk annotation on `main < association` fk with false set to turn off the feature for main prefill
      - leaf prefill should not check any annotation and use the heuristics for the table
   3. visible columns entry from `main < association` fk with false set to turn off the feature for main prefill
      - leaf prefill should not check any annotation and use the heuristics for the table

More details about these annotation changes can be found in the [chaise issue](https://github.com/informatics-isi-edu/chaise/issues/2416).


js/column.js
```diff
@@ -2164,6 +2164,12 @@ Object.defineProperty(ForeignKeyPseudoColumn.prototype, "display", {
                 if (module._foreignKeyInputModes.indexOf(displ.selector_ux_mode) !== -1) {
                     sourceDisplay.inputDisplayMode = displ.selector_ux_mode;
                 }
+
+                // check visible-columns annotation display property and set to override property from fk annotation or display annotation
+                // make sure `false` is not ignored since that turns this feature off
+                if (displ.bulk_create_foreign_key !== null) {
```
RFSH, line 2170: `null` is also a valid value here, right? So if I have

```
"vis-col": {
  "*": [
    { 
       "source": ...,
       "bulk_create_foreign_key": null
    }
  ]
}
```

and then on the fk
```
"foreign-key": {
  "display": {
    "*" : {
       "bulk_create_foreign_key": false
    }
  }
}
```

We're supposed to use he heuristics, right?

docs/user-docs/annotation.md
```diff
@@ -615,6 +622,7 @@ Supported JSON _option_ payload patterns:
 	- The _pathsuffix_ MAY join additional tables to the path and MAY project from these tables as well as the table bound to the `S` table alias.
 	- The _pathsuffix_ SHOULD reset the path context to `$S` if it has joined other tables.
 - `"selector_ux_mode"`: The display mode for the recordedit input field when this table is part of a foreignkey relationship as the `outbound` table. Supported values are `"facet-search-popup"` and `"simple-search-dropdown"`, with `"facet-search-popup"` being the default. Currently only supported in `entry` contexts.
+- `"bulk_create_foreign_key_candidates"`: Use this property to control the bulk selection of foreign key values in `entry/create` context when there is a prefill query parameter. Supported value is an array of foreign key `names` in the format of `[['schema_name', 'foreign_key_name'], ...]`. This will override the `bulk_create_foreign_key` property defined in the display annotation. Currently only supported in `entry/create` context.
```
RFSH, line 625: Doesn't this property also support `false` and `null`?

docs/user-docs/annotation.md
```diff
@@ -1337,6 +1345,7 @@ The following attributes can be used to manipulate the presentation settings of
       - `csv` for comma-seperated values.
       - `raw` for space-seperated values.
   - `"selector_ux_mode"`: The display mode for the recordedit input field when this column directive is a foreign key relationship. Supported values are `"facet-search-popup"` and `"simple-search-dropdown"`, with `"facet-search-popup"` being the default. Currently only supported in `entry` contexts.
+  - `"bulk_create_foreign_key"`: Use this property to control the bulk selection of foreign key values in `entry/create` context when there is a prefill query parameter. Supported values are a foreign key `name` in the format of `['schema_name', 'foreign_key_name']` from the schema document, `false`, and `null`. Using a foreign key name will use that foreign key as the one being bulk selected if that foreign key is in the visible columns list. `false` turns off the heuristics that trigger this feature. `null` will override inheritance for this property and use the default heuristics. This will override the `bulk_create_foreign_key` property defined in the display property of the foreign-key annotation. Currently only supported in `entry/create` context.
```
RFSH, line 1348: Please also add this to `docs/user-docs/column-directive.md`, and don't forget to update the table of contents.

RFSH: This is not related to this line in the code, but I'm commenting this here since it's related to the annotation doc:

It seems like in the previous set of changes you used `compactSelectBulkForeignKey` instead of the proper context name in the "Context Names" section. Please also update that.

js/core.js
```diff
@@ -4781,9 +4788,37 @@
...
+                 */
+                // check display annotation on table/schema/catalog
+                // if true, don't set anything and let the heuristics be used
+                // if false, set to false to turn off heuristics
+                if (this.table._shouldUseBulkCreateForeignKey === false) bulkCreateConstraintName = false;
+
+                if (this.table.annotations.contains(module._annotations.TABLE_DISPLAY)) {
+                    var tableAnnotation = module._getRecursiveAnnotationValue(context, this.table.annotations.get(module._annotations.TABLE_DISPLAY).content);
+
+                    if (tableAnnotation.bulk_create_foreign_key_candidates !== null) {
```
RFSH, line 4807: Same comment about allowing `null`. I think for all these checks it's better if you check for the exact types that we allow. For example here it could be

```tsx
var prop = tableAnnotation.bulk_create_foreign_key_candidates;

if (prop === null || prop === false || Array.isArray(prop)) {
  bulkCreateConstraintName = prop;
}
```

js/core.js
```diff
@@ -4781,9 +4788,37 @@
...
+                        bulkCreateConstraintName = tableAnnotation.bulk_create_foreign_key_candidates;
+                    }
+                }
+
+                // check foreign key annotation
+                // make sure `false` is not ignored since that turns this feature off
+                if (displayAnnot.bulk_create_foreign_key !== undefined && displayAnnot.bulk_create_foreign_key !== null) {
```
RFSH, line 4814: Same comment about checking for the values that we allow instead of the way you currently wrote this.

js/reference.js
```diff
@@ -4081,39 +4083,111 @@
...
+                            return foundColumn;
+                        }
+
+                        var leafCol = null;
+                        // use for loop so we can break if we find the leaf column
+                        for (i = 0; i < self.columns.length; i++) {
+                            var column = self.columns[i];
+
+                            // column should be a foreignkey pseudo column
+                            if (!column.isForeignKey) continue;
```
RFSH, line 4144: Does the chaise code handle composite fks for the "leaf column"? I think it handles composite fks for the "main column" but no the leaf. If that's the case you should also check for that here and don't allow data modelers to define this for a composite fk.

RFSH (changes requested):
Overall I like how you modified the `ForeignKeyRef.getDisplay` to return a `bulkForeignKeyCreateConstraintName` property regardless of how this property was defined. This makes the changes that you did to `Reference` a lot easier to follow. But I think you need to be more restrictive when you're populating this.

Since you're not checking for the exact value, I could technically define a `bulk_create_foreign_key` property as an array of constraint names. Or define `bulk_create_foreign_key_candidates` with just one constraint name. So I think you should modify the checks that you have to make sure it's following the structure that we expect. Since you're doing this in multiple places, you could define a helper function similar to `_isValidModelComment` function.

Also please do `make all` once you're done with your changes.

docs/user-docs/annotation.md
jrchudy: The `bulk_create_foreign_key_candidates` property of `table-display` annotation is only an array of constraint names. The `bulk_create_foreign_key` property of `display` annotation can be defined on the table and allows for only `false` or `null` so I don't think the `*_candidates` property should do the same thing on the same model element.

RFSH (approved):

PR #1022, opened 2025-04-15
Change build process to allow TypeScript

RFSH:
## Summary

This PR changes the build step to use [Vite](https://vite.dev/) to allow using newer JavaScript syntax and TypeScript. With this new method, we can gradually rewrite everything to TypeScript.

## Details

By forcing ES5, the code has become very difficult to follow. The generated `api.md` was supposed to make this easier, but given the complexity of the code, that document is not easy to read either. The new requirements would need a lot of ermrestjs changes, so instead of continuing the same route, I decided to refactor the code to allow TypeScript.

To do so, I introduced a proper build tool called [Vite](https://vite.dev/). Before committing to this tool, I also tried webpack, but found it too slow and verbose for ermrestjs. With webpack, we would have to include a lot of other dependencies (e.g., `babel` and `babel-loader`). But with Vite, everything is included (so we won't have the dependency management issue that Chaise has). 

The following are the other packages that I introduced to improve the development experience:

- `prettier`: To ensure consistent code styles.
- `eslint`: To find the errors and warnings in the code.
- `husky`, `commitlint`, `lint-staged`: To automatically run `eslint` and `prettier` on each commit. This is an industry-standard method to ensure consistency in the code. But it might be annoying if we want to make small document changes. I'll have to evaluate this.

As part of this PR, I also rewrote part of the code to TypeScript. I only rewrote the more essential and globally used code paths. For the rest, I just fixed the bugs or linter issues so the build can pass. We should rewrite it to TypeScript as we add features or fix bugs in the old code.


RFSH:
After a discussion we decided to not include `husky`, `commitlint`, and `lint-staged`. We didn't like how this process automatically changes the code.

PR #1128, opened 2026-06-10
fix: fix the canUseTCRS check

### LINKED ISSUE #1127: Dynamic ACLs (tcrs) are not properly used in all cases
Currently we're only asking for `tcrs` if the table has dynamic ACLs. So if the table has static ACLs but one of the columns has a dynamic ACL, ermrestjs is not properly handling that.

This results in chaise not properly disabling inputs in the recordedit page.

RFSH:
This PR fixes the `canUseTCRS` check. The previous check was only ensuring that the table has dynamic ACLs. So in cases where the table was using static ACLs and we only had `acl_bindings` on the columns, we were not properly fetching `tcrs`.

### DIFF
--- a/src/models/reference/reference.ts
+++ b/src/models/reference/reference.ts
@@ -1688,7 +1688,7 @@ export class Reference {
   /**
    * Returns true if
    *  - ermrest supports tcrs, and
-   *  - table has dynamic acls, and
+   *  - table or any of the columns of the table have dynamic acls, and
    *  - table has RID column, and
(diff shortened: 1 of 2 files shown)


RFSH:
The codeQL failed because of [the ongoing authentication issues in GitHub](https://www.githubstatus.com/incidents/fcj3088jg1wx). I didn't make any substantial code changes that would cause it to fail, so I merged the PR.

